PR #3, opened 2015-02-16
Make nxagent aware of its NX'ish version string (and number).

sunweaver:
Until now, nxagent uses a hard-coded version string internally that is set to "3.5.0" in nx-X11/programs/Xserver/hw/nxagent/Init.c.

This change picks the version number from <base>/VERSION, populates nx-X11/config/cf/nxversion.def with it and propagates it through to the nxagent's Imakefile.


Makefile
```diff
@@ -19,6 +19,11 @@ NXLIBDIR    ?= $(PREFIX)/lib/nx
 X2GOLIBDIR  ?= $(PREFIX)/lib/x2go
 CONFIGURE   ?= ./configure
 
+NX_VERSION_MAJOR=$(shell cat VERSION | cut -d"." -f1)
+NX_VERSION_MINOR=$(shell cat VERSION | cut -d"." -f2)
+NX_VERSION_MICRO=$(shell cat VERSION | cut -d"." -f3)
+NX_VERSION_PATCH=$(shell cat VERSION | cut -d"." -f4)
+
```
Ionic: Maybe it's better to use a "real" shell script here which fetches the version number part and, if it is unset, defaults to "0", to always have the parts well-defined?

Makefile
```diff
@@ -44,7 +49,20 @@ build-full:
 	cd nxcomp && autoconf
 	cd nxcompext && autoconf
 	cd nxcompshad && autoconf
+
+	# prepare nx-X11/config/cf/nxversion.def
+	cat nx-X11/config/cf/nxversion.def.in | sed \
```
Ionic: Useless use of cat.

```Makefile
sed \
	-e 's/###NX_VERSION_MAJOR###/$(NX_VERSION_MAJOR)/' \
	-e 's/###NX_VERSION_MINOR###/$(NX_VERSION_MINOR)/' \
	-e 's/###NX_VERSION_MICRO###/$(NX_VERSION_MICRO)/' \
	-e 's/###NX_VERSION_PATCH###/$(NX_VERSION_PATCH)/' \
	nx-X11/config/cf/nxversion.def.in \
	> nx-X11/config/cf/nxversion.def
```

mikedep333:
Is there a reason why we are differing from upstream? Like because our build system is so different? With VcXsrv, both I and marha use the convention that X.org uses:
http://cgit.freedesktop.org/xorg/xserver/commit/?id=3a06faf3fcdb7451125a46181f9152e8e59e9770&utm_source=anzwix
http://cgit.freedesktop.org/xorg/xserver/tree/hw/xfree86/common/xorgVersion.h?id=xorg-server-1.17.1

In particular, note that they use:
XORG_VERSION_MAJOR
XORG_VERSION_MINOR
XORG_VERSION_PATCH
XORG_VERSION_SNAP
(Although I agree that major, minor, micro, and patch is more intuitive.)

Ionic:
NX always used their own version numbering scheme and has incremented the MAJOR version often, while X never has. I feel the MAJOR/MINOR/MICRO/PATCH definition is a better fix for nx-libs.

Makefile
Ionic: The script can be as easy as:

```bash
#!/bin/bash
COMPONENT="$1"
VERSION_FILE="VERSION"

# More than one line is not supported.
VER="$(head -n "1" "${VERSION_FILE}" | cut -d"." -f"${COMPONENT}")"

[ "x${VER}" = "x" ] && VER="0"

printf "${VER}"
```

mikedep333:
Understood, I will continue reviewing.

mikedep333:
I need to learn make syntax better. Based on my review, this looks OK to me.

Ionic:
Well, this is additionally complicated, because this is not exactly make syntax. That's imake syntax, which is different and even requires a pre-processor...

sunweaver:
Hi Mihai, hi Michael,

I just pushed three more commits to the feature/nxagent-version PR branch.

If there are no more objections within the next 9h (1300 UTC), I will merge.

Mike

sunweaver:
... I will rebase the new commits with commit 0e341f5 before merging...

sunweaver:
About the versioning in NX. In X2Go we received a patch that allows for having for digit versions. There the original author also used M.m.µ.p as naming scheme. I like it back then and adopted it here.

PR #4, opened 2015-03-03
library clean-up: Don't build libNX_Xpm anymore. Use system's libXpm shared library.

sunweaver:
One release goal for version 3.6.x of nx-libs is dropping as many bundled libraries as possible that haven't been adapted to nx-libs.

  Starting with libNX_Xpm here.

  The libNX_Xpm library is only referenced once (nx-X11/programs/Xserver/hw/nxagent/Holder.c).
  When grepping through Xserver/hw/nxagent the suspicion comes up, that the libXpm linkage is not
  needed at all, because none of the provided functions are used inside the nxagent Xserver.


mikedep333:
Looks good.

My one concern is that you use wording like "With recent versions of NX". Should we use the term "NX" instead of "nx-libs"?

sunweaver:
Yeah, you are probably right...

mikedep333:
Mike, do you want to continue using the term "NX"?

sunweaver:
Hi Michael,

I had changed NX -> nx-libs after your first comment on that. Have I missed something?

Mike

mikedep333:
Hi Mike,

I think you forgot to push that commit. I do not see any 2nd commit for this PR. I also see "NX" on line 693 (and later lines) here:
https://github.com/ArcticaProject/nx-libs/blob/3.6.x/debian/control

sunweaver:
oh... changes like those are not subject to this PR. Please open an issue for that to lead the discussion there. Thanks!
Mike

mikedep333:
Agreed, libXpm is not the only package whose description uses the term "NX".

mikedep333:
I opened #5

PR #10, opened 2015-04-11
Xlib cleanup (part 1)

sunweaver:
This PR drops the following shipped-with X11 libraries from NX:
- Xdmcp
- XComposite
- Xdamage
- Xfixes
- Xtst

The system-wide installed X11 libraries of the same names are used instead now.


sunweaver:
@Ionic: If you have time, please review.

Ionic:
Ok, will do. That's a huge one, 154 changed/deleted files with about 15k deleted lines.

sunweaver:
@Ionic: yeah, fully aware of that. More to come!

debian/libnx-x11-6.symbols
```diff
@@ -547,7 +547,7 @@ libNX_X11.so.6 libnx-x11-6 #MINVER#
  XcmsTekHVCWhiteShiftColors@Base 3.5.0.29
  XcmsUNDEFINEDColorSpace@Base 3.5.0.29
  XcmsVisualOfCCC@Base 3.5.0.29
- XdmcpWrap@Base 3.5.0.29
+#MISSING: 2:3.5.99.0-0~build1# XdmcpWrap@Base 3.5.0.29
```
Ionic, line 550: Hm… are those symbols really missing? Won’t linking fail?

nx-X11/config/cf/X11.tmpl
```diff
@@ -1154,19 +1143,23 @@ FCHOWN_DEFINES = -DHAS_FCHOWN
...
 
 #ifndef XDamageDefines
-#if BuildDamage
+#if UseInstalledDamageLib
 #define XDamageDefines -DDAMAGE
 #else
 #define XDamageDefines /**/
 #endif
+#else
+#define XDamageDefines /**/
```
Ionic: This looks suspicious. If `XDamageDefines` is not defined and we’re using the external `libXDamage`, set `XDamageDefines` to `-DDAMAGE`. Ok. If there’s no external `libXDamage`, set `XDamageDefines` to empty, so that `XDamage` is not used/supported by `NX` at all. Makes sense.

What **doesn’t make sense** is to **reset** XDamageDefines to empty if it’s **already defined**, though?!

nx-X11/config/cf/X11.tmpl
```diff
@@ -1154,19 +1143,23 @@ FCHOWN_DEFINES = -DHAS_FCHOWN
...
 
 #ifndef CompositeDefines
-#if BuildComposite
+#if UseInstalledXcompositeLib
 #define CompositeDefines -DCOMPOSITE
 #else
 #define CompositeDefines /**/
 #endif
+#else
+#define CompositeDefines /**/
```
Ionic: Comment from above applies here, too.

nx-X11/config/cf/X11.tmpl
```diff
@@ -625,23 +624,13 @@ XORGRELSTRING = XorgManVersionString
 #ifndef BuildXfixes
 #define BuildXfixes		YES
 #endif
-#ifndef BuildXfixesLibrary
-#define BuildXfixesLibrary	!BuildServersOnly
-#endif
```
Ionic, line 630: Btw, what’s the difference between `BuildFoo` and `BuildFooLibrary`? Does the first one define whether to build the `Foo-Proto` part or not and the second one is relating to `libfoo`?

nx-X11/config/cf/host.def
```diff
@@ -133,18 +133,10 @@ XCOMM $XFree86: xc/config/cf/xf86site.def,v 3.186 2003/06/25 18:06:22 eich Exp $
...
-#define BuildXfixesLibrary              YES
-
-#define BuildCompositeLibrary           YES
-
 #define BuildRandRLibrary               YES
 
 #define BuildDocs                       NO
 
-#define BuildComposite			YES
-
```
Ionic, line 147: Note that  `nx-X11/config/cf/X11.tmpl` sets `BuildComposite` to `(BuildXfixes && BuildDamage)` if not defined. As none are defined here, it will default to `YES`.

nx-X11/include/extensions/Imakefile
```diff
@@ -42,15 +42,6 @@ XINERAMAHEADERS = Xinerama.h panoramiXext.h panoramiXproto.h
...
 #endif
-#if BuildXfixes || BuildXfixesLibrary
-XFIXESHEADERS = xfixeswire.h xfixesproto.h
-#endif
-#if BuildDamage || BuildDamageLibrary
-DAMAGEHEADERS = damagewire.h damageproto.h
-#endif
-#if BuildComposite || BuildCompositeLibrary
-COMPOSITEHEADERS = composite.h compositeproto.h
-#endif
```
Ionic, line 53: Removing these lines is potentially  problematic, because `BuildXfixes`, `BuildDamage` and `BuildComposite` are set to `YES` in `nx-X11/config/cf/X11.tmpl`.

If `BuildXfixes`, `BuildDamage` and `BuildComposite` mean, that the internal header files should be used, it’s probably better to either unset or set thee to `NO` (not sure what method is correct for imake) in `nx-X11/config/cf/X11.tmpl`.

nx-X11/include/extensions/Imakefile
```diff
@@ -69,7 +60,7 @@ EXTRAHEADERS = $(SCREENSAVERHEADERS) $(XF86MISCHEADERS) $(XF86BIGFONTHEADERS) \
 	       $(XF86VIDMODEHEADERS) $(XF86DGAHEADERS) $(XINERAMAHEADERS) \
 	       $(LBXHEADERS) $(XVHEADERS) $(XVMCHEADERS) $(XF86RUSHHEADERS) \
 	       $(FONTCACHEHEADERS) $(RENDERHEADERS) $(RANDRHEADERS) \
-	       $(XFIXESHEADERS) $(DAMAGEHEADERS) $(COMPOSITEHEADERS) \
+	       $(XFIXESHEADERS) \
```
Ionic: `XFIXESHEADERS` will be unset according to the previous hunk.

nx-X11/include/extensions/xfixeswire.h
```diff
@@ -1,104 +0,0 @@
...
-
-#define SaveSetMap			    0
-#define SaveSetUnmap			    1
-
-/*************** Version 2 ******************/
-
-#define WindowRegionBounding		    0
-#define WindowRegionClip		    1
-
-#endif	/* _XFIXESWIRE_H_ */
```
Ionic, line 104: - `nx-X11/include/extensions/damageproto.h`
- `nx-X11/include/extensions/damagewire.h`

Haven’t been removed — probably overlooked?

nx-X11/lib/Imakefile
```diff
@@ -130,22 +130,6 @@ RANDRLIBDIR = Xrandr
...
-DAMAGELIBDIR = Xdamage
-#endif
-
-#if BuildCompositeLibrary
-COMPOSITELIBDIR = Xcomposite
-#endif
-
-#if BuildXcursorLibrary
-XCURSORLIBDIR = Xcursor
-#endif
```
Ionic, line 147: Why removing `BuildXcursorLibrary` here? We never shipped this, but still? This should probably not be part of this PR but a unique one?

nx-X11/lib/Imakefile
```diff
@@ -223,7 +202,7 @@ LINTSUBDIRS = $(XAULIBDIR) $(XDMCPLIBDIR) $(X11LIBDIR) \
 	     $(XFTLIBDIR) $(XVMCLIBDIR) $(RANDRLIBDIR) $(XTRAPLIBDIR) \
 	     $(XRESLIBDIR) $(APPLELIBDIR) $(DMXLIBDIR) $(WINDOWSLIBDIR) \
 	     $(XEVIELIBDIR) $(XFIXESLIBDIR) $(DAMAGELIBDIR) \
```
Ionic: `XFIXESLIBDIR` and `DAMAGELIBDIR` are now unset.

nx-X11/lib/X11/Imakefile
```diff
@@ -1063,13 +1060,12 @@ EXTRASHAREDOBJS = $(STATICLCOBJS)
 
 SpecialCLibObjectRule(ErrDes,$(ICONFIGFILES),$(EDB_DEFINES))
 SpecialCLibObjectRule(StrKeysym,$(ICONFIGFILES),$(KDB_DEFINES))
-SpecialCLibObjectRule(ConnDis,$(ICONFIGFILES),$(XDMAUTHDEFS) $(RPCDEFS) $(CONN_DEFINES) $(SOCK_DEFINES) $(POLL_DEFINES) $(XTRANS_X_DEFINES) $(K5INCL) $(K5DEFS))
```
Ionic, line 1066: `XDMAUTHDEFS` is still defined as `-DHASXDMAUTH`. Should it really be removed here, or the whole `#if HasXdmAuth` section above be removed, too?

nx-X11/lib/Xcursor/AUTHORS
```diff
@@ -1,2 +0,0 @@
-Keith Packard, HP
-
```
Ionic, line 2: I still think `Xcursor` removal should be part of a unique PR (or this PR be extended to `Xcursor` in general — by also removing references in X11.templ and host.def)

nx-X11/lib/Xrender/Imakefile
```diff
@@ -159,7 +159,7 @@ all:: xrender.pc
 
 xrender.pc: xrender.pc.in
 	RemoveFile($@)
-	sh ../Xcursor/config-subst $(SUBSTVARS) < xrender.pc.in > $@
```
Ionic, line 162: Reminder: part of `Xcursor` removal.

nx-X11/programs/Xserver/hw/nxagent/Events.c
```diff
@@ -62,7 +62,7 @@
 #include "NXvars.h"
 #include "NXproto.h"
 
-#include "xfixesproto.h"
+#include <X11/extensions/xfixesproto.h>
 #define Window     XlibWindow
 #define Atom   XlibAtom
 #define Time XlibXID
```
Ionic, line 68: `nx-X11/programs/Xserver/hw/nxagent/Composite.c` includes `composite.h`, which must also be changed in the same vain.

nx-libs.spec
```diff
@@ -551,7 +527,6 @@ Group:          Development/Libraries
 Summary:        Include files and libraries for NX development
 Requires:       libNX_X11-devel%{?_isa} = %{version}-%{release}
 Requires:       libNX_Xau-devel%{?_isa} = %{version}-%{release}
-Requires:       libNX_Xdmcp-devel%{?_isa} = %{version}-%{release}
 Requires:       libNX_Xext-devel%{?_isa} = %{version}-%{release}
 Requires:       libNX_Xfixes-devel%{?_isa} = %{version}-%{release}
```
Ionic: `libNX_Xfixes-devel` should probably also be removed?

nx-libs.spec
```diff
@@ -733,30 +708,22 @@ ln -s -f ../../../../%{_lib}/libXext.so.6 %{buildroot}%{_libdir}/nx/X11/Xinerama
 %post -p /sbin/ldconfig
 %post -n libNX_X11-6 -p /sbin/ldconfig
 %post -n libNX_Xau6 -p /sbin/ldconfig
-%post -n libNX_Xcomposite1 -p /sbin/ldconfig
-%post -n libNX_Xdamage1 -p /sbin/ldconfig
-%post -n libNX_Xdmcp6 -p /sbin/ldconfig
 %post -n libNX_Xext6 -p /sbin/ldconfig
 %post -n libNX_Xfixes3 -p /sbin/ldconfig
```
Ionic: Same here: also removing `libNX_Xfixes3`?

nx-libs.spec
```diff
@@ -733,30 +708,22 @@ ln -s -f ../../../../%{_lib}/libXext.so.6 %{buildroot}%{_libdir}/nx/X11/Xinerama
...
 %post -n libXcompext3 -p /sbin/ldconfig
 %post -n libXcompshad3 -p /sbin/ldconfig
 %postun -p /sbin/ldconfig
 %postun -n libNX_X11-6 -p /sbin/ldconfig
 %postun -n libNX_Xau6 -p /sbin/ldconfig
-%postun -n libNX_Xcomposite1 -p /sbin/ldconfig
-%postun -n libNX_Xdamage1 -p /sbin/ldconfig
-%postun -n libNX_Xdmcp6 -p /sbin/ldconfig
 %postun -n libNX_Xext6 -p /sbin/ldconfig
 %postun -n libNX_Xfixes3 -p /sbin/ldconfig
```
Ionic: Same here.

nx-libs.spec
```diff
@@ -804,23 +771,6 @@ ln -s -f ../../../../%{_lib}/libXext.so.6 %{buildroot}%{_libdir}/nx/X11/Xinerama
...
-
-%files -n libNX_Xdmcp-devel
-%defattr(-,root,root)
-%{_libdir}/libNX_Xdmcp.so
-%{_includedir}/nx/X11/Xdmcp.h
-
-%files -n libNX_Xdmcp6
-%defattr(-,root,root)
-%{_libdir}/libNX_Xdmcp.so.6*
-
```
Ionic, line 823: Btw: will rpmlint accept empty packages?

Ionic, line 823: Also, there's a chance a %files section per package is strictly required and removing them will completely break the specfile. But I'll have to look/ask around to find that out.

Ionic:
I'm through the diff now, but I wouldn't feel good merging this PR in its current state.

I'd appreciate input to my previous comments.

debian/libnx-x11-6.symbols
sunweaver, line 550: The XdmcpWrap() function is only used once in nx-X11/lib/X11/ConnDis.c.

XdmcpWrap() is actually a symbol of libXdmcp that used to be source-symlinked-into libX11 in the old days. Nowadays, the symbol is available via <X11/Xdmcp.h> (which already gets included in nx-X11/lib/X11/ConnDis.c).

nx-X11/config/cf/X11.tmpl
sunweaver: Indeed. Fixed in the coming-in push (576f83cb0caa6520aac7196ce8b627f51de9adf5).

nx-X11/config/cf/X11.tmpl
sunweaver: Dito, fixed in soon-to-come 1ea6f186d70d952eb0f1401a12332cdd1c54f14e.

nx-X11/config/cf/X11.tmpl
sunweaver, line 630: Exactly.

nx-X11/config/cf/host.def
sunweaver, line 147: Exactly, I want to use the default here.

nx-X11/include/extensions/Imakefile
sunweaver, line 53: Due to your review I more clearly see the differences between lib and proto part of the Xserver.

I will re-add all proto-related stuff to this PR and update it (to latest X.org or so!) later with another PR.

nx-X11/lib/X11/Imakefile
sunweaver, line 1066: Good catch! $(XDMAUTHDEFS) should stay configured for ConnDis.c.

nx-X11/lib/Xcursor/AUTHORS
sunweaver, line 2: Agreed.

nx-X11/programs/Xserver/hw/nxagent/Events.c
sunweaver, line 68: Actually no. That is, as said earlier/above, I will undo the proto stuff removal in this PR, so we need to stick with "..." header inclusion.

nx-libs.spec
sunweaver: Good catch!

nx-libs.spec
sunweaver: Good catch!

nx-libs.spec
sunweaver: Good catch!

nx-libs.spec
sunweaver, line 823: We will find out...

sunweaver:
I have started a new round with this PR, changing several parts of the PR. Will provide a second approach in some minutes.

nx-X11/include/extensions/xfixeswire.h
sunweaver, line 104: Actually yes, but now that I chose to keep the wire/proto headers, it is wanted to have them.

debian/libnx-x11-6.symbols
Ionic, line 550: "via [`libXdmcp`]", I guess? Ok.

nx-X11/include/extensions/Imakefile
Ionic, line 53: No, please don't. I have been talking to keithp et. al.

If we're switching to system libraries, we should ALSO switch to system headers, including headers for the protocols.

Whatever we do, we're screwed anyway and will likely have to fix either subtle or not so subtle bugs.

Thus, it's better to completely base `damage`, `composite` and friends on the newer system libraries and headers.

nx-X11/include/extensions/xfixeswire.h
Ionic, line 104: Bad idea. We should compile against what we link.

nx-libs.spec
Ionic, line 823: rpmlint will accept empty packages. But a %files section **must** exist per package, or otherwise no RPM will be created at all.

The %files section can be empty.

This said, the whole dummy package stuff isn't a good idea and severely frowned upon by SUSE staff. I'll have to get a clearer cross-distro understanding here.

sunweaver, line 823: In Debian/Ubuntu, the dummy packages are a clear workflow element for autoremoving packages on dist-upgrades.

I have no clue, how RPM based distros handle this, but it seems logic to provide an upgrade path to an empty package that has no other (forward + reverse) dependencies. So that the actual obsolete files (shipped with previous package versions) get removed from the system.

Another option might be the "Obsoletes:" field in .spec.files. But as said before, I have no clue.

Ionic, line 823: The preferred way for SUSE seems to be leaving the old package around. It will be automatically removed on dist-upgrades if the new dist version doesn't provide a package anymore.

So, I propose the following: completely remove the `libnx-x{damage,composite,dmcp,tst}*` packages from the `nx-libs.spec` file and for Debian keep the dummy package change you have proposed, BUT also add `libxcomposite1` etc. dependencies to `Depends:` as mentioned on https://wiki.debian.org/Renaming_a_Package#Method_2

sunweaver, line 823: Agreeing on the SUSE approach.

Adding libxSOMELIBX manually under Depends: is not needed in Debian/Ubuntu and not correct. The dependencies are all gathered automatically under Depends: in bin:package nxagent via automatic library detection. We are not renaming those packages, we are switching to a different library via Build-Depends:.

nx-X11/include/extensions/Imakefile
sunweaver, line 53: Ok. Interesting. This may become tricky here then.

To be clear on this, if we use e.g. libxcompose1 instead of shipping outdate libnx-xcompose1, means we drop composite.h and compositeproto.h from nx-X11/include/extensions. Which leads to using the headers of the same name from x11proto-composite-dev (Debian package name). Right?

Does it also mean, that we update nx-X11/programs/Xserver/composite and take some latest X.org version of that folder and replace the one in nx-libs with that?

I have already played with the latter, and I am almost sure it will end up in replacing the whole nx-X11/programs/Xserver/ code tree, because of the manifold ABI changes inside xorg-xserver over the past ten years.

sunweaver, line 53: Maybe we should actually look at how it could be possible to squeeze nx-X11/programs/Xserver/hw/nxagent into a real hardware driver project building against xserver-xorg.

However, this will fail, as we will need to continue building against libnx-x11-6, libnx-xrender3, libnx-xinerama1, etc.

nx-libs.spec
Ionic, line 823: For Fedora/EPEL: once X2Go upgrades to nx-libs 3.6.x, we'll need the `libXcomposite` (etc.) maintainer to add an `Obsoletes:` tag for libNX-Xcomposite. We can't do this ourselves...

sunweaver:
This approach is to bulky. We learned a lot from it in our discussions, but the clean-up has partially been done already, partially needs more work in smaller steps.

PR #19, opened 2015-04-28
libnx-xinerama1: also create libXinerama symlink in postinst (and remove in prerm.)

Ionic:
Due to the `nx-x11-common` package being a `noarch`/`allarch` package,
creating the symlink in nx-libs' `Makefile` will lead to the symlink
referencing the "default" architecture dpkg uses for building
`noarch`/`allarch` packages.

Incidentally, this worked fine for Debian, as `amd64` seems to be the
default architecture. On Ubuntu, however, the default architecture up to
Vivid (15.04) was `i386`. For those builds, the symlink pointed to the 32
bit library of `libNX_Xinerama.so.1` -- essentially breaking this feature.

Move the symlink creation to the arch-sensitive `libnx-xinerama1` package.
The `postinst` and `prerm` scriptlets will work fine, unless someone
installs the `i386` package version **after** the `amd64` version. Given that
we already create symlinks to `libNX_X11` and friends using that method,
no new regression is introduced. Strictly speaking that's a bug, but
we'll hopefully clean that up later...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,11 +155,12 @@ install-full:
 	    done; \
 
 	# Provide means for Xinerama support in NX/X2Go sessions. This
-	# This also requires two post-install symlinks:
-	# (DESTDIR)$(NXLIBDIR)/X11/Xinerama/libNX_X11.so.6 -> /usr/<libdir>/libX11.so.6
-	# (DESTDIR)$(NXLIBDIR)/X11/Xinerama/libNX_Xext.so.6 -> /usr/<libdir>/libXext.so.6
-	$(INSTALL_DIR) $(DESTDIR)$(NXLIBDIR)/X11/Xinerama
-	$(INSTALL_SYMLINK) $(USRLIBDIR)/libNX_Xinerama.so.1 $(DESTDIR)$(NXLIBDIR)/X11/Xinerama/libXinerama.so.1
+	# This also requires three post-install symlinks created by libnx-xinerama1:
+	# $(DESTDIR)$(NXLIBDIR)/X11/Xinerama/libNX_X11.so.6 -> /usr/<libdir>/libX11.so.6
+	# $(DESTDIR)$(NXLIBDIR)/X11/Xinerama/libNX_Xext.so.6 -> /usr/<libdir>/libXext.so.6
+	# $(DESTDIR)$(NXLIBDIR)/X11/Xinerama/libXinerama.so.1 -> /usr/<libdir>/libNX_Xinerama.so.1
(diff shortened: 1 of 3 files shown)


sunweaver:
You may have noticed the recent discussion on x2go-dev [1].

So, basically that kind of libxinerama1 symlinking is obsolete. But we actually need to move the X2Go/Arctica-specific code from libNX_Xinerama1 to the Xserver code tree.

Back to your patch... I think adding the symlinking magic to the postinst of libnx-xinerama1 .deb package is not the approach to go. The symlinks in place definitely need to match the installed nxagent architecture. Of course, we could place the symlinks into a multi-arch directory, as well, but let's not make this more complex if the midterm directive will be the removal of the symlinking stuff, anyway.

I personally think that the symlinking of the libNX_Xinerama1.so file should happen in nx-x11-common (on postinst/prerm) for nx-x11-common, but this nx-x11-common should be Architecture any, not all (and Multi-Arch: foreign). The symlinking target (multi-arch path to libNX_Xinerama1) should be detected at build time and filled into a template debian/libnx-xinerama1.postinst.in. (like with the .install.in files in debian/.

[1] http://comments.gmane.org/gmane.linux.terminal-server.x2go.devel/9523

Ionic:
Why is it obsolete? Everything still depends on it, and our builds are currently broken on Ubuntu due to this.

Detecting the target at build time doesn't seem to be possible. Or would DEB_HOST_ARCH do the trick?

sunweaver:
On  Di 28 Apr 2015 23:50:59 CEST, Mihai Moldovan wrote:

> Why is it obsolete? Everything still depends on it, and our builds  
> are currently broken on Ubuntu due to this.
> 
> Detecting the target at build time doesn't seem to be possible. Or  
> would DEB_HOST_ARCH do the trick?

Yes, it is.

Actually the code is still present in debian/rules, but does not  
apply, because we use the wrong package name.

Let me suggest something on top of your PR.

Mike

## 

DAS-NETZWERKTEAM
fon: +49 (1520) 1976 148

GnuPG Key ID 0x25771B31

freeBusy:
https://mail.das-netzwerkteam.de/freebusy/m.gabriel%40das-netzwerkteam.de.xfb

Ionic:
Ok. Will wait for your reply.

Btw; If you don't want to apply this PR to the 3.6.x branch I'm fine with that. But I will need to apply it to 3.5.0.x, so your experience as a DD is very welcome...

Ionic:
New approach with your input.

PR #28, opened 2015-05-20
Fix negotiation in stage 10 error

vatral:
Negotiation would sometimes fail during the connection stage, because writes to non-blocking sockets may fail with EAGAIN.

Problem fixed by adding a select() call to implement a timeout, and retrying writes if needed.

### DIFF
--- a/nxcomp/Loop.cpp
+++ b/nxcomp/Loop.cpp
@@ -8100,16 +8100,52 @@ int ReadRemoteData(int fd, char *buffer, int size, char stop)
 int WriteLocalData(int fd, const char *buffer, int size)
 {
   int position = 0;
+  int ret = 0;
+  fd_set writeSet;
+  struct timeval selectTs = {30, 0};
 
   while (position < size)
   {
+
+    // A write to a non-blocking socket may fail with EAGAIN. The problem is
+    // that cache data is done in several writes, and there's no easy way
+    // to handle failure without rewriting a significant amount of code.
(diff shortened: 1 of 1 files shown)


sunweaver:
@vatral: Thanks for submitting this patch. Is there any way to reproduce the "negotiation stage '10'" issue? From [1], I see that on some sort of (Rev)NAT connection the issue is likely to appear. Do you have a test case where an EAGAIN is reproducible? (I presume enforcing a busy network connection while starting a session could be the way to go here, right?)

sunweaver:
Forgot to mention [1]:

[1] https://mail.kde.org/pipermail/freenx-knx/2007-March/004834.html

sunweaver:
@nitomartinez, @vatral : I asked Nito to provide some info on how to trigger this bug or the context of introducing this fix. Please provide that info, so I can merge this PR.

vatral:
Hello, sorry for the late reply.

The error is unfortunately quite random and sometimes happens constantly and sometimes is hard to reproduce. I found it fairly easy to reproduce with our C client:

https://github.com/theqvd/theqvd/tree/master/libqvdclient

The error happens only during the negotiation stage, so if that goes well, just close the window and try again until it fails. When the bug is triggered it fails with a stage 10 error.

sunweaver:
@vatral: thanks for your feedback. I remember having it seen in X2Go a while back. I will merge this PR now.

Issue #29, opened 2015-05-21
CVE backports incomplete or wrong


sunweaver:
The below has been filed via X2Go BTS [1] by Ulrich Sibiller (@uli42)and should be fixed here (3.6.x branch) first before rebasing against 3.5.0.x.

"""
Recently a lot of CVE fixes have been added to nx-libs.

E.g.
debian/patches/1027-render-check-request-size-before-reading-it-CVE.full.patch
and
debian/patches/1028-render-unvalidated-lengths-in-Render-extn.-swap.full.patch
add missing checks to nx-X11/programs/Xserver/render/render.c.

However, there's a file called
nx-X11/programs/Xserver/hw/nxagent/NXrender.c which is derived from
render.c and in that file those checks are missing, too.

(I suspect the original render/render.c is not used at all in favour
of hw/nxagent/NXrender.c but I am not 100% sure here.)

If render.c is used a all (I am not sure) the patches should be
extended to also fix NXrender.c.
If render.c is not used it should be removed and the patches should be
applied to NXrender.c instead.

There might be more cases like this, I only picked this one as an example.
"""

[1] https://bugs.x2go.org/879

mikedep333:
I've started looking into this.
- Deleting `nx-X11/programs/Xserver/render/render.c` does break the build in 3.6.x because `nx-X11/programs/Xserver/Imakefile` builds it. However, that might just be leftover logic in the Imakefile.
- I looked at several of the `nx-X11/programs/Xserver/hw/nxagent/NX*` files. They all appear to be derived from files elsewhere in the source tree. Several of them are derived from other files under 
  `nx-X11/programs/Xserver/render/`.

mikedep333:
Update: It looks like the NX\* files do depend on some of the files under `nx-X11/programs/Xserver/render/`, such as `animcur.c`.

mikedep333:
I created branch pr/render-cleanup-and-cve-fixes for the cleanup and CVE fixes in the Render (XRender) code specifically. I have not submitted a PR yet.

mikedep333:
I have identified what files under nx-X11/programs/Xserver/hw/nxagent/ are duplicates/derivatives of other files. I am keeping track of it in [this spreadsheet.](https://docs.google.com/spreadsheets/d/1fc_4lBKb9o9iwMmrRl8ua-Hf6iJXsNeJJsx0OYJbeb4/edit?usp=sharing)
[...]

PR #35, opened 2015-05-23
XRender cleanup and CVE fixes

mikedep333:
This is one of the PRs to address #29.

I do not think it is ready to be merged yet because I do not understand why there are duplicate header files. (e.g., picturestr.h and NXpicturestr.h) If you delete either of the header files, the build fails.


mikedep333:
After discussion this PR with Ionic on IRC, we have agreed to split it up into multiple PRs. I am closing without merging.

PR #37, opened 2015-05-24
XRender - remove duplicate code

mikedep333:
This PR was split off from #35 

As explained in PR #29, there is duplicate XRender code. Much of it appears to be unused.

This PR is to cleanup the unused code.

I still need to check whether the duplicate headers can be removed (or at least replaced with a symlink.)

PR #38, opened 2015-05-24
Apply misc patches to duplicate XRender code

mikedep333:
This PR split off from #35 

This PR also depends on the outcome of PR #37. Once we have removed all the duplicate code that we possibly can remove, we need to apply any previous non-CVE patches/commits to the remaining duplicate code.

commit 0d56c45 was:
nx-X11: handle source pictures (those without a Drawable surface) gracefully.


sunweaver:
Not sure if this is the right approach... See https://github.com/ArcticaProject/nx-libs/pull/37#issuecomment-105443733

PR #39, opened 2015-05-26
library clean-up: Don't compile (i.e., remove) libNX_Xres.a anymore.

sunweaver:
The nx-X11/lib/Xres/ code does not get used anywhere in nx-libs.


Ionic:
Looks good, but one minor thing: `debian/rules` still references `rm -R debian/tmp/usr/include/$(DEB_BUILD_MULTIARCH)/nx/X11/extensions/XRes*.h` which is for once removing XResProto.h and previously also XRes.h (which is getting removed by that PR.)

So on Debian, not even the proto header is shipped? I don't know if that was intended from the beginning. It's definitely being used within `nx-X11/programs/Xserver`.

This is not a problem with your PR, but some general thing I discovered while going through this.

sunweaver:
@Ionic: I just rebased this branch against 3.6.x.

About the proto file removals in debian/rules. Until now, we only packaged the proto files relevant for the libnx-\* packages. The proto files used in the Xserver never got packaged in nx-libs, so far.

Honestly, I would love to avoid the workload of creating dozens of new bin:packages that are likely to be removed again in the long run and not of any use to others.

Ionic:
Yeah, packaging all nxagent protos as individual packages is insane.

PR #40, opened 2015-05-27
nx-X11/lib/GL cleanup

sunweaver:
The client-side library libNX_GL.{a,so} is not built when building nx-libs.

However, nx-X11/lib/GL/*\* ships several imake include files (Imakefile.inc) that are also used in nx-X11/programs/Xserver/GL/**. These files have been moved from the nx-X11/lib/GL/ code subtree to the nx-X11/programs/Xserver/GL/.

Furthermore, we don't provide module builds of the GL extension anymore, as that feature is neither used in nx-libs.

Last but not least, enforce build of Mesa GLX, remove build-logic for all other GLX variants (DRI, opengl, etc.).

**Note:** Please merge https://github.com/ArcticaProject/nx-libs/pull/39 before this one.

PR #41, opened 2015-05-27
imake-cleanup: Hardware driver cleanup.

sunweaver:
Drop all references to X11 hardware drivers from nx-libs's build-logic.

**Note**: Please merge https://github.com/ArcticaProject/nx-libs/pull/40 before this one.


nx-X11/config/cf/Imakefile
```diff
@@ -32,115 +31,24 @@ nxversion.def
 
 RULEFILES = \
 Imake.rules \
-Motif.rules \
-Win32.rules \
 X11.rules \
-bsdLib.rules \
```
Ionic: Are you sure we want to do that? I figure this will make it impossible to build nxagent on *BSD... it's probably not a too big deal because no one did so far, but do we really want to disable it completely?

Ionic: Deleting that... upon closer inspection, it turns out all these files have already been deleted anyway.

nx-X11/config/cf/xf86.tmpl
```diff
@@ -1,128 +0,0 @@
-XCOMM $Xorg: xf86.tmpl,v 1.3 2000/08/17 19:41:48 cpqbld Exp $
```
Ionic, line 1: `nx-X11/config/cf/Server.tmpl` still references `xf86.tmpl` (and has a `XFree86Version` define.)

Other places which use `XFree86Version`:
- `nx-X11/config/cf/Imakefile`
- `nx-X11/programs/Xserver/GL/mesa/X/Imakefile`
- `nx-X11/programs/Xserver/Imakefile`
- `nx-X11/programs/Xserver/include/Imakefile`
- `nx-X11/programs/Xserver/mi/Imakefile`
- `nx-X11/programs/Xserver/Xext/Imakefile`

Places with `XFree86VersionString`:
- `nx-X11/config/Imakefile`

nx-X11/config/cf/xf86site.def
```diff
@@ -1,479 +0,0 @@
...
- * uncomment the following
- *
-#define InstallEmptyHostDef
- */
-
-/*
- * Which servers to build.  There is now only one XFree86 server.
- * It can be disabled by the following.
- *
-#define XF86Server		NO
```
Ionic, line 44: Can also be removed in
-  `nx-X11/programs/Xserver/Xext/Imakefile`
- `nx-X11/programs/Xserver/xkb/Imakefile`

PR #46, opened 2015-05-31
Xext CVE Fixes

mikedep333:
This is one of the PRs to address #29.

I did a test build on Ubuntu 14.04 64-bit. It passed.

However, I am trying to find an application that uses the Xvideo extension to test this for regressions. I suppose I need to find an application that is linked (`ldd` command) against `/usr/lib/x86_64-linux-gnu/libXv.so`, but neither VLC nor Totem are. Neither are Firefox or Google Chrome (which both support HTML5 video.)

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/NXxvdisp.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/NXxvdisp.c
@@ -1423,6 +1423,7 @@ SProcXvQueryExtension(ClientPtr client)
 {
   register char n;
   REQUEST(xvQueryExtensionReq);
+  REQUEST_SIZE_MATCH(xvQueryExtensionReq);
   swaps(&stuff->length, n);
   return ProcXvQueryExtension(client);
 }
@@ -1432,6 +1433,7 @@ SProcXvQueryAdaptors(ClientPtr client)
 {
   register char n;
   REQUEST(xvQueryAdaptorsReq);
+  REQUEST_SIZE_MATCH(xvQueryAdaptorsReq);
   swaps(&stuff->length, n);
   swapl(&stuff->window, n);
   return ProcXvQueryAdaptors(client);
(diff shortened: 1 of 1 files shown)


mikedep333:
With all 3 of these versions of nx-libs, I am seeing output like this under an X2Go MATE session on my Ubuntu 14.04 64-bit server. The output implies that XVideo never worked, even though the extension is listed as present.
- 3.5.0.31
- 3.6.x without this commit applied
- 3.6.x with this commit applied

Note that I am using X2Go Client for Windows 4.0.4.0-2015.05.27

```
mike@dropship-hv:/gv-homes-AD/mike :) [1] $ xdpyinfo | grep -i xvideo
    XVideo
```

```
mike@dropship-hv:/gv-homes-AD/mike :) [2] $ xvinfo
X-Video Extension version 2.2
screen #0
 no adaptors present
```

Furthermore, `mplayer` uses XVideo. It produces output including these lines:

```
Xlib:  extension "XFree86-VidModeExtension" missing on display ":50.0".
Failed to open VDPAU backend libvdpau_nvidia.so: cannot open shared object file: No such file or directory
[vdpau] Error when calling vdp_device_create_x11: 1
[VO_XV] It seems there is no Xvideo support for your video card available.
[VO_XV] Run 'xvinfo' to verify its Xv support and read
[VO_XV] DOCS/HTML/en/video.html#xv!
[VO_XV] See 'mplayer -vo help' for other (non-xv) video out drivers.
[VO_XV] Try -vo x11.
```

`mplayer` later outputs

```
Reverting to normal Xlib
```

sunweaver:
@mikedep333: Currently the xf86vmode extension code does only get built for XF86Server or XorgServer [1]. It should be an interesting experiment to provide it in NXAGENT_SERVER (as it is present in Xephyr).

I will merge this PR and open a wishlist issue for investigating xf86vmode in nxagent builds.

[1] https://github.com/ArcticaProject/nx-libs/blob/3.6.x/nx-X11/programs/Xserver/Xext/Imakefile#L57

PR #50, opened 2015-06-08
Pr/backported fixes

uli42:
backports of some fixes from current x.org upstream (1.17).

### DIFF
--- a/nx-X11/programs/Xserver/dix/dispatch.c
+++ b/nx-X11/programs/Xserver/dix/dispatch.c
@@ -579,6 +579,7 @@ ProcGetWindowAttributes(register ClientPtr client)
 					   SecurityReadAccess);
     if (!pWin)
         return(BadWindow);
+    memset(&wa, 0, sizeof(xGetWindowAttributesReply));
     GetWindowAttributes(pWin, client, &wa);
     WriteReplyToClient(client, sizeof(xGetWindowAttributesReply), &wa);
     return(client->noClientException);
@@ -834,6 +835,7 @@ ProcGetGeometry(register ClientPtr client)
     xGetGeometryReply rep;
     int status;
 
+    memset(&rep, 0, sizeof(xGetGeometryReply));
     if ((status = GetGeometry(client, &rep)) != Success)
 	return status;
 
@@ -856,6 +858,7 @@ ProcQueryTree(register ClientPtr client)
 					   SecurityReadAccess);
     if (!pWin)
         return(BadWindow);
+    memset(&reply, 0, sizeof(xQueryTreeReply));
     reply.type = X_Reply;
     reply.root = WindowTable[pWin->drawable.pScreen->myNum]->drawable.id;
     reply.sequenceNumber = client->sequence;
@@ -909,6 +912,7 @@ ProcInternAtom(register ClientPtr client)
     if (atom != BAD_RESOURCE)
     {
 	xInternAtomReply reply;
+	memset(&reply, 0, sizeof(xInternAtomReply));
 	reply.type = X_Reply;
 	reply.length = 0;
 	reply.sequenceNumber = client->sequence;
@@ -932,6 +936,7 @@ ProcGetAtomName(register ClientPtr client)
     if ( (str = NameForAtom(stuff->id)) )
     {
 	len = strlen(str);
+	memset(&reply, 0, sizeof(xGetAtomNameReply));
 	reply.type = X_Reply;
 	reply.length = (len + 3) >> 2;
 	reply.sequenceNumber = client->sequence;
@@ -1061,6 +1066,7 @@ ProcGetSelectionOwner(register ClientPtr client)
 	i = 0;
         while ((i < NumCurrentSelections) && 
 	       CurrentSelections[i].selection != stuff->id) i++;
+	memset(&reply, 0, sizeof(xGetSelectionOwnerReply));
         reply.type = X_Reply;
 	reply.length = 0;
 	reply.sequenceNumber = client->sequence;
@@ -1112,6 +1118,7 @@ ProcConvertSelection(register ClientPtr client)
 #endif
 	    )
 	{        
+	    memset(&event, 0, sizeof(xEvent));
 	    event.u.u.type = SelectionRequest;
 	    event.u.selectionRequest.time = stuff->time;
 	    event.u.selectionRequest.owner = 
@@ -1125,6 +1132,7 @@ ProcConvertSelection(register ClientPtr client)
 		NoEventMask /* CantBeFiltered */, NullGrab))
 		return (client->noClientException);
 	}
+	memset(&event, 0, sizeof(xEvent));
 	event.u.u.type = SelectionNotify;
 	event.u.selectionNotify.time = stuff->time;
 	event.u.selectionNotify.requestor = stuff->requestor;
@@ -1221,6 +1229,7 @@ ProcTranslateCoords(register ClientPtr client)
 					   SecurityReadAccess);
     if (!pDst)
         return(BadWindow);
+    memset(&rep, 0, sizeof(xTranslateCoordsReply));
     rep.type = X_Reply;
     rep.length = 0;
     rep.sequenceNumber = client->sequence;
@@ -1370,6 +1379,7 @@ ProcQueryFont(register ClientPtr client)
 	    return(BadAlloc);
 	}
 
+	memset(reply, 0, rlength);
 	reply->type = X_Reply;
 	reply->length = (rlength - sizeof(xGenericReply)) >> 2;
 	reply->sequenceNumber = client->sequence;
@@ -2112,6 +2122,8 @@ DoGetImage(register ClientPtr client, int format, Drawable drawable,
         return(BadValue);
     }
     SECURITY_VERIFY_DRAWABLE(pDraw, drawable, client, SecurityReadAccess);
+
+    memset(&xgi, 0, sizeof(xGetImageReply));
     if(pDraw->type == DRAWABLE_WINDOW)
     {
       if( /* check for being viewable */
@@ -2165,7 +2177,7 @@ DoGetImage(register ClientPtr client, int format, Drawable drawable,
     xgi.length = length;
 
     if (im_return) {
-	pBuf = (char *)xalloc(sz_xGetImageReply + length);
+	pBuf = (char *)xcalloc(1, sz_xGetImageReply + length);
 	if (!pBuf)
 	    return (BadAlloc);
 	if (widthBytesLine == 0)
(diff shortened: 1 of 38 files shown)


Ionic:
While looking through your PR, especially

> Backport: RRModeCreate: plug memory leak of newModes if AddResource f…

I may have found two other memory leaks. Reported to alanc on IRC and waiting for discussion...

Ionic:
X-ref: http://lists.x.org/archives/xorg-devel/2015-June/046627.html

nx-X11/programs/Xserver/dix/dispatch.c
```diff
@@ -1370,6 +1379,7 @@ ProcQueryFont(register ClientPtr client)
 	    return(BadAlloc);
 	}
 
+	memset(reply, 0, rlength);
```
Ionic, line 1382: The original patch does

``` C
    reply = xcalloc(1, rlength);
```

You should probably use

``` C
    memset(reply, 1, rlength);
```

here?

nx-X11/programs/Xserver/dix/dispatch.c
```diff
@@ -2205,6 +2217,7 @@ DoGetImage(register ClientPtr client, int format, Drawable drawable,
 	}
 	if(!(pBuf = (char *) ALLOCATE_LOCAL(length)))
 	    return (BadAlloc);
+	memset(pBuf, 0, length);
```
Ionic: Also here.

nx-X11/programs/Xserver/dix/dispatch.c
```diff
@@ -2973,6 +2986,7 @@ ProcQueryColors(register ClientPtr client)
 	prgbs = (xrgb *)ALLOCATE_LOCAL(count * sizeof(xrgb));
 	if(!prgbs && count)
             return(BadAlloc);
+	memset(prgbs, 0, count * sizeof(xrgb));
```
Ionic: And here.

nx-X11/programs/Xserver/dix/dispatch.c
uli42, line 1382: No. xcalloc() is defined in include/os.h as

#define xcalloc(_num, _size) Xcalloc((unsigned long)(_num)*(unsigned long)(_size))

So xalloc(1, 5) will alloc 1*5 Bytes and clear them with zeros.

memset(..,1,5) however would fill the memory with 0x01 0x01 0x01 0x01 0x01

Ionic, line 1382: Oh, crap. Thanks. I remember having seen `xcalloc` before now. Yes, it's multiplying its arguments, not initializing with the first parameter. Sorry.

nx-X11/programs/Xserver/dix/window.c
```diff
@@ -2350,6 +2354,7 @@ ConfigureWindow(register WindowPtr pWin, register Mask mask, XID *vlist, ClientP
 	if (size_change && ((pWin->eventMask|wOtherEventMasks(pWin)) & ResizeRedirectMask))
 	{
 	    xEvent eventT;
+	    memset(&eventT, 0, sizeof(xEvent));
```
Ionic, line 2357: This should be also added to line 2404 (not with `eventT`, but `event`.)

(Re-adding this comment so that it's not removed if you rebase the commits.)

nx-X11/programs/Xserver/dix/window.c
```diff
@@ -2832,6 +2840,7 @@ MapSubwindows(register WindowPtr pParent, ClientPtr client)
 	    pWin->mapped = TRUE;
 	    if (parentNotify || StrSend(pWin))
 	    {
+	        memset(&event, 0, sizeof(xEvent));
```
Ionic: Should also be added to line 2831.

nx-X11/programs/Xserver/hw/nxagent/NXdispatch.c
```diff
@@ -1512,7 +1518,9 @@ ProcConvertSelection(register ClientPtr client)
 					CurrentSelections[i].pWin))
 #endif
 	    )
+
 	{        
+	    memset(&event, 0, sizeof(xEvent));
```
Ionic, line 1523: I'd just rebase this like you have done in `nx-X11/programs/Xserver/dix/dispatch.c`: move this up to line 1480, drop here and on line 1537.

nx-X11/programs/Xserver/hw/nxagent/NXdixfonts.c
```diff
@@ -929,6 +929,7 @@ doListFontsAndAliases(ClientPtr client, LFclosurePtr c)
     for (i = 0; i < nnames; i++)
 	stringLens += (names->length[i] <= 255) ? names->length[i] : 0;
 
+    memset(&reply, 0, sizeof(xListFontsReply));
     reply.type = X_Reply;
     reply.length = (stringLens + nnames + 3) >> 2;
     reply.nFonts = nnames;
```
Ionic, line 935: Second part from `nx-X11/programs/Xserver/dix/dixfonts.c` should also be backported to this file.

nx-X11/programs/Xserver/hw/nxagent/NXevents.c
```diff
@@ -4335,6 +4338,7 @@ ProcGrabKeyboard(ClientPtr client)
 			KeyPressMask | KeyReleaseMask, &rep.status);
     if (result != Success)
 	return result;
+    memset(&rep, 0, sizeof(xGrabKeyboardReply));
```
Ionic: Please move this up to line 4326, as done in `nx-X11/programs/Xserver/dix/events.c`.

nx-X11/programs/Xserver/dix/property.c
```diff
@@ -531,6 +531,7 @@ ProcGetProperty(ClientPtr client)
 	pProp = pProp->next;
     }
 
+    memset(&reply, 0, sizeof(xGetPropertyReply));
```
Ionic, line 534: Needs to be backported to `nx-X11/programs/Xserver/hw/nxagent/NXproperty.c`. Also, there are many other places in `NXproperty.c` where events are generated not correctly zero-initialized... maybe do them, too?

Ionic:
I think I'm through with this for now.

Please take a look at my other comments. Also, I'd like the patch I have committed to xorg-server upstream, iff it gets reviewed successfully. Once that's done, I'll ask you to please add this patch to your set.

nx-X11/programs/Xserver/dix/property.c
uli42, line 534: I think memset is only needed where a reply structure is reused. Upstream did not add it everywhere. Ports to NXproperty.c should be done of course. I can so that at the end of the month.

Von: "Mihai Moldovan" notifications@github.com
Gesendet: ‎15.‎06.‎2015 02:53
Cc: "uli42" uli42@gmx.de
Betreff: Re: [nx-libs] Pr/backported fixes (#50)

In nx-X11/programs/Xserver/dix/property.c:

nx-X11/programs/Xserver/hw/nxagent/NXdispatch.c
uli42, line 1523: I just had a quick glance at it: This code generates TWO events and therefore must clear the memory twice!

uli42, line 1523: I have instead changed nx-X11/programs/Xserver/dix/dispatch.c

nx-X11/programs/Xserver/hw/nxagent/NXevents.c
uli42: done

nx-X11/programs/Xserver/dix/window.c
uli42, line 2357: done

nx-X11/programs/Xserver/dix/window.c
uli42: done

nx-X11/programs/Xserver/hw/nxagent/NXdixfonts.c
uli42, line 935: done

nx-X11/programs/Xserver/dix/property.c
uli42, line 534: I have backported that single memset now, but not all the other stuff. dix/property.c does not have them either. So I consider this as done.

uli42:
I have updated the code to reflect most of your proposed changes except the new memory leak you identified. We could fix this but I'd prefer waiting for an upstream patch. Di you get any feedback?

Ionic:
Like I said before, I also want to wait for my patch to be accepted and pushed to the upstream repo.

Unfortunately, no reaction yet. Might take some time.

Ionic:
Uhm no, sorry, that was too overzealous. We don't want to change whitespace within the original code to make diffing against newer or original code more easy.

Only fixup whitespaces for lines your commits have touched. (I.e., the two lines I referenced.)

Also, please drop the whitespace cleanup commit (delete it out of the list of `git rebase -i HEAD~5`.)

Then create a new commit and `git push -f` to your branch.

I'll be merging your PR after that (will backport my commit myself because it hasn't been reviewed yet... looks like this might take half a year or so. I've only recently seen patches being included dating back to last year's September.)

Ionic:
Just to make sure you see this: I updated my last comment (I think github doesn't sent notification emails for changes to comments.)

It would be great if you could squash the (new) whitespace fix into the previous commit or amend the previous commit ("correct previous commit") directly. If you don't know how to do this, it's fine (that is, not mandatory.) Just create a new commit.

uli42:
Am 01.07.2015 um 00:48 schrieb Mihai Moldovan:

> Uhm no, sorry, that was too overzealous. We don't want to change whitespace within the original code
> to make diffing against newer or original code more easy.
> 
> Only fixup whitespaces for lines your commits have touched. (I.e., the two lines I referenced.)

Yeah, I was thinking about that, too. But the problem here is that the current state of the code is 
totally chaotic in this respect so fixing my two, three lines will not really help. And we will be 
faced with fixing the code structure anyway while proceeding to a newer Xorg as base. Upstream has 
restructured the whole code using a script that is part of the source.

> I'll be merging your PR after that (will backport my commit myself because it hasn't been reviewed
> yet... looks like this might take half a year or so. I've only recently seen patches being included
> dating back to last year's September.)

Ok, I'll try to fix that tonight.

Uli

uli42:
I have dropped the full whitespace cleanup and replaced it by a small whitespace cleanup that only affects patched lines. I merged that new commit and the "correct previous commit" commit into the "Avoid sending unitialized data" commit. I also found at least two bugs while doing this and fixed them also. Finally I have updated the branch on github using push --force. So you need to be careful if you have cloned this branch previously

uli42:
I have updated the repo. I have used push --force so be careful in
case you have cloned the repo.

Ionic:
>  I also found at least two bugs while doing this and fixed them also.

Uhm... which bugs? I can't really diff against a previous version, so have no idea what you did...

uli42:
Am 03.07.2015 um 01:37 schrieb Mihai Moldovan:

> ```
> I also found at least two bugs while doing this and fixed them also.
> ```
> 
> Uhm... which bugs? I can't really diff against a previous version, so have no idea what you did...

Don't bother, these were not real bugs but misspellings I had introduced...

Uli

PR #53, opened 2015-06-22
Feature/drop xext and xtst

woglinde:
Hi,

I had no problems removing Xtst and Xext. I could compile all stuff without errors. Proably we can drop more headers from X11/extensions, but for that we need to edit some more files to include the renamed headers from the present X11 stuff.


sunweaver:
Closing this issue. Xtst has been dropped, Xext is on the todo lists. Anyhow, the approach taken here is too simple and requires a second, third, fourth, ... glance.

PR #54, opened 2015-06-23
Replace "pointer" by "void *"

sunweaver:
While looking at libNX_Xtst and libNX_Xext  I stumbled over a type declaration named "pointer" (defined as "void *") that got removed all over the Xserver code in X.Org [1] for a good reason (dupilicate use of pointer as type and pointer as var name).

Rather than backporting the exact patch, I have used several rounds of sed magic + build tests + manual editings.

Note: Don't merge this PR before #57 has been merged.

[1] http://cgit.freedesktop.org/xorg/xserver/commit/?id=60014a4a98ff924ae7f6840781f768c1cc93bbab


sunweaver:
@Ionic: can you please review this one? Thanks.

PR #44, opened 2015-05-28
Various fixes / improvements for tarball creation.

sunweaver:
See individual commits.

Note: Don't merge this PR before https://github.com/ArcticaProject/nx-libs/pull/43 has been merged.

PR #57, opened 2015-06-23
Remove autotools cruft in nx-X11/lib/**

sunweaver:
Tidying up non-used autotools files in nx-X11/lib/**. Once we move back to autotools, we will pick up latest autotooling files from X.Org upstream. Nonetheless, I hope that we won't have to convert any NX_Xlib to autotools as we plan to remove them completely.


sunweaver:
Note: Don't merge this PR before #44 has been merged.

Ionic:
Removing `configure.ac` means we cannot regenerate `configure`. Are you sure about this?

`ChangeLog`, `NEWS`, `README` and `COPYING` aren't really autotools-related... so the name is a bit misleading, but it's okay.

sunweaver:
On  Do 25 Jun 2015 00:56:23 CEST, Mihai Moldovan wrote:

> Removing `configure.ac` means we cannot regenerate `configure`. Are  
> you sure about this?
> 
> `ChangeLog`, `NEWS`, `README` and `COPYING` aren't really  
> autotools-related... so the name is a bit misleading, but it's okay.

Yes they are autotools related, because autotools fail without those files.

Moreover, we use imake, so configure.ac and configure are not used at  
all for the Xlibs in nx-libs to build.

Furthermore, the nx-X11/lib dir is about to be emptied anyway. So I  

## consider this an intermediate clean-up step before complete removal.

DAS-NETZWERKTEAM
fon: +49 (1520) 1976 148

GnuPG Key ID 0x25771B31

freeBusy:
https://mail.das-netzwerkteam.de/freebusy/m.gabriel%40das-netzwerkteam.de.xfb

sunweaver:
@Ionic: I have found an issue with this PR. Will update in a minute. No merge actions before further notice from me, please. Thanks!

sunweaver:
@Ionic: You can now continue with this PR. Please note that I have purposefully removed the pkg-config files of various libraries for consistency reasons (X.Org 6.9 did not have those files, whereas X.Org 7.0 had them). Furthermore, we are removing the Xlibs step-by-step, so there is no reason to ship those .pc (pkg-config) files.

PR #68, opened 2016-01-01
Add instructions for building using Mock under Fedora/EPEL

mikedep333:
Comments please before I merge.

In particular:
1. I am adding a new section underneath the signature of @sunweaver . This seems awkward.
2. I am thinking of adding note like "you can use 3.6.x with X2Go, which is included in Fedora/EPEL, without following any extra steps." I do not want to promote X2Go specifically, but it seems appropriate due to its inclusion in Fedora/EPEL and its ease of setup/testing (a single x2goserver is easy to setup.) I also do not want people to think that they have to go through lots of extra work in order to upgrade nx-libs underneath X2Go; it's just a package upgrade.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,3 +66,29 @@ TheQVD developers via the project's Github issue tracker.
 
 thanks+light+love, 20150531
 Mike Gabriel <mike.gabriel@das-netzwerkteam.de>
+
+## Building Under Fedora or EPEL using Mock
+
+Assuming:
+
+1. The branch you are building is 3.6.x
+2. The current version is 3.5.99.0 (specified in the .spec file)
+3. The current release is 0.0build1 (specified in the .spec file)
+4. You wish for the RPM files and the mock build logs to be under ~/result
+
+Prerequisites:
+
+1. Install package "mock"
+2. Add your user account to the "mock" group (recommended)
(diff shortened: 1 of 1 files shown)


sunweaver:
1.
Please remove my signature and add "The nx-libs developers (a joined force of ArcticaProject / TheQVD / X2Go). Or such... (you are the native English speaker... ;-) ).

2.
nx-libs is a joined code project of people from X2Go, TheQVD, ArcticaProject. Mentioning any of those projects in the README is absolutely fine. And at the moment, either local nxagent calls (nxagent :1) or X2Go sesions are my test cases.

3.
Please note that since 3 days ago (29th Dec) nx-libs 3.6.x handles Xinerama stuff very differently from nx-libs 3.5.x. X2Go Client will fail to launch Xinerama sessions, unless the RRXinerama extension is enabled server-side though x2goagent.options. (add x2goagent option +xinerama to the command line). The new logic for enabling/disabling nxagent/x2goagent Xinerama still needs implementation in X2Go.

sunweaver:
Please also add a short-section that explains building on Debian:

```
git clone <url> nx-libs
cd nx-libs
debuild -uc -us
cd ..
sudo dpkg -i *3.5.99*.deb
```

Issue #101, opened 2016-03-15
Reduce Xserver/hw/nxagent compiler warnings


sunweaver:
When working with the nxagent code, the manifold compiler warnings are sort of disturbing and should be reduced to zero:

```
NXwindow.c: In function ‘nxagentClearSplash’:
NXwindow.c:342:12: warning: variable ‘h’ set but not used [-Wunused-but-set-variable]
     int w, h;
            ^
NXwindow.c:342:9: warning: variable ‘w’ set but not used [-Wunused-but-set-variable]
     int w, h;
         ^
In file included from ../../include/scrnintstr.h:54:0,
                 from NXwindow.c:105:
NXwindow.c: In function ‘CreateRootWindow’:
NXwindow.c:516:43: warning: the address of ‘box’ will always evaluate as ‘true’ [-Waddress]
     REGION_INIT(pScreen, &pWin->clipList, &box, 1);
                                           ^
../../include/regionstr.h:256:9: note: in definition of macro ‘REGION_INIT’
     if (_rect) \
         ^
NXwindow.c:517:42: warning: the address of ‘box’ will always evaluate as ‘true’ [-Waddress]
     REGION_INIT(pScreen, &pWin->winSize, &box, 1);
                                          ^
../../include/regionstr.h:256:9: note: in definition of macro ‘REGION_INIT’
     if (_rect) \
         ^
NXwindow.c:518:45: warning: the address of ‘box’ will always evaluate as ‘true’ [-Waddress]
     REGION_INIT(pScreen, &pWin->borderSize, &box, 1);
                                             ^
../../include/regionstr.h:256:9: note: in definition of macro ‘REGION_INIT’
     if (_rect) \
         ^
NXwindow.c:519:45: warning: the address of ‘box’ will always evaluate as ‘true’ [-Waddress]
     REGION_INIT(pScreen, &pWin->borderClip, &box, 1);
                                             ^
../../include/regionstr.h:256:9: note: in definition of macro ‘REGION_INIT’
     if (_rect) \
         ^
NXwindow.c: In function ‘ShapeOverlap’:
NXwindow.c:2161:24: warning: variable ‘pScreen’ set but not used [-Wunused-but-set-variable]
     register ScreenPtr pScreen;
                        ^
NXwindow.c: In function ‘WhereDoIGoInTheStack’:
NXwindow.c:2267:24: warning: variable ‘pScreen’ set but not used [-Wunused-but-set-variable]
[...]

PR #93, opened 2016-02-27
library-cleanup: Don't build libNX_Xrender anymore. Use system's libXrender shared library.

sunweaver:
Dropping NX's libXrender library also drops the XRenderCleanGlyphs symbol [1] from nx-libs. On first testings, no negative impact could be observed. Any feedback is highly welcome on this.

[1] https://github.com/ArcticaProject/nx-libs/blob/3.6.x/doc/nx-X11_vs_XOrg69_patches/nx-X11_lib_Xrender_Glyph.c.X.original

PR #102, opened 2016-03-22
Amend all issues generating compiler warnings in progams/Xserver/hw/nxagent/.

sunweaver:
Fixes ArcticaProject/nx-libs#101.

Before this PR can be reviewed the following PR needs reviewing:
https://github.com/ArcticaProject/nx-libs/pull/93


sunweaver:
@Ionic: Can you please review this PR? Thanks! Mike

nx-X11/programs/Xserver/hw/nxagent/Display.c
```diff
@@ -2014,6 +2018,7 @@ Bool nxagentMakeIcon(Display *display, Pixmap *nxIcon, Pixmap *nxMask)
      }
   }
 
+  xfree(agentIconData);
```
woglinde: Why did you use xfree here, instead of the normal free or is the memcpy also from xserver?

sunweaver: On  Di 22 Mär 2016 11:59:42 CET, Henning Heinold wrote:

> > @@ -2014,6 +2018,7 @@ Bool nxagentMakeIcon(Display *display, Pixmap  
> > *nxIcon, Pixmap *nxMask)
> >       }
> >    }
> > -  xfree(agentIconData);
> 
> Why did you use xfree here, instead of the normal free or is the  
> memcpy also from xserver?

Indeed. Using xfree is deprecated as of today. I will change this with  
a forced push.

Thanks for the first feedback.

Mike

## 

DAS-NETZWERKTEAM
fon: +49 (1520) 1976 148

GnuPG Key ID 0x25771B31

freeBusy:
https://mail.das-netzwerkteam.de/mailxchange/kronolith/fb.php?u=m.gabriel%40das-netzwerkteam.de

sunweaver:
@woglinde: I have checked the nxagent DDX code once more. The code still uses xfree() at various places. We should probably change that and follow X.Org devs on the change over from xfree() to free().

So for code consistency, I will leave xfree() in this patch set for now and we should remember to modify all xfree() calls later on.

woglinde:
How about to open an issue. So it will not be forgotten?

Done via #105

sunweaver:
Hi Henning,

On Tue Mar 22 21:18:50 2016 GMT+0100, Henning Heinold wrote:

> How about to open an issue. So it will not be forgotten?

Could you possibly do that? It would be good to have X.Org commits in that issue report as references and/or   human readable recipe for the required steps of changing the code. a

Thanks+Greets,
Mike

Mike

> —
> Reply to this email directly or view it on GitHub￼

## 

Sent from my Jolla

sunweaver:
@Ionic: whenever you get to reviewing nx-libs PRs. This is the first to review.

sunweaver:
@Ionic: actually, there is one PR that is older than this one:
https://github.com/ArcticaProject/nx-libs/pull/93

nx-X11/programs/Xserver/hw/nxagent/Screen.h
```diff
@@ -102,6 +102,8 @@ Bool nxagentResizeScreen(ScreenPtr pScreen, int width, int height,
 
 int nxagentChangeScreenConfig(int screen, int width, int height, int mmWidth, int mmHeight);
 
+int nxagentAdjustRandRXinerama(ScreenPtr pScreen);
```
Ionic, line 105: Okay, obsoletes https://github.com/ArcticaProject/nx-libs/pull/102/commits/5ef08822d57cba41c09452b5fce466c62948b157

nx-X11/programs/Xserver/hw/nxagent/NXmiexpose.c
```diff
@@ -710,6 +710,12 @@ int what;
     ChangeGCVal gcval[7];
     ChangeGCVal newValues [COUNT_BITS];
 
+    /* don't let these undefined before using them */
+    int t;
+    for (t=0;t<COUNT_BITS;t++) {
+	newValues[t].ptr = NULL;
+	newValues[t].val = 0;
+    }
```
Ionic: I think that instead of this loop, you could just replace the line

``` C
    ChangeGCVal newValues [COUNT_BITS];
```

with

``` C
    ChangeGCVal newValues [COUNT_BITS] = {{ 0 }};
```

Making use of the universal 0 initializer (two braces, because we're initializing an array [first brace] of structs [second brace] with a scalar as the first element [no more braces required].)

nx-X11/programs/Xserver/hw/nxagent/Holder.c
```diff
@@ -23,6 +23,7 @@
 #include "scrnintstr.h"
 #include "Agent.h"
 #define GC XlibGC
+#define Pixmap XlibPixmap
```
Ionic, line 26: Why do we need this? It doesn't seem to actually fix any compiler warning, as far as I've seen. I didn't test on x86, though, only x86_64.

nx-X11/programs/Xserver/hw/nxagent/Display.c
```diff
@@ -62,7 +62,11 @@ is" without express or implied warranty.
 #include "Init.h"
 #include "Args.h"
 #include "Image.h"
+
+#define Pixmap XlibPixmap
 #include "Icons.h"
+#undef Pixmap
```
Ionic, line 68: I didn't quite understand why `#define Pixmap ... #undef Pixmap` is necessary. I see the compile warnings on x86_64, which stem from "Pixmap" being replaced by "Pixmap64", but I haven't found out why exactly. In Agent.h, Pixmap is defined as Pixmap64, but later on undefined again, so it shouldn't come from that.

Not saying the fix is wrong, I just don't understand why "Pixmap" was replaced by "Pixmap64" in the first place.

Ionic:
Otherwise no further comments. I could merge this any time. Maybe you want to take the universal-0-initializer suggestion into account, but it's not mandatory.

sunweaver:
Hi Mihai,

On  Fr 13 Mai 2016 04:23:19 CEST, Mihai Moldovan wrote:

> Otherwise no further comments. I could merge this any time. Maybe  
> you want to take the universal-0-initializer suggestion into  
> account, but it's not mandatory.

This Agent.h beast is tricky.

Originally, it was designed to allow nxagent compilation in 32bit  
against a 64bit Xlib library. At least, that's what I understood from  
the comments in the code. That actually has been adopted from XNest.

Nowadays, that we partially use X.org's libraries it also serves the  
purpose of having Xlib related data types and Xserver related data  
types. The have the same name, but the data structures have diverged  
slightly (e.g. the GC structure).

So, when playing with data structures and types, you always have to  
keep in mind, what data structure you are presenting to included  
header files.

If you include Xlib headers, then you have to do that before including  
Agent.h. After Agent.h inclusion, all headers should stem from the  
Xserver.

If you need an exception for that above rule, you need to define /  
undefine some of the Xlib types before and after inclusion.

Hope this is somehow explained ok.

Also: I have adopted the {{0}} initialization now.

Please go ahead and merge.

Thanks,

## Mike

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

GnuPG Key ID 0x25771B31

freeBusy:
https://mail.das-netzwerkteam.de/mailxchange/kronolith/fb.php?u=m.gabriel%40das-netzwerkteam.de

Issue #108, opened 2016-03-30
Review nxcomp to drop all compatibility code for versions prior to 3.5.x


fcarvajaldev:
For clean-up purposes all compatibility code for nxcomp < 3.5.x should be dropped. Anyone who is still using nxcomp 3.4 and earlier, should really stop using that code and upgrade.

At this moment nxcomp has compatibility code for versions up to 1.5.0 (protocol step 6) and there are many blocks of conditional code depending on the protocol step value.

Setting a new compatibility limit at 3.5.x versions would imply working only with protocol step 10 (which also includes previous ones), so the conditional code depending on the protocol step value could be simplified.

Those blocks which are specific for an older protocol step can be removed and every block that conforms to protocol step values previous to the current value 10 and forward will remain in the code but now without any condition, as it is assumed that if protocol's version negotiation doesn't fail and exits, then both sides (local and remote) will be working with protocol step 10 (or with a higher one in the future).

sunweaver:
Thanks for looking into this.

One question: would it make sense to keep the step 10 condition checking (even if always resulting as true) as documentation of history? If the code becomes much more readable when dropping the if clauses, forget my question.

Another reason for keeping the conditionals could be future work on the protocol. If we know where we checked for conditions earlier, we maybe are able to more easily identify where new condition checks have to be put later on. Just theoretical thinking... without having looked at the specific code blocks, actually.

fcarvajaldev:
Both reasons seem reasonable but I see some drawbacks in keeping the conditionals.

The main problem is that keeping the conditionals (even if always resulting as true) would also require to keep the corresponding else branches just as they are right now, or at least with an error handling procedure to manage the case of working with an older protocol version (despite this should not really happen at all).

As I see it this would leave the code much more the same as it is right now and the clean-up goal wouldn't be accomplished. But not including those else branches would lead to inconsistent and incomplete code, possibly leading to confusion and errors when trying future work on the protocol.

So in my opinion the code becomes more readable when dropping the conditional clauses, so we can get rid of blocks, methods and classes which are no longer used. In such cases to provide some documentation of history my first intention is to leave a comment instead of the conditional, dropping the whole else branch which no longer makes sense.

For example, in ClientChannel::handleRead the conditional from line 1290:

```
          if (control -> isProtoStep7() == 1)
          {
            encodeBuffer.encodeFreeXidValue(GetULONG(inputMessage + 4, bigEndian_),
[...]

PR #111, opened 2016-04-12
Remove compatibility code for nxcomp before 3.5.0

fcarvajaldev:
As a part of nxcomp's clean-up process, compatibility with old protocol
versions has been set now at 3.5.0.

This commit removes compatibility code from nxcomp so the minimum supported
version changes from 1.5.0 (protocol step 6) to 3.5.0 (protocol step 10).

At this moment the removal is not complete and it will be followed by some
other commits, in order to get rid of several components and files that are
no longer used as they were only needed to work with old protocol versions.

Refs: ArcticaProject/nx-libs#108


sunweaver:
@ fcarvajaldev: Thanks for providing this patch.

As we don't support proto versions before 3.5.0 anymore after this patch has been applied, I think we should also handle those situations when older proxies try to connect with nxcomp 3.6.x based proxies.

From a quick look, I think this should be done in Loop.cpp: int ReadProxyVersion(int fd).

Can you add such a check / bail-out code to Loop.cpp and add that commit to this PR? Thanks.

sunweaver:
@fcarvajaldev: signing off the commits provided so far. Please let me know what you think about adding some bail-out code to Loop.cpp.

fcarvajaldev:
@sunweaver: After reviewing the code again and assuming than I'm not misunderstanding what you mean, I think that we are already handling the situation when an older proxy is trying to connect with an nxcomp 3.6.x based proxy.

I will try to summarize the negotiation process that takes place between both proxies in order to set the protocol version that they can both use in their nx communications.

Once the connection between the local proxy (server mode) and the remote proxy (client mode) has been established, the proxy in server mode sends its version and options to the other proxy (in client mode) that is waiting for this information. In both sides this is taking place from function "Loop.cpp:handleNegotiationInLoop", but with the server proxy in "stage_sending_proxy_options" and the client proxy in "stage_waiting_proxy_version".

Then the client proxy gets the server proxy version through "Loop.cpp:ReadProxyVersion" and uses "Loop.cpp: SetVersion" to agree to a negotiated proxy version which corresponds to the lower compatible version in both sides. Finally from this negotiated version it tries to set the protocol version that it can use in its nx communications with the other proxy (the one in server mode).

So if the server proxy is older than 3.5.0 and the client proxy is 3.6.x then this negotiation process will return -1 causing a fatal error, killing the client proxy and closing the connection which the server proxy (which tries to show a dialog).

On the contrary if the server proxy is 3.6.x and the client proxy is older than 3.5.0 (for example 1.5.0) then the negotiation process in the client proxy will succeed setting for itself a protocol version lower than 10 (step=6 in our example). But then the negotiation process reverses and the client proxy sends its version (1.5.0) to the server proxy (3.6.x), so when the server proxy calls "Loop.cpp:SetVersion" and tries to set the protocol version from the older proxy version (1.5.0 in this case), then the code in 3.6.x version will return -1 again causing a fatal error, killing the server proxy and closing the connection with the client proxy.

I will made some extra tests to be sure on this, but in my opinion actual code already handles the situation.

sunweaver:
Hi Fernando,

On  Mi 13 Apr 2016 14:08:23 CEST, fcarvajaldev wrote:

> @sunweaver: After reviewing the code again and assuming than I'm not  
> misunderstanding what you mean, I think that we are already handling  
> the situation when an older proxy is trying to connect with an  
> nxcomp 3.6.x based proxy.
> 
> I will try to summarize the negotiation process that takes place  
> between both proxies in order to set the protocol version that they  
> can both use in their nx communications.
> 
> Once the connection between the local proxy (server mode) and the  
> remote proxy (client mode) has been established, the proxy in server  
> mode sends its version and options to the other proxy (in client  
> mode) that is waiting for this information. In both sides this is  
> taking place from function "Loop.cpp:handleNegotiationInLoop", but  
> with the server proxy in "stage_sending_proxy_options" and the  
> client proxy in "stage_waiting_proxy_version".
> 
> Then the client proxy gets the server proxy version through
> "Loop.cpp:ReadProxyVersion" and uses "Loop.cpp: SetVersion" to agree  
> to a negotiated proxy version which corresponds to the lower  
> compatible version in both sides. Finally from this negotiated  
> version it tries to set the protocol version that it can use in its  
> nx communications with the other proxy (the one in server mode).
> 
> So if the server proxy is older than 3.5.0 and the client proxy is  
> 3.6.x then this negotiation process will return -1 causing a fatal  
> error, killing the client proxy and closing the connection which the  
> server proxy (which tries to show a dialog).
> 
> On the contrary if the server proxy is 3.6.x and the client proxy is  
> older than 3.5.0 (for example 1.5.0) then the negotiation process in  
> the client proxy will succeed setting for itself a protocol version  
> lower than 10 (step=6 in our example). But then the negotiation  
> process reverses and the client proxy sends its version (1.5.0) to  
> the server proxy (3.6.x), so when the server proxy calls  
> "Loop.cpp:SetVersion" and tries to set the protocol version from the  
> older proxy version (1.5.0 in this case), then the code in 3.6.x  
> version will return -1 again causing a fatal error, killing the  
> server proxy and closing the connection with the client proxy.
> 
> I will made some extra tests to be sure on this, but in my opinion  
> actual code already handles the situation.

Ok. If this really is the case, then I am fine with the changeset.  
Have you tested old proxy vs new proxy connecting each other with your  
changes applied to the new proxy?

Is the -1 return code theory or has it been tested. If not, please  
consider running some tests.

Thanks,

## Mike

DAS-NETZWERKTEAM
fon: +49 (1520) 1976 148

GnuPG Key ID 0x25771B31

freeBusy:
https://mail.das-netzwerkteam.de/mailxchange/kronolith/fb.php?u=m.gabriel%40das-netzwerkteam.de

fcarvajaldev:
Hi Mike.

The -1 return code has been tested but not exhaustively with every possible combination, so I'm going to run some additional tests with old proxy vs. new proxy in both sides, just to confirm the -1 return code in all the scenarios.

As I don't have any real proxy older than 3.5.0 for these tests, I'm going to simulate one by recompiling an nxcomp 3.5.x proxy after downgrading its version number to a lower value (for example 3.1.0), but with no further changes.

That should lead to an scenario where the fake 3.1.0 proxy tries to set the protocol version at 9 (when it acts in client mode), but the negotiation always fails when the 3.6.x proxy tries to manage the 3.1.0 proxy version from the other side.

Greets,
Fernando

sunweaver:
On  Mi 13 Apr 2016 15:58:09 CEST, fcarvajaldev wrote:

> Hi Mike.
> 
> The -1 return code has been tested but not exhaustively with every  
> possible combination, so I'm going to run some additional tests with  
> old proxy vs. new proxy in both sides, just to confirm the -1 return  
> code in all the scenarios.
> 
> As I don't have any real proxy older than 3.5.0 for these tests, I'm  
> going to simulate one by recompiling an nxcomp 3.5.x proxy after  
> downgrading its version number to a lower value (for example 3.1.0),  
> but with no further changes.
> 
> That should lead to an scenario where the fake 3.1.0 proxy tries to  
> set the protocol version at 9 (when it acts in client mode), but the  
> negotiation always fails when the 3.6.x proxy tries to manage the  
> 3.1.0 proxy version from the other side.

Ack. Thank you!

## Mike

DAS-NETZWERKTEAM
fon: +49 (1520) 1976 148

GnuPG Key ID 0x25771B31

freeBusy:
https://mail.das-netzwerkteam.de/mailxchange/kronolith/fb.php?u=m.gabriel%40das-netzwerkteam.de

fcarvajaldev:
Hi Mike,

After running a pair of symmetric tests with a fake 3.1.0 proxy vs. a nxcomp 3.6.x based proxy (explained in more detail in the pdf file attached to this comment), I think we can reasonably ensure that the new code will handle properly the situations when proxies older than 3.5.0 try to connect with nxcomp 3.6.x based proxies.

[Error_tests_for_pull_request_111.pdf](https://github.com/ArcticaProject/nx-libs/files/219211/Error_tests_for_pull_request_111.pdf)

And related to nx-libs versions I would like to ask you how do you manage the VERSION file and when should we change it. For now I have left it unchanged in the commits for this pull-request and I have just changed it in my local repository from 3.5.99.0 to 3.6.0.0 (and to 3.1.0.0 in the fake version) in the tests commented before.

Greets,
Fernando

sunweaver:
Hi Fernando,

On  Do 14 Apr 2016 15:55:36 CEST, fcarvajaldev wrote:

> Hi Mike,
> 
> After running a pair of symmetric tests with a fake 3.1.0 proxy vs.  
> a nxcomp 3.6.x based proxy (explained in more detail in the pdf file  
> attached to this comment), I think we can reasonably ensure that the  
> new code will handle properly the situations when proxies older than  
> 3.5.0 try to connect with nxcomp 3.6.x based proxies.
> 
> [Error_tests_for_pull_request_111.pdf](https://github.com/ArcticaProject/nx-libs/files/219211/Error_tests_for_pull_request_111.pdf)
> 
> And related to nx-libs versions I would like to ask you how do you  
> manage the VERSION file and when should we change it. For now I have  
> left it unchanged in the commits for this pull-request and I have  
> just changed it in my local repository from 3.5.99.0 to 3.6.0.0 (and  
> to 3.1.0.0 in the fake version) in the tests commented before.

I would like to change the VERSION file on the day that we release  
3.6.0.0. Is a change prior to this moment needed for your patch to  
work properly?

Regarding the upcoming merge of this PR....

Please find a meaningful name for that PDF file you created (e.g.  
"nxcomp-3.6-drops-compat-code-3.4.x.pdf (thanks for this documentation  
step!!!) and add the PDF with another commit to this PR (by simply  
pushing another commit to the branch where you initiated this PR  
from). Place the PDF file into <nx-libs>/doc/nxcomp/ (where  
doc/nxcomp/ is a new subfolder).

Thanks+Greets,
Mike

## 

DAS-NETZWERKTEAM
fon: +49 (1520) 1976 148

GnuPG Key ID 0x25771B31

freeBusy:
https://mail.das-netzwerkteam.de/mailxchange/kronolith/fb.php?u=m.gabriel%40das-netzwerkteam.de

fcarvajaldev:
Hi Mike,

> I would like to change the VERSION file on the day that we release 3.6.0.0. Is a change prior to this moment needed for your patch to work properly? 

No problem here with your timeline as the patch isn't affected by this value, as far as it's bigger than the new lower limit for the support of old protocol versions that is going to be fixed at 3.5.0 by this patch.

> Please find a meaningful name for that PDF file you created (e.g. "nxcomp-3.6-drops-compat-code-3.4.x.pdf (thanks for this documentation step!!!) and add the PDF with another commit to this PR (by simply pushing another commit to the branch where you initiated this PR from). Place the PDF file into <nx-libs>/doc/nxcomp/ (where doc/nxcomp/ is a new subfolder).

Maybe a meaningful name could be "nxcomp-3.6-drops-compat-code-3.4.x-testing.pdf" so if it's ok for you also I'll push the file and its new folder to the branch associated with this PR.

Greets,
Fernando

sunweaver:
Hi Fernando,

The pdf filename you proposed is fine. Please push.

Mike

On Thu Apr 14 16:52:06 2016 GMT+0200, fcarvajaldev wrote:

> Hi Mike, 
> I would like to change the VERSION file on the day that we release 3.6.0.0. Is a change prior to this moment needed for your patch to work properly? 
> No problem here with your timeline as the patch isn't affected by this value, as far as it's bigger than the new lower limit for the support of old protocol versions that is going to be fixed at 3.5.0 by this patch. 
> Please find a meaningful name for that PDF file you created (e.g. "nxcomp-3.6-drops-compat-code-3.4.x.pdf (thanks for this documentation step!!!) and add the PDF with another commit to this PR (by simply pushing another commit to the branch where you initiated this PR from). Place the PDF file into /doc/nxcomp/ (where doc/nxcomp/ is a new subfolder). 
> Maybe a meaningful name could be "nxcomp-3.6-drops-compat-code-3.4.x-testing.pdf" so if it's ok for you also I'll push the file and its new folder to the branch associated with this PR. 
> Greets,
> Fernando 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub￼

## 

Sent from my Jolla

sunweaver:
@fcarvajaldev: I saw that you added another commit before last weekend. Is this PR still work-in-progress? Or can we merge it?

fcarvajaldev:
@sunweaver: If your reviewing process has finished with no complains you can merge it whenever you think it's appropriate. Maybe in the future there will be further removals of classes or methods that were related with old protocol versions and are no longer needed, but for now I think we can consolidate this PR and a new one will be opened in the future if necessary.

sunweaver:
Hi Fernando,

On  Di 19 Apr 2016 10:14:14 CEST, fcarvajaldev wrote:

> @sunweaver: If your reviewing process has finished with no complains  
> you can merge it whenever you think it's appropriate. Maybe in the  
> future there will be further removals of classes or methods that  
> were related with old protocol versions and are no longer needed,  
> but for now I think we can consolidate this PR and a new one will be  
> opened in the future if necessary.

Ack. Thanks for feedback. Considering this PR as consolidated.

## Mike

DAS-NETZWERKTEAM
fon: +49 (1520) 1976 148

GnuPG Key ID 0x25771B31

freeBusy:
https://mail.das-netzwerkteam.de/mailxchange/kronolith/fb.php?u=m.gabriel%40das-netzwerkteam.de

PR #113, opened 2016-04-20
Fix copyright notice in nxcomp/Version.c

salva:
It was GPL3 instead of GPL2!


salva:
CC @sunweaver, @nitomartinez

uli42:
"a copy to Quindel" should read "a copy from Quindel"

sunweaver:
@salva: and the FSF address is not correct...

```
    You should have received a copy of the GNU General Public License along
    with this program; if not, write to the Free Software Foundation, Inc.,
    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
```

PR #114, opened 2016-04-20
nxcomp: unix sockets for channels

sunweaver:
With this change, nxcomp's channel model becomes able to use UNIX file sockets as channel end points on either end for all channel types (except font server channel).

### DIFF
--- a/nxcomp/ChannelEndPoint.cpp
+++ b/nxcomp/ChannelEndPoint.cpp
@@ -1,0 +1,245 @@
+/**************************************************************************/
+/*                                                                        */
+/* Copyright (c) 2015 Qindel Formacion y Servicios SL.                    */
+/*                                                                        */
+/* This program is free software; you can redistribute it and/or modify   */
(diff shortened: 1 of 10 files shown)


sunweaver:
@salva: Please review this PR and esp. take a close look at my manual patch application. The Loop.cpp.rej file is attached to this msg.

[Loop.cpp.rej.txt](https://github.com/ArcticaProject/nx-libs/files/228099/Loop.cpp.rej.txt)

sunweaver:
@salva: please review, but do not yet merge. Thanks.

salva:
ok!

The conflict was caused by 1be1c4a21bb54e60ec60456374d9ef55aaf55e2f which set the socket listening address to `INADDR_LOOPBACK` or `INADDR_LOOPBACK` in accordance to `lookbackBind`. That logic is now in the new `ListenConnectionTCP` function. So I think this reworked patch is right.

PR #125, opened 2016-05-27
Feature/composite update

Ionic:
Updates Composite extension to version 0.4 in X server parts - with a lot of rough edges.

### DIFF
--- a/nx-X11/programs/Xserver/composite/compalloc.c
+++ b/nx-X11/programs/Xserver/composite/compalloc.c
@@ -28,36 +28,110 @@
 
 #include "compint.h"
 
-void
-compReportDamage (DamagePtr pDamage, RegionPtr pRegion, void *closure)
+static void
+compScreenUpdate(ScreenPtr pScreen)
 {
-    WindowPtr	    pWin = (WindowPtr) closure;
-    ScreenPtr	    pScreen = pWin->drawable.pScreen;
-    CompScreenPtr   cs = GetCompScreen (pScreen);
-    CompWindowPtr   cw = GetCompWindow (pWin);
-
-    cs->damaged = TRUE;
-    cw->damaged = TRUE;
+    compCheckTree(pScreen);
+    compPaintChildrenToWindow(pScreen, WindowTable[pScreen->myNum]);
 }
 
 static void
-compDestroyDamage (DamagePtr pDamage, void *closure)
+compBlockHandler(int i, void *blockData, void *pTimeout, void *pReadmask)
 {
-    WindowPtr	    pWin = (WindowPtr) closure;
-    CompWindowPtr   cw = GetCompWindow (pWin);
+    ScreenPtr pScreen = screenInfo.screens[i];
+    CompScreenPtr cs = GetCompScreen(pScreen);
+
+    pScreen->BlockHandler = cs->BlockHandler;
+    compScreenUpdate(pScreen);
+    (*pScreen->BlockHandler) (i, blockData, pTimeout, pReadmask);
+
+    /* Next damage will restore the block handler */
+    cs->BlockHandler = NULL;
+}
+
+static void
+compReportDamage(DamagePtr pDamage, RegionPtr pRegion, void *closure)
+{
+    WindowPtr pWin = (WindowPtr) closure;
+    ScreenPtr pScreen = pWin->drawable.pScreen;
+    CompScreenPtr cs = GetCompScreen(pScreen);
+    CompWindowPtr cw = GetCompWindow(pWin);
+
+    if (!cs->BlockHandler) {
+        cs->BlockHandler = pScreen->BlockHandler;
+        pScreen->BlockHandler = compBlockHandler;
+    }
+    cs->damaged = TRUE;
+    cw->damaged = TRUE;
+
+    /* Mark the ancestors */
+    /* We can't do this, Dave. No damagedDescendants support. */
+    /*
+    pWin = pWin->parent;
+    while (pWin) {
+        if (pWin->damagedDescendants)
+            break;
+        pWin->damagedDescendants = TRUE;
+        pWin = pWin->parent;
+    }
+    */
+}
+
+static void
+compDestroyDamage(DamagePtr pDamage, void *closure)
+{
+    WindowPtr pWin = (WindowPtr) closure;
+    CompWindowPtr cw = GetCompWindow(pWin);
 
     cw->damage = 0;
+}
+
+static Bool
+compMarkWindows(WindowPtr pWin, WindowPtr *ppLayerWin)
+{
+    ScreenPtr pScreen = pWin->drawable.pScreen;
+    WindowPtr pLayerWin = pWin;
+
+    if (!pWin->viewable)
+        return FALSE;
+
+    (*pScreen->MarkOverlappedWindows) (pWin, pWin, &pLayerWin);
+    (*pScreen->MarkWindow) (pLayerWin->parent);
+
+    *ppLayerWin = pLayerWin;
+
+    return TRUE;
+}
+
+static void
+compHandleMarkedWindows(WindowPtr pWin, WindowPtr pLayerWin)
+{
+    ScreenPtr pScreen = pWin->drawable.pScreen;
+
+    (*pScreen->ValidateTree) (pLayerWin->parent, pLayerWin, VTOther);
+    (*pScreen->HandleExposures) (pLayerWin->parent);
+    if (pScreen->PostValidateTree)
+        (*pScreen->PostValidateTree) (pLayerWin->parent, pLayerWin, VTOther);
 }
 
 /*
  * Redirect one window for one client
  */
 int
-compRedirectWindow (ClientPtr pClient, WindowPtr pWin, int update)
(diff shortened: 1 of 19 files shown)


nx-X11/programs/Xserver/composite/compalloc.c
```diff
@@ -481,14 +633,21 @@ compAllocPixmap (WindowPtr pWin)
 	DamageRegister (&pWin->drawable, cw->damage);
 	cw->damageRegistered = TRUE;
     }
+
+    /* Make sure our borderClip is up to date */
+    REGION_UNINIT(pWin->drawable.pScreen, &cw->borderClip);
+    REGION_COPY(pWin->drawable.pScreen, &cw->borderClip, &pWin->borderClip);
```
sunweaver: REGION_UNINIT and REGION_COPY used to be macros, but are now inline functions (RegionUninit, RegionCopy). See https://github.com/ArcticaProject/nx-libs/commit/4a8d6eca395e1ec0490e20e7971177f688da18ec for details.

The old macro implementation has been kept as backwards compat mech, but we should really use the inline function calls here, I'd say.

nx-X11/programs/Xserver/composite/compext.c
```diff
@@ -27,18 +27,44 @@
...
+#endif
+
+#ifndef SERVER_COMPOSITE_MINOR_VERSION
+#define SERVER_COMPOSITE_MINOR_VERSION 4
+#endif
 
 static CARD8	CompositeReqCode;
-int		CompositeClientPrivateIndex;
+
+#ifndef NXAGENT_SERVER
```
sunweaver, line 44: I am not convinced that using NXAGENT_SERVER here is the best choice. The lack of the privates API is not directly related the NXAGENT_SERVER build mode. It is really to the underlying Xserver.

Maybe should rather define something like NX_XSERVER_LACKS_PRIVATES_API in the Imakefile and use that here. Once we fix that (i.e. add include/privates.h to the Xserver code), then we can undefine NX_XSERVER_LACKS_PRIVATES_API in the Imakefile and everything works out well immediately.

We will always have to keep NXAGENT_SERVER everywhere in the code, so esp. here the NXAGENT_SERVER actually would have a misleading meaning.

nx-X11/programs/Xserver/composite/compext.c
```diff
@@ -27,18 +27,44 @@
...
 RESTYPE		CompositeClientSubwindowsType;
+RESTYPE CompositeClientOverlayType;
 
 typedef struct _CompositeClient {
     int	    major_version;
     int	    minor_version;
 } CompositeClientRec, *CompositeClientPtr;
 
-#define GetCompositeClient(pClient)    ((CompositeClientPtr) (pClient)->devPrivates[CompositeClientPrivateIndex].ptr)
+#ifndef NXAGENT_SERVER
```
sunweaver, line 61: see above, NX_XSERVER_LACKS_PRIVATES_API

nx-X11/programs/Xserver/composite/compext.c
```diff
@@ -372,38 +607,476 @@ SProcCompositeDispatch (ClientPtr client)
...
 	return;
 
-    CompositeClientPrivateIndex = AllocateClientPrivateIndex ();
-    if (!AllocateClientPrivate (CompositeClientPrivateIndex, 
+    CompositeClientOverlayType = CreateNewResourceType
+        (FreeCompositeClientOverlay);
+    if (!CompositeClientOverlayType)
+        return;
+
+#ifndef NXAGENT_SERVER
```
sunweaver, line 688: see above, NX_XSERVER_LACKS_PRIVATES_API

nx-X11/programs/Xserver/composite/compinit.c
```diff
@@ -27,36 +27,57 @@
...
 
 #include "compint.h"
-
-int	CompScreenPrivateIndex;
-int	CompWindowPrivateIndex;
-int	CompSubwindowsPrivateIndex;
-int	CompGeneration;
+#include "compositeext.h"
+
+#ifndef NXAGENT_SERVER
```
sunweaver, line 32: see above, NX_XSERVER_LACKS_PRIVATES_API

nx-X11/programs/Xserver/composite/compinit.c
```diff
@@ -139,192 +218,190 @@ compFindVisuallessDepth (ScreenPtr pScreen, int d)
...
 
+/*
+ * Add a list of visual IDs to the list of visuals to implicitly redirect.
+ */
+static Bool
+compRegisterAlternateVisuals(CompScreenPtr cs, VisualID * vids, int nVisuals)
+{
+    VisualID *p;
+
+#ifndef NXAGENT_SERVER
```
sunweaver, line 229: Is this ifdef NX specific? Or actually related to another feature lacking in the Xserver code tree? I'd say let's only use NXAGENT_SERVER for changes to the Xserver code that have to do with better performance on remote connection and such stuff.

If we have to work around missing Xserver features, let's name that explicitly. Let's use one or more definitions where each definition stands for a lacking Xserver feature.

Whenever we get that feature in, we may simply undef that definition.

nx-X11/programs/Xserver/composite/compinit.c
```diff
@@ -139,192 +218,190 @@ compFindVisuallessDepth (ScreenPtr pScreen, int d)
...
-	CompScreenPrivateIndex = AllocateScreenPrivateIndex ();
-	if (CompScreenPrivateIndex == -1)
-	    return FALSE;
-	CompWindowPrivateIndex = AllocateWindowPrivateIndex ();
-	if (CompWindowPrivateIndex == -1)
-	    return FALSE;
-	CompSubwindowsPrivateIndex = AllocateWindowPrivateIndex ();
-	if (CompSubwindowsPrivateIndex == -1)
+#ifndef NXAGENT_SERVER
+    if (!dixRegisterPrivateKey(&CompScreenPrivateKeyRec, PRIVATE_SCREEN, 0))
```
sunweaver, line 385: see above, lack of privates API

nx-X11/programs/Xserver/composite/compwindow.c
```diff
@@ -524,6 +588,7 @@ compCopyWindow (WindowPtr pWin, DDXPointRec ptOldOrg, RegionPtr prgnSrc)
 		}
 		FreeScratchGC (pGC);
 	    }
+            REGION_UNINIT (pWin->drawable.pScreen, &rgnDst);
```
sunweaver: REGION_UNINIT -> use inline function RegionUninit() instead.

nx-X11/programs/Xserver/hw/nxagent/Composite.c
```diff
@@ -32,7 +32,7 @@
 
 #define PANIC
 #define WARNING
-#undef  TEST
+#define  TEST
```
sunweaver: Let's keep this undef'ed here, right?

sunweaver:
@Ionic: ok, I am through with commenting. Please give feedback on my comments or add follow-up commits that we (after a second review round) merge into commits where they best fit in.

Indeed, various rough edges. But let's get these changes in and fix occurring issues on the way.

nx-X11/programs/Xserver/composite/compext.c
Ionic, line 44: True, but it's not just the missing privates header, there's also stuff like Xace and generally a reworked Dix Permissions system (or the like.)

I used `NXAGENT_SERVER`, because when the randr extension was updated within nx-libs, the same macro was used to distinguish between the "new upstream implementation" and the "backported implementation", see for instance `nx-X11/programs/Xserver/randr/randr.c:77`.

I just went through with that whenever I needed to backport stuff that was incompatible with our old X.Org server base code.

nx-X11/programs/Xserver/composite/compinit.c
Ionic, line 229: Nope, just another missing feature in the X.Org base code again.

nx-X11/programs/Xserver/composite/compalloc.c
Ionic: Taken care of in https://github.com/ArcticaProject/nx-libs/pull/125/commits/d64eabf3d4b233bdbc10f40fa3df653d489101fc - I must have overlooked these when rebasing against 3.6.x. Most other places lead to a merge commit and were ported to `Region*` appropriately.

nx-X11/programs/Xserver/composite/compwindow.c
Ionic: Also in https://github.com/ArcticaProject/nx-libs/pull/125/commits/d64eabf3d4b233bdbc10f40fa3df653d489101fc.

nx-X11/programs/Xserver/hw/nxagent/Composite.c
Ionic: Not supposed to be committed, undone in https://github.com/ArcticaProject/nx-libs/pull/125/commits/1dfd07b26a6c61bd191ef30dfc6630d415740d35.

sunweaver:
@Ionic: please merge your corrections into the original commits where they correct something and force-push to the PR branch. Ping me on IRC once the PR branch has been updated.

PR #126, opened 2016-05-27
nx-X11/programs/Xserver/hw/nxagent/Handlers.h: fix stack smashing rel…

Ionic:
…ated to different data type sizes between Xserver and Xlib.

While on the X server side we were expecting a 32bit value, the Xlib side in
nxcompext wrote a 64bit integer to the 32bit location, hence overwriting
"random" data on the stack (most notably the return adress.)

We can work around this by conditionally using a 64bit data type on the
server-side as well, but this problem is likely a more general one.

Calling functions from libcompext in nxagent sounds like a bad idea. Other code
locations might well be affected by issues alike to that as well.

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Handlers.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Handlers.c
@@ -32,7 +32,10 @@
 #include "Screen.h"
 #include "Millis.h"
 
+#define Window XlibWindow
 #include <nx/NXlib.h>
+#undef Window
+
 #include <nx/Shadow.h>
 
 /*
@@ -956,7 +959,24 @@ void nxagentShadowWakeupHandler(void * data, int count, void * mask)
 
 void nxagentHandleCollectInputFocusEvent(int resource)
 {
-  Window window;
+  /*
+   * While we don't even need window or revert_to later on, a discrepancy in
+   * data type sizes between the X server (Window being a 32bit ID) and
+   * the Xlib (Window being a 64bit ID) will lead to stack corruption here.
+   * Calling functions from nxcompext from nxagent sounds like a very bad idea
+   * to begin with, but let's assume that's necessary for now and work around
+   * the corruption issue.
+   *
+   * Even though the NXlib header shows that the function expects a Window-sized
+   * parameter, it's not the Window type as defined and used within the X.Org
+   * Server, but an Xlib type. Hence, we'll be using the "XlibWindow" type here
+   * and to avoid compiler warnings, "rewrite" the NXlib.h header file via
+   * overriding the original "Window" type with the XlibWindow type, including
+   * the header file and undefining the macro again, essentially unshadowing
+   * the original type.
+   */
+  XlibWindow window;
+
   int revert_to;
 
   if (NXGetCollectedInputFocus(nxagentDisplay, resource, &window, &revert_to) == 0)


Ionic:
If this is merged before #125, I'll have to rebase #125 first.

nx-X11/programs/Xserver/hw/nxagent/Handlers.c
```diff
@@ -956,7 +956,20 @@ void nxagentShadowWakeupHandler(void * data, int count, void * mask)
...
+   * the Xlib (Window being a 64bit ID) will lead to stack corruption here.
+   * Calling functions from nxcompext from nxagent sounds like a very bad idea
+   * to begin with, but let's assume that's necessary for now and work around
+   * the corruption issue.
+   */
+#ifdef _XSERVER64
+  Window64 window;
+#else
   Window window;
+#endif /* defined (_XSERVER64) */
```
sunweaver: Why not use XlibWindow here as defined in hw/nxagent/Agent.h?

sunweaver: See also how that is handled in hw/nxagent/Events.c

Ionic: Then the NX code is inconsistent again... I've been looking mostly at `nx-X11/programs/Xserver/hw/nxagent/Colormap.c`, which uses `_XSERVER64` and switches between `Window` and `Window64`.

For our use case in this commit, it's probably better to use `XlibWindow`, I agree. Sadly stuff of Xlib\* types can't be passed down to other things within the X.Org Server code (exactly because the sizes might differ), but especially in this case the data is unused anyway.

nx-X11/programs/Xserver/hw/nxagent/Handlers.c
```diff
@@ -963,12 +963,16 @@ void nxagentHandleCollectInputFocusEvent(int resource)
...
    * the corruption issue.
+   *
+   * What we're doing here will lead to compile warnings when XlibWindow is
+   * actually a Window64, but that's "really" the type of "Window" libcompext
+   * expects.
+   *
+   * One way to maybe work around this issue is to define Window, XID and maybe
+   * Atom to their Xlib* types, include NXlib.h and then undefine these macros
+   * again. Might have unforseen consequences, though...
    */
```
Ionic, line 977: I've replaced the macro-condition with XlibWindow now. Please take a look at the updated comment, though. Is doing the workaround hack a viable way to silence the compile warnings on 64bit platforms? I really don't know what other consequences this might have, but generally libcompext is Xlib-based so all types in there should be "supposed" to be Xlib types.

sunweaver, line 977: @Ionic: the workaround you describe is actually the method that gets used in som places of the nxagent code. Please do that, rebase against 3.6.x, so that I can merge this tonight.

Thanks,
Mike

PS: Did you check for other occurrences of #include "NXlib.h" that may also need fixing???

uli42, line 977: I am wondering if all this is working correctly in 3.5.0...

Vom Smartphone gesendet.

Von: "Mike Gabriel" notifications@github.com
Gesendet: ‎04.‎06.‎2016 00:17
Betreff: Re: [ArcticaProject/nx-libs] nx-X11/programs/Xserver/hw/nxagent/Handlers.h: fix stack smashing rel… (#126)

In nx-X11/programs/Xserver/hw/nxagent/Handlers.c:

Ionic:
I pushed another commit doing the macro-hack to get rid of the compiler warnings. Testing on the X.Org 7.0 backport indicates that it should work fine. Does this look good to you?

If yes, I'll rebase the branch against current "master"/3.6 and optionally if you want it squash all commits into one. Tell me how to proceed.

@uli42 I honestly have no idea... the crash never happened with nx-libs 3.5.x or 3.6, but only with your X.Org-7.0-based backport, so something must be different. Most likely this code path was/is never taken in "vanilla" nx-libs 3.5 and 3.6. I suspect there are more problems alike to this one lingering around, like @sunweaver suspected as well.

sunweaver:
@Ionic: Please flatten the three commits into one, force-push and then I will merge.

Ionic:
Squashed, rebased against 3.6 and force-pushed.

Issue #117, opened 2016-04-25
Not able to disable Xinerama 


HolySephi:
While debugging another issue we encountered the problem that we were unable to turn off the new xinerama feature. Neither "-xinerama" nor "-extension XINERAMA" disables Xinerama.

Steps to reproduce:

`/usr/bin/nxagent -extension XINERAMA :1`

In any combination you are trying you get the same message from the new xinerama-Code:

```
nxagentAdjustRandRXinerama: Xinerama is disabled
nxagentAdjustRandRXinerama: faking xinerama
nxagentAdjustRandRXinerama: numCrtcs [0], numOutputs [0]
nxagentAdjustRandRXinerama: RRGetInfo returned [1]
nxagentAdjustRandRXinerama: bounding box: left [8] right [2888] top [213] bottom[1023]
nxagentAdjustRandRXinerama: numCrtcs [1], numOutputs [1]
nxagentAdjustRandRXinerama: numCrtcs [1], numOutputs [1]
nxagentAdjustRandRXinerama: destroying output [default]
nxagentAdjustRandRXinerama: created new output [NX1]
nxagentAdjustRandRXinerama: adjusting output [NX1]
nxagentAdjustRandRXinerama: output 0: prevmode [2880x810] ([0x1819200]) refcnt [2]
nxagentAdjustRandRXinerama: output 0: intersection is x [0] y [0] width [2880] height [810]
nxagentAdjustRandRXinerama: output 0: mode [nx_2880x810] ([0x1819360]) created/received, refcnt [2]
nxagentAdjustRandRXinerama: setting mode [nx_2880x810] ([0x1819360]) refcnt [2] for output 0
nxagentAdjustRandRXinerama: setting mode [nx_2880x810] ([0x1819360]) refcnt [2] for crtc 0
nxagentAdjustRandRXinerama: destroying prevmode [2880x810]
nxagentAdjustRandRXinerama: crtc 0 has mode [nx_2880x810] ([0x1819360]), refcnt [3] and 1 outputs
                            output[0]->crtc=[0x18184c0]
nxagentAdjustRandRXinerama: Min 320x240, Max 3840x1200
```

uli42:
On Mon, Apr 25, 2016 at 12:54 PM, Florian Wicke
notifications@github.com wrote:

> While debugging another issue we encountered the problem that we were unable
> to turn off the new xinerama feature. Neither "-xinerama" nor "-extension
> XINERAMA" disables Xinerama.

[...]

PR #134, opened 2016-06-09
Drop libXinerama support from nxagent entirely.

sunweaver:
- Only provide Xinerama support via RANDR's Xinerama implementation.
- Make RRXinerama configurable via Xinerama's cmdline options:
  - {+|-}xinerama (enable|disable RRXinerama in nxagent)
    (default: disabled)
  - nx/nx option: xinerama={0|1}

Fixes ArcticaProject/nx-libs#117.


sunweaver:
@uli42: Can you please review PR#134? Thanks.

uli42:
I am not sure if we should do this. If someone wants - for whatever reason - disable randr he will lose xinerama, too.

sunweaver:
Yes. That is correct. Does nx-X11/programs/Xserver's libXinerama based implementation actually query correct screen geometry?

sunweaver:
Have you tested both path ways? (1: Obtaining XineramaScreenInfo from libXinerama vs. 2: obtaining it via RANDR)?

sunweaver:
In the process of porting RANDR extension 1.5 to nx-libs (which makes mostly sense for API from 1.3), I played a bit with disabling RANDR. It does not make sense at all for desktops, because then the desktop sessions inside nxagent won't get notified about resize events of the nxagent root window. So whoever disables RANDR can probably live without Xinerama, as well.

sunweaver:
Plus: having one code path that is well tested and actually works is much preferred to a non-well-tested, but possible second code path.

uli42:
I have now read your comment in #117 about XINERAMA being broken by design. What do you mean by that? My first version of Xinerama-in-Nxagent was modifying XINERAMA and not RANDR and worked fine, AFAIR. 

Also, I was under the impression that disabling RANDR will cause NXagent to fall back to XINERAMA, not the other way round. But maybe im confusing something here.

Further: You mention libXinerama, which confuses me. I think you mean Xext/panoramix.c, right?

If you start a desktop session in fullscreen I'd say you do not need RANDR. But if you press CTRL-ALT-F in a running non-fullscreen desktop session you'd always need it. So disabling RANDR might be something which should not be supported at all.

sunweaver:
> I have now read your comment in #117 about XINERAMA being broken by design. What do you 
> mean by that?

Actually, I may be wrong with that. I made that statement by studying the code paths.

> My first version of Xinerama-in-Nxagent was modifying XINERAMA and not RANDR and worked fine,
> AFAIR. 

When talking about XINERAMA in this context, I mean how the configuration of the client-side screen setup is obtained. That can be obtained via libXinerama (XineramaQueryScreens) or via libXrandr (preferred approach at the moment).

> Also, I was under the impression that disabling RANDR will cause NXagent to fall back to 
> XINERAMA, not the other way round. But maybe im confusing something here.

With recent nx-libs it is/was not possible to disable Xinerama faking due to a flaw I introduced earlier. This fix amends that flaw:
https://github.com/ArcticaProject/nx-libs/pull/134/commits/711d05f8294919e8bcef8a1676edd3337ec8fc0e#diff-1a28065cd3388a1bbb1941601c759cb8L1120

> Further: You mention libXinerama, which confuses me. I think you mean Xext/panoramix.c, right?

No. The XineramaScreenInfo record (screeninfo in Screen.c) is obtained via XineramaQueryScreens from libXinerama (Xinerama.h). I would like to get rid of that.

> If you start a desktop session in fullscreen I'd say you do not need RANDR. But if you press
> CTRL-ALT-F in a running non-fullscreen desktop session you'd always need it. So disabling
> RANDR might be something which should not be supported at all.

RANDR is important. That is why I think that we can live with people disabling RANDR and loosing XINERAMA with that. Most setups will have RANDR running.

With this pull request, these achievements are made:

  o drop of build requirement libxinerama-dev
  o switch on / off rrXinerama through the named switches
  o drop a code path, we don't have to maintain

uli42:
Now I am even more puzzled! The XineramaQueryScreens() requests Xinerama information _from the real xserver_. Therefore nxagent is the client and must use libXinerama.

The code now looks like it gets the screeninfo information from the nxagent itself. But this is the wrong source. Somehow the data of the real x server must get into nxagent!

sunweaver:
@uli42: Ignore me for now. Looking into it all over again.

Issue #92, opened 2016-02-24
[abrt] nxagent: nxagentDispatchEvents(): nxagent killed by SIGSEGV


sunweaver:
From X2Go BTS:
http://bugs.x2go.org/cgi-bin/bugreport.cgi?bug=791

Package: nx-libs
Version: 3.5.0.28

Events.c:1251 is:

x.u.u.detail = inputInfo.pointer -> button ->
map[nxagentReversePointerMap[X.xbutton.button - 1]];

X.xbutton.button = 0 :

 X = {type = 4, xany = {type = 4, serial = 0, send_event = 0, display =
0x159ba50, window = 0}, xkey = {type = 4, serial = 0, send_event = 0, display
= 0x159ba50, window = 0, root = 7848854, subwindow = 11409376, time = 0, x =
-772, y = -13337, x_root = 32767, y_root = 0, state = 57, keycode = 0,
same_screen = 0}, xbutton = {type = 4, serial = 0, send_event = 0, display =
0x159ba50, window = 0, root = 7848854, subwindow = 11409376, time = 0, x =
-772, y = -13337, x_root = 32767, y_root = 0, state = 57, button = 0,
same_screen = 0}

-------- Forwarded Message --------
Subject: [Bug 1193943] New: [abrt] nxagent: nxagentDispatchEvents(): nxagent
killed by SIGSEGV
Date: Wed, 18 Feb 2015 15:30:49 +0000
From: bugzilla@redhat.com
To: orion@cora.nwra.com

https://bugzilla.redhat.com/show_bug.cgi?id=1193943

```
        Bug ID: 1193943
       Summary: [abrt] nxagent: nxagentDispatchEvents(): nxagent
                killed by SIGSEGV
       Product: Fedora
       Version: 20
     Component: nx-libs
      Assignee: orion@cora.nwra.com
      Reporter: unixi@freeshell.org
    QA Contact: extras-qa@fedoraproject.org
            CC: orion@cora.nwra.com
```

Version-Release number of selected component:
nxagent-3.5.0.28-1.fc20

Additional info:
reporter:       libreport-2.2.3
backtrace_rating: 4
cmdline:        /usr/lib64/nx/bin/nxagent -persistent -D -name 'NX -
jboyle@mnich.myquotient.net:2000 - mnich (GPL Edition)' -option
/home/jboyle/.nx/C-mnich.myquotient.net-2000-9C1FED17795787217FBC80B6F1F2E86E/options
[...]

Issue #118, opened 2016-04-25
nxagent segfaults on session resume


HolySephi:
Our environment:
- Ubuntu 14.04
- X2Go from PPA/X2Go
- Arctica from `deb http://packages.arctica-project.org/ubuntu-nightly trusty main`

nxagent segfaults reproducable after a few resumes with the same message every time.

Steps to reproduce:
1. `/usr/bin/nxagent :1`
2. get the pid of the nxagent process
3. `while :; do kill -SIGHUP <pid>; sleep 10; done`
     -> the segfault appears faster with sleep 3 - but even with 10 seconds (to give the process enough time to finish every action of resuming) the bug is appearing
4. Wait for the segmentation fault:

nxagent output:

```
Session: Resuming session at 'Mon Apr 25 13:00:27 2016'.
nxagentSetScreenSaverTime: ScreenSaverTime was [600000], ScreenSaverInterval was [600000].
nxagentSetScreenSaverTime: ScreenSaverTime now is [600000], ScreenSaverInterval now is [600000].
nxagentOpenScreen: Called for screen index [0].
nxagentOpenScreen: Root window at offset (0,0) size (2880,810).
nxagentOpenScreen: Default window at offset (2,139) size (2880,810) border size 0.
nxagentOpenScreen: Span between root window and default window is (0,0).
nxagentOpenScreen: Default window in window mode has offset (0,0) and size (2880,810).
nxagentOpenScreen: Fullscreen is OFF.
nxagentOpenScreen: Desktop resize mode is ON.
nxagentOpenScreen: Resize desktop at startup is OFF.
nxagentOpenScreen: Going to create new default window.
nxagentOpenScreen: Created new default window with id [8388686].
nxagentOpenScreen: Setting WM_CLASS and WM_NAME for window withid [8388686].
nxagentOpenScreen: Found WM, delete window atom [240].
Info: Not using local device configuration changes.
nxagentSaveScreen: Called for screen at [0x17aed50] with parameter [3].
nxagentSaveScreen: SCREEN_SAVER_ON is [0] SCREEN_SAVER_OFF is [1] SCREEN_SAVER_FORCER is [2] SCREEN_SAVER_CYCLE is [3].
SessionPath not defined
The XKEYBOARD keymap compiler (xkbcomp) reports:
> Error:            Cannot open "compiled/server-1.xkm" to write keyboard description
>                   Exiting
Session: Session resumed at 'Mon Apr 25 13:00:29 2016'.
[...]

PR #138, opened 2016-06-13
nxagentReversePointerMap: Don't assume Xevent.xbutton.button is alway…

sunweaver:
…s greater than zero. We have seen Xevent.xbutton.button being set to AnyButton (equaling 0).

Fixes ArcticaProject/nx-libs#92
Fixes X2GoBTS#791

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Events.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Events.c
@@ -1252,7 +1252,7 @@ FIXME: Don't enqueue the KeyRelease event if the key was
                     X.xbutton.subwindow == None))
         {
           x.u.u.type = ButtonPress;
-          x.u.u.detail = inputInfo.pointer -> button -> map[nxagentReversePointerMap[X.xbutton.button - 1]];
+          x.u.u.detail = inputInfo.pointer -> button -> map[nxagentReversePointerMap[X.xbutton.button]];
           x.u.keyButtonPointer.time = nxagentLastEventTime = GetTimeInMillis();
 
           if (nxagentOption(Rootless))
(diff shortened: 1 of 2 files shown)


sunweaver:
@opoplawski: I finally got to looking into issue ArcticaProject/nx-libs#92 (X2GoBTS#791). Do you have time and resources to review my change? Thanks in advance.

uli42:
I don't now about that ReversePointerMap but your change uses another offset now. Is this really equivalent?

sunweaver:
The offset is changed everywhere in nxagent afaict. Don't you think? The nxagentReversePointerMap is an internal mapping between nxagent's "buttons" and the underlying Xserver's "buttons". I changed the offset at all occurring locations.

uli42:
On Mon, Jun 13, 2016 at 1:20 PM, Mike Gabriel notifications@github.com wrote:

> The offset is changed everywhere in nxagent afaict. Don't you think? The nxagentReversePointerMap is an internal mapping between nxagent's "buttons" and the underlying Xserver's "buttons". I changed the offset at all occurring locations.

Right, I have overlooked that you also changed the part where data is
actually written.

BTW: I have not been aware of this bug. But I occassionally have seen
exactly this crash when trying to debug issue #118

sunweaver:
Hi Uli,

On Mon Jun 13 13:35:09 2016 GMT+0200, Ulrich Sibiller wrote:

> On Mon, Jun 13, 2016 at 1:20 PM, Mike Gabriel notifications@github.com wrote:
> 
> > The offset is changed everywhere in nxagent afaict. Don't you think? The nxagentReversePointerMap is an internal mapping between nxagent's "buttons" and the underlying Xserver's "buttons". I changed the offset at all occurring locations.
> 
> Right, I have overlooked that you also changed the part where data is
> actually written.
> 
> BTW: I have not been aware of this bug. But I occassionally have seen
> exactly this crash when trying to debug issue #118
> 
> —
> Assuming this is an approval?
> Mike

## 

Sent from my Jolla

uli42:
yes, this is an approval

PR #142, opened 2016-06-15
Provide feature of proxy-to-(proxy|agent)-communication over a unix file socket.

sunweaver:
This PR adds a long awaited feature in nx-libs. Communication of the proxy endpoints (nxproxy <-> nxproxy, nxproxy <-> nxagent) over Unix file sockets.

Note for the reviewer, there are some more todos on my list for this PR. All todos will be added on top of the current commit history.

ToDos:
- [DONE] remove <socket> file after nxagent/nxproxy have exited (Update: socket file gets removed after connection has been established!!!)
- [DONE] reduce code duplication in ChannelEndPoint.cpp due to "forking" our own code
  in Socket.cpp (IsPort, GetUnixSocketPath, GetComputerName, GetTCPHostAndPort) -> Moved over to use ChannelEndPoint class in the first place.
- [POSTPONED] allow "listen=<specific-ip>", so we can have IP address specific binding when waiting
  for an incoming connection (-> not in the scope of this PR)

### DIFF
--- a/nxcomp/ChannelEndPoint.cpp
+++ b/nxcomp/ChannelEndPoint.cpp
@@ -24,6 +24,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <stdio.h>
+#include <sys/stat.h>
 
 #include "ChannelEndPoint.h"
 
@@ -31,8 +32,21 @@
 
 ChannelEndPoint::ChannelEndPoint(const char *spec)
   : defaultTCPPort_(0), defaultTCPInterface_(0),
-    defaultUnixPath_(NULL) {
-  spec_ = (spec ? strdup(spec) : NULL);
+    defaultUnixPath_(NULL), spec_(NULL) {
+  setSpec(spec);
+}
+
+ChannelEndPoint::~ChannelEndPoint()
+{
+  char *unixPath = NULL;
+
+  if (getUnixPath(&unixPath))
+  {
+    struct stat st;
+    lstat(unixPath, &st);
+    if(S_ISSOCK(st.st_mode))
+      unlink(unixPath);
+  }
 }
 
 void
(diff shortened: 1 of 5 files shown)


nxcomp/Socket.cpp
```diff
@@ -743,3 +745,61 @@ int GetHostAddress(const char *name)
     return (*((int *) host -> h_addr_list[0]));
   }
 }
+
+bool IsPort(char* value, long *port)
```
salva: Outside of its original class context, the names `IsPort` and `GetUnixSocketPath` are not very informative. For instance, `ParseArgAsPortNumber` and `ParseArgAsUnixSocketPath`would be better names.

In any case, is there any reason to not just use a `ChannelEndPoint` object to represent the listening address?

sunweaver: Agreeing on the function names. Will check usage of ChannelEndPoint class.

sunweaver:
@salva: I have fixed the proxy-to-proxy communication over Unix sockets for nxcomp now. I have not changed the way, data gets stored in the spec_ property of ChannelEndPoint (as earlier discussed on IRC). Please review my changes and give feedback. Thanks!!!

nxcomp/ChannelEndPoint.cpp
```diff
@@ -24,37 +24,126 @@
...
+  int length;
+
+  if (spec_) free(spec_);
+  isUnix_ = false;
+  isTCP_ = false;
+
+  if (hostName && strlen(hostName) && port >= 1)
+  {
+    length = snprintf(NULL, 0, "tcp:%s:%ld", hostName, port);
+    newSocketUri = (char *)calloc(length + 1, sizeof(char));
```
salva: Any reason to use the intermediate variable `newSocketUri` here instead of just `_spec`?

nxcomp/Loop.cpp
```diff
@@ -6702,12 +6761,20 @@ int WaitForRemote(int portNum)
...
-      socklen_t addrLen = sizeof(sockaddr_in);
 
-      newFD = accept(proxyFD, (sockaddr *) &newAddr, &addrLen);
+      sockaddr_un newAddrUnix;
+      sockaddr_in newAddrINET;
 
+      if (socketAddress.isUnixSocket())
+      {
+        socklen_t addrLen = sizeof(sockaddr_un);
+        newFD = accept(proxyFD, (sockaddr *) &newAddrUnix, &addrLen);
```
salva: It seems you are not using `newAddrUnix` afterwards, so you could just pass NULL as the second `accept` argument here.

nxcomp/ChannelEndPoint.cpp
sunweaver: > Any reason to use the intermediate variable newSocketUri here instead of just _spec?

Valid point. Commit has been adapted as proposed.

nxcomp/Loop.cpp
sunweaver: > It seems you are not using newAddrUnix afterwards, so you could just pass NULL as the second 
> accept argument here.

Indeed, implementation changed as proposed. Thanks for spotting this.

sunweaver:
Patchset approved by @salva on #arctica IRC:

```
Day changed to 04 Jul 2016
[...]
16:30 < salv0> sunweaver: I have made a couple of comments on the GitHub patch view: 
               https://github.com/ArcticaProject/nx-libs/pull/142/commits/d62e3721dd77406010cb46335efedf94e2ec4a80
16:31 < salv0> but overall I am ok with your changes
```

Issue #132, opened 2016-06-02
nxagent a) hangs for long periods when no client b) overly strict about reconnection


mbecroft:
a) Currently, nxagent has deliberate delays inserted in several places when no client is connected. This is clearly designed to mitigate things like Flash web sites hogging the CPU when no user is connected to a session. However it has the following (and probably other) unintended side-effects:
- Depending on the system configuration, the long delays in responding to X protocol requests can lead to buffers filling up and ssh sessions becoming unresponsive or disconnected while the user is disconnected.
- It is impossible to use a client such as x11vnc to access the nxagent's display while the user is disconnected.
- Unexpected behaviour with any X client software displaying on multiple displays, when the user is disconnected. Example, start emacs in nxagent. make-frame-on display <some other display>. Disconnect from your nxagent session. emacs may become unresponsive even on the non-nxagent display.
- Unexpected behaviour from any X11 client software that has any other semi-real-time I/O requirements (processing, writing log files etc.) while a user is logged out from the nxagent session.

b) Currently, nxagent is excessively pedantic about the displays it will allow itself to be reconnected to. This has even resulted in the situation where it became impossible to reconnect to an nxagent from the original workstation from which it was created! Some of the checks being carried out are necessary, but some of them definitely appear to serve no purpose. Another very silly example is that nxagent refuses to reconnect to an X server that has _more_ pixmap depths than were available on the server where it was created. All of the server capability checks on reconnection need thorough review.

mbecroft:
The attached patch is hack workaround for these issues that got things going for me. use it only as a template to find the relevant sections of code requiring attention.

_I have done:_

a) comment out all deliberate delays in response to X protocol requests with `#ifdef MARIO_ENABLE`

b) Short-circuit several dubious compatibility checks by adding `return 1` (meaning "target X server is compatible") near the top.

_To be done:_

a) replace all instances of deliberate delays in response to nxagent client X protocol requests with a configurable delay (rather than the current fixed delay), and if that delay is configured to 0, then do not even call `usleep()` at all.

b) Closely evaluate the purpose and need of all the reconnection checks and rework these to correctly identify the scenarios where: 1) reconnection can be allowed with no caveats 2) reconnection can be allowed but some resources might not be available leading to non-fatal consequences (e.g. a pixmap depth is not available, but no client is using it), 3) reconnection cannot proceed because to do so would definitely fail immediately.

[...]

PR #143, opened 2016-06-17
Make the sleeping slowdown configurable.

Ionic:
As part of the first enhancement in https://github.com/ArcticaProject/nx-libs/issues/132, we want to make optional sleep delays adjustable.

This PR handles this.

Sleep delays are given in milliseconds from 0 (disable slowdowns completely) to UINT_MAX (although high values will probably not be useful.)

If necessary, we can also change this to take values in microseconds.

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -1339,6 +1339,44 @@ static void nxagentParseOptions(char *name, char *value)
       nxagentChangeOption(Clipboard, ClipboardBoth);
     }
   }
+  else if (!strcmp(name, "sleep"))
+  {
+    long sleep_parse = 0;
+
+    errno = 0;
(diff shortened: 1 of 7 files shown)


nx-X11/programs/Xserver/hw/nxagent/Display.c
```diff
@@ -2169,6 +2169,7 @@ static int nxagentCheckForDefaultDepthCompatibility()
 {
   int dDepth;
 
+  return 1;
```
sunweaver: This and the below 'return's addressing deactivations of color depth checks should not be part of this commit, IMHO.

I wonder if Mario's original diff (the first commit of this PR) should appear in the nx-libs history, at all.

sunweaver: not be part of this PR..., I mean...

Ionic:
Rebased and force-pushed.

PR #146, opened 2016-06-22
Drop embedded imake and makedepend

sunweaver:
Instead of using nx-libs's embedded imake and makedepend tool, we switch over to using the tool as found in the xutils-dev package.


sunweaver:
@vatral: Here is your next review. It would be gourgeous, if you could go over this PR within the next hours (ideally give feedback before 2pm). Thanks.

vatral:
Looks good to me

PR #150, opened 2016-06-23
 build flags: Use -D_DEFAULT_SOURCE instead of deprecated -D_BSD_SOURCE and -D_SVID_SOURCE.

sunweaver:
build flags: Use -D_DEFAULT_SOURCE instead of deprecated -D_BSD_SOURCE and -D_SVID_SOURCE.

The deprecation warnings are thrown up in Debian stretch.


vatral:
Looks good, matches documentation, compiles. But it looks like this breaks compatibility with Ubuntu 10.04, which we still support. Its glibc version doesn't recognize _DEFAULT_SOURCE

PR #152, opened 2016-06-23
nxauth/libXau clean-up: Drop NX_Xau and nxauth executable from nx-X11.

sunweaver:
The X.org xauth utility should be used instead.

Some libNX_Xau6 have been kept (copied into nx-X11/lib/X11). Also a reduced version of the Xauth.h headers file.


sunweaver:
```
14:51 < sunweaver> vtq: can you already estimate if you will get through with the libXau-cleanup review?
14:55 < vtq> sunweaver: I've taken a look at it, and it looks good to me. Compiles fine
14:55 < sunweaver> vtq: do you feel it has removed all remnants of libXau? Esp. in the Imakefiles?
14:58 < vtq> There's a lot of 'Xau' in the tree, so that would take a while to properly look over, but I do see: ./nx-X11/programs/Xserver/os/Imakefile:    
             -I$(TOP)/lib/Xau
14:58 < sunweaver> ah...
14:59 < sunweaver> vtq: you may use some random grepping...
14:59 < sunweaver> I have done that, but you may come up with other searches that help cleaning up more stuff.
15:00 < sunweaver> including libXau here and there is actually ok.
15:00 < sunweaver> including libNX_Xau would be a flaw...
15:00 < vtq> there also are remains in debian/copyright.in
15:02 < sunweaver> yeah, that will be machine generated later on.
15:02 < sunweaver> actually, I forgot to cleanup nx-X11/config completely this time.
15:02 < sunweaver> Doing that now.
```

@vatral: The nx-X11/config/cf/\* files have been cleaned-up. Assuming the above as an ok. Thanks for taking your time!!!

PR #160, opened 2016-06-27
Introduce Xserver's protocols-version.h file as done in X.org

sunweaver:
The X.org Xserver has one file specifying the proto versions of provided Xserver extensions. Immitating this for nx-libs's Xserver (aka nxagent).

### DIFF
--- a/nx-X11/programs/Xserver/damageext/damageext.c
+++ b/nx-X11/programs/Xserver/damageext/damageext.c
@@ -27,6 +27,7 @@
 #endif
 
 #include "damageextint.h"
+#include "protocol-versions.h"
 
 unsigned char	DamageReqCode;
 int		DamageEventBase;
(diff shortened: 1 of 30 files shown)


nx-X11/programs/Xserver/damageext/damageext.c
```diff
@@ -147,12 +148,12 @@ ProcDamageQueryVersion(ClientPtr client)
 	rep.majorVersion = stuff->majorVersion;
```
vatral, line 148: This part seems to be missing the change. rep.majorVersion is changed elsewhere to the new constant, and above is the only remaining ocurrence of DAMAGE_MAJOR

vatral:
I don't see anything else objectionable. Done with this one.

nx-X11/programs/Xserver/damageext/damageext.c
sunweaver, line 148: > This part seems to be missing the change. rep.majorVersion is changed elsewhere to the new 
> constant, and above is the only remaining ocurrence of DAMAGE_MAJOR

Good catch!!! I changed that and rebase. Merging in a minute...

PR #168, opened 2016-06-29
Bulk PR silencing various compiler warnings

sunweaver:
This PR is a clean-up PR silencing various compiler warnings. The default workflow for fixing warnings in nx-libs is:
- peak over to X.org and see how they solved issues
- if X.org has something, backport their fix
- if not, come up with our own solution


nx-X11/programs/Xserver/xkb/xkb.c
```diff
@@ -6200,7 +6200,7 @@ XkbResetProc(ExtensionEntry *extEntry)
 }
 
 void
-XkbExtensionInit(INITARGS)
```
vatral, line 6203: INITARGS remains defined, and used in xkb and other places in the code. Seems inconsistent to remove it only from one place.

nx-X11/programs/Xserver/Xext/xres.c
```diff
@@ -25,7 +25,6 @@
 static int
 ProcXResQueryVersion (ClientPtr client)
 {
-    REQUEST(xXResQueryVersionReq);
```
vatral, line 28: REQUEST is defined as:

```
#define REQUEST(type) \
|       register type *stuff = (type *)client->requestBuffer
```

`

?! 

"stuff" hardly seems like a good name for something to magically declare in a macro. "register" also seems suspicious. But I guess that is for later.

nx-X11/programs/Xserver/os/log.c
```diff
@@ -561,15 +561,6 @@ VAuditF(const char *f, va_list args)
     prefix = AuditPrefix();
     len = vsnprintf(buf, sizeof(buf), f, args);
 
-#if 1
-    /* XXX Compressing duplicated messages is temporarily disabled to
-     * work around bugzilla 964:
-     *     https://freedesktop.org/bugzilla/show_bug.cgi?id=964
```
vatral, line 567: The code is reenabled without change, which according to the bugzilla entry might cause problems leading to 100% CPU consumption. Apparently in the unmaintained xprint, but it seems worth noting. Perhaps danger of DoS?

nx-X11/programs/Xserver/os/utils.c
```diff
@@ -411,7 +411,9 @@ LockServer(void)
   if (lfd < 0)
     FatalError("Could not create lock file in %s\n", tmp);
   (void) sprintf(pid_str, "%10ld\n", (long)getpid());
-  (void) write(lfd, pid_str, 11);
```
vatral, line 414: Hardcoded magic number. Seems the right value and unlikely to change, though.

nx-X11/lib/X11/XKBMisc.c
```diff
@@ -666,7 +666,7 @@ register int	i;
 int		width,nOldGroups,oldWidth,newTypes[XkbNumKbdGroups];
 
     if ((!xkb) || (!XkbKeycodeInRange(xkb,key)) || (!xkb->map) ||
-	(!xkb->map->types)||(!newTypes)||((groups&XkbAllGroupsMask)==0)||
```
vatral, line 669: It could be nice to add (at a later time) some spaces around the &s in the code, it's hard to read and makes things look like a single identifier at a first glance.

vatral:
Everything else looks fine. Done.

nx-X11/programs/Xserver/xkb/xkb.c
sunweaver, line 6203: > INITARGS remains defined, and used in xkb and other places in the code. Seems inconsistent 
> to remove it only from one place. 

Good point. I checked what X.org did with that and found (and backported the below commit on X.org's xserver Git repo:

```
commit 5079db78aea8fa42e781dd876c1ee1c31571b3ae
Author: Tomas Carnecky <tom@dbservice.com>
Date:   Tue Jul 10 02:02:53 2012 +0100

    Replace INITARGS with void

    INITARGS was a hardcoded define to void.  Since knowing the function
    signature for your extensions is kinda useful, just replace it with a
    hardcoded void, but leave the define there for API compatibility.

    Signed-off-by: Tomas Carnecky <tom@dbservice.com>
    Reviewed-by: Daniel Stone <daniel@fooishbar.org>
    Reviewed-by: Cyril Brulebois <kibi@debian.org>
    Reviewed-by: Jamey Sharp <jamey@minilop.net>
    Reviewed-by: Peter Hutterer <peter.hutterer@who-t.net>
    Signed-off-by: Peter Hutterer <peter.hutterer@who-t.net>
```

nx-X11/programs/Xserver/Xext/xres.c
sunweaver, line 28: > "stuff" hardly seems like a good name for something to magically declare in a macro. "register" 
> also seems suspicious. But I guess that is for later.

```
#define REQUEST(type) \
        type *stuff = (type *)client->requestBuffer
```

The "stuff" variable name gets used quite often in X.org Xserver, actually. Above you find the REQUEST definition in X.org's Xserver. And the "register" is for later, indeed. Basically, all "register" attributes can be removed IMHO.

nx-X11/programs/Xserver/os/log.c
sunweaver, line 567: > The code is reenabled without change, which according to the bugzilla entry might cause 
> problems leading to 100% CPU consumption. Apparently in the unmaintained xprint, but it seems
> worth noting. Perhaps danger of DoS?

The issue occurred in Xprt which has been fully removed from nx-libs. I have reenabled this as done in X.org Xserver after Xprt removal. Valid point, though.

nx-X11/programs/Xserver/os/utils.c
sunweaver, line 414: > Hardcoded magic number. Seems the right value and unlikely to change, though.

pid_str is defined as char[12], so using 11 here seems appropriate. Also backported from X.org Xserver.

nx-X11/lib/X11/XKBMisc.c
sunweaver, line 669: > It could be nice to add (at a later time) some spaces around the &s in the code, it's hard to read 
> and makes things look like a single identifier at a first glance. 

I am indeed thinking about running the format scripts that has been applied to X.org's Xserver way back. Only in coordination with the team, to not wrack some pending PRs / work-in-progress.

Issue #161, opened 2016-06-27
Segfault in nxagent and screen flickering when maximizing x2go session window


HolySephi:
Server O/S: Ubuntu 14.04

x2goserver version: 4.1.0.0-0~1315~ubuntu14.04.1
nxagent version: 3.5.99.0-0~git20160625.1249+t

Client O/S: Windows 10
X2Goclient Version: 4.0.5.1

maybe related to #162 

Steps to reproduce:
- Connect to the server as a normal window (800x600)
- Maximize the x2goclient window of the session

I now get a very flickering screen because the resolution of the maximized window gets set multiple times per second:

> ...
> Info: Screen [0] resized to geometry [1920x1027] fullscreen [0].
> Info: Screen [0] resized to geometry [1920x1027] fullscreen [0].
> Info: Screen [0] resized to geometry [1920x1027] fullscreen [0].
> Info: Screen [0] resized to geometry [1920x1027] fullscreen [0].
> Info: Screen [0] resized to geometry [1920x1027] fullscreen [0].
> Info: Screen [0] resized to geometry [1920x1027] fullscreen [0].
> Info: Screen [0] resized to geometry [1920x1027] fullscreen [0].
> ...

After an undefined amount of time the process segfaults:

> x2goagent[733]: segfault at 8 ip 000000000042a263 sp 00007fff95bcec60 error 4 in nxagent[400000+4f0000]

uli42:
Is this new? Does it also happen with last week's version before the Event patches?

Vom Smartphone gesendet.

----- Ursprüngliche Nachricht -----
Von: "Florian Wicke" notifications@github.com
Gesendet: ‎27.‎06.‎2016 14:27
An: "ArcticaProject/nx-libs" nx-libs@noreply.github.com
Betreff: [ArcticaProject/nx-libs] Segfault in nxagent and screen flickeringwhen maximizing x2go session window (#161)

Server O/S: Ubuntu 14.04
x2goserver version: 4.1.0.0-0~1315~ubuntu14.04.1
nxagent version: 3.5.99.0-0~git20160625.1249+t
Client O/S: Windows 10
X2Goclient Version: 4.0.5.1
Steps to reproduce:
Connect to the server as a normal window (800x600) 
Maximize the x2goclient window of the session 
I now get a very flickering screen because the resolution of the maximized window gets set multiple times per second:
...
Info: Screen [0] resized to geometry [1920x1027] fullscreen [0].
[...]

Issue #162, opened 2016-06-27
libX11 segfault when resizing x2go session window


HolySephi:
Server O/S: Ubuntu 14.04

x2goserver version: 4.1.0.0-0~1315~ubuntu14.04.1
nxagent version: 3.5.99.0-0~git20160625.1249+t

Client O/S: Windows 10
X2Goclient Version: 4.0.5.1

Maybe related to #161 

> x2goagent[27755]: segfault at 0 ip 00007f2017001dbb sp 00007fff7cba6320 error 4 in libX11.so.6.3.0[7f2016fc6000+130000]

x2goagent segfaults in libX11.so.6.3.0 when resizing the x2go session window - happens more likely when running ANY application within the desktop like Thunar, Firefox or xterm. Just resizing the desktop (XFCE) results in nxagent segfault described in #161 

Steps to reproduce:
- Connect to the x2goserver as a normal windows (800x600) with XFCE as session
- start any program like Thunar or Firefox
- resize the x2goclient session window a few times - most of the time the segfault occurs on the first resizing attempt.

I am unable to reproduce this with a plain "nxagent :1" service and starting xfce4-session manually - it occurs only with x2goclient.

HolySephi:
~~As stated in #161 i'm unable to reproduce the segfault without the xinerama code (-rrxinerama)~~

Edit: wrong - it just took longer this time - sorry!

uli42:
I can reproduce this with an lxde session, even with a standalone agent: `nxagent -ac :55`. It works when I specify `-norender` in addition. Does this help for you, too?

sunweaver:
@uli42: https://github.com/sunweaver/nx-libs/commit/fd1af6d1288562065e44646e686c2038e8b291ce

sunweaver:
Also closing, same reasoning as given in #163.

PR #169, opened 2016-06-29
backport _XGetRequest() and _XEatDataWords()

uli42:
libXrender needs them

### DIFF
--- a/nx-X11/lib/X11/XlibInt.c
+++ b/nx-X11/lib/X11/XlibInt.c
@@ -2751,6 +2751,25 @@ void _XEatData(
 #undef SCRATCHSIZE
 }
 
+/*
+ * Read and discard "n" 32-bit words of data
+ * Matches the units of the length field in X protocol replies, and provides
+ * a single implementation of overflow checking to avoid having to replicate
+ * those checks in every caller.
+ */
+#ifndef HAVE__XEATDATAWORDS
+#include <X11/Xmd.h>  /* for LONG64 on 64-bit platforms */
+#include <limits.h>
+
+static inline void _XEatDataWords(Display *dpy, unsigned long n)
+{
+# ifndef LONG64
+    if (n >= (ULONG_MAX >> 2))
+        _XIOError(dpy);
+# endif
+    _XEatData (dpy, n << 2);
+}
+#endif
 
 /*
  * _XEnq - Place event packets on the display's queue.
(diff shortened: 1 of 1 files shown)


nx-X11/lib/X11/XlibInt.c
```diff
@@ -2751,6 +2751,25 @@ void _XEatData(
 #undef SCRATCHSIZE
 }
 
+/*
+ * Read and discard "n" 32-bit words of data
+ * Matches the units of the length field in X protocol replies, and provides
+ * a single implementation of overflow checking to avoid having to replicate
+ * those checks in every caller.
+ */
+#ifndef HAVE__XEATDATAWORDS
```
sunweaver, line 2760: where is this HAVE__XEATDATAWORDS macro coming from. I can't find any reference in X.org's libX11...

Ionic, line 2760: I think he invented it deliberately, so that we can just disable the function here should that need arise.

Not quite sure why we'd want to do this, if it's really missing and messing up Xrender, though...

sunweaver:
Have you tested a build with this patch without the libNX_X11 -> libX11 renaming hack you emailed me earlier today?

Would it be an option to symlink libNX_X11.so.6 to <somepath>/X11/libX11.so.6 (plus other .so symlinks in place and then set an LD_LIBRARY_PATH in /usr/bin/nxagent (wrapper script)?

sunweaver:
Please also update your PR branch:

Updated...

```
git checkout 3.6.x
git pull <arcticaproject-nx-libs-remote> 3.6.x
git checkout pr/backport-xgetrequest-and-xeatdatawords
git rebase 3.6.x
git push <uli42-remote> --force pr/backport-xgetrequest-and-xeatdatawords:pr/backport-xgetrequest-and-xeatdatawords
```

Thanks.

sunweaver:
Maybe we should close #161 and #162 with the commit in this PR?

uli42:
I have tested a build, but I am currently doing it again to be sure. Also, I found that there's an upstream commit for XGetRequest (https://cgit.freedesktop.org/xorg/lib/libX11/commit/src/XlibInt.c?id=4a060f993bf676cf21ad9784e010f54134da7b40) so I will take that one. For XEatDataWords  there's an upstream commit but it is xcb based. So I took this but have not added a link to that in the commit message yet: https://cgit.freedesktop.org/xorg/lib/libXfixes/commit/src/Xfixesint.h?id=b031e3b60fa1af9e49449f23d4a84395868be3ab

So please wait until I have reworked this PR.

Update: had a wrong link for the XEatDataWords commit

nx-X11/lib/X11/XlibInt.c
sunweaver, line 2760: Actually, we stumbled over a libXrender originated crash in Ubuntu 14.04 which I cannot reproduce on Debian jessie (not got to testing on Debian unstable). In that crash, we end up in libX11's _Xsend() function, because libXrender calls a function not shipped in libNX_X11. The code path then switches over to libX11 at some point (I think it was the _XGetRequest function where things went astray) and never returns into libNX_X11's objects.

sunweaver:
@uli42: please be a bit more verbose about reasoning in the commit msg. Thanks!!!

uli42:
I have re-tested a build with this pr and it is _not_ sufficient to only add these symbols. Links from libNX_X11 to libX11 in exports/lib are needed to run the nxagent without crashes. They are _not_ needed for building, though.

So the fix for #161 and #162 would be to provide libX11 compatibility links somewhere (maybe on the fly at runtime).

sunweaver:
@uli42: ok. so let's get this in first and I will take a look at packaging / symlinking magic tomorrow afternoon. Please rework your PR, so I can merge tomorrow. Thanks!!!

PR #172, opened 2016-06-30
Rework nxagent and nxproxy wrapper scripts.

sunweaver:
o Install nxagent/nxproxy executables into multi-arch path.
  o Fake libNX_X11.so\* as $LIBDIR/nx-X11/libX11.so*.
  o Allow injecting different $NXAPP containing the executable name.

  o For e.g. x2goagent this requires the following changes:

``````
- Move x2goagent symlink from /usr/lib/x2go/bin/x2goagent to
  $LIBDIR/nx/bin/x2goagent
- Replace /usr/bin/x2goagent by a two-liner...

```
#!/bin/sh

export NXAPP=x2goagent
nxagent
```
``````

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ BINDIR      ?= $(PREFIX)/bin
 LIBDIR      ?= $(PREFIX)/lib
 USRLIBDIR   ?= $(LIBDIR)
 INCLUDEDIR  ?= $(PREFIX)/include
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by sunweaver** [comment]
@Ionic, @uli42: Thanks for your review on IRC. I will merge this PR directly after #152.

```
Fri Jul  1 00:38:00 CEST 2016 ->
00:38 < Ionic> okay, there are more potential errors but let's call it good for now
```

PR #173, opened 2016-06-30
Pr/use system libXext for shape

uli42:
please review/test


uli42:
I have made a complete NX_Xext drop, so this PR is obsolete now

PR #183, opened 2016-07-09
Feature/reconnect checks

Ionic:
Adds the new `ReconnectTolerance` `nxagentOption` and `tolerance` `nx/nx` parameter.

The known values are `Strict` (= 0), `Safe` (= 1), `Risky` (= 2) and `None` (= 3). Any higher value is mapped to `None`.

The code within checking functions used at reconnect time has been adapted to utilize the `ReconnectTolerance` `nxagentOption` value and modify its behavior based on it. Users can now select for these checks to be less strict.

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -1405,6 +1405,82 @@ static void nxagentParseOptions(char *name, char *value)
 
     return;
   }
+  else if (!strcmp(name, "tolerancechecks"))
+  {
+    if (strcmp(value, "strict") == 0)
+    {
+      nxagentChangeOption(ReconnectTolerance, ToleranceChecksStrict);
+    }
+    else if (strcmp(value, "safe") == 0)
+    {
+      nxagentChangeOption(ReconnectTolerance, ToleranceChecksSafe);
+    }
+    else if (strcmp(value, "risky") == 0)
+    {
+      nxagentChangeOption(ReconnectTolerance, ToleranceChecksRisky);
+    }
+    else if (strcmp(value, "none") == 0)
+    {
+      nxagentChangeOption(ReconnectTolerance, ToleranceChecksBypass);
+    }
+    else
+    {
+      /*
+       * Check for a matching integer. Or any integer, really.
+       */
+      long tolerance_parse = 0;
+
+      errno = 0;
+      tolerance_parse = strtol(value, NULL, 10);
+
+      if ((errno) && (0 == tolerance_parse))
+      {
+        fprintf(stderr, "nxagentParseOptions: Unable to convert value [%s] of option [%s]. "
+                        "Ignoring option.\n",
+                        validateString(value), validateString(name));
+
+        return;
+      }
+
+      if ((long) UINT_MAX < tolerance_parse)
+      {
+        tolerance_parse = UINT_MAX;
+
+        fprintf(stderr, "nxagentParseOptions: Warning: value [%s] of option [%s] "
+                        "out of range, clamped to [%u].\n",
+                        validateString(value), validateString(name), tolerance_parse);
+      }
+
+      if (0 > tolerance_parse)
+      {
+        tolerance_parse = 0;
+
+        fprintf(stderr, "nxagentParseOptions: Warning: value [%s] of option [%s] "
+                        "out of range, clamped to [%u].\n",
+                        validateString(value), validateString(name), tolerance_parse);
+      }
+
+      #ifdef TEST
+      switch (tolerance_parse) {
+        case ToleranceChecksStrict:
+        case ToleranceChecksSafe:
+        case ToleranceChecksRisky:
+        case ToleranceChecksBypass:
+                               break;
+        default:
+                               fprintf(stderr, "nxagentParseOptions: Warning: value [%s] of "
+                                               "option [%s] unknown, will be mapped to "
+                                               "\"Bypass\" [%u] value internally.\n",
+                                       validateString(value), validateString(name),
+                                       (unsigned int)ToleranceChecksBypass);
+      }
+      #endif
+
+      nxagentChangeOption(ReconnectTolerance, tolerance_parse);
+    }
+
+    return;
+  }
   else
   {
     #ifdef DEBUG
--- a/nx-X11/programs/Xserver/hw/nxagent/Display.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Display.c
@@ -2401,60 +2401,139 @@ static int nxagentCheckForDepthsCompatibility()
 
 static int nxagentCheckForPixmapFormatsCompatibility()
 {
-  int i, j;
-  int matched;
-  int compatible;
+  /*
+   * Depending on the (reconnect) tolerance checks value, this
+   * function checks stricter or looser:
+   *   - Strict means that the number of internal and external
+   *     pixmap formats must match exactly and every
+   *     internal pixmap format must be available in the
+   *     external pixmap format array.
+   *   - Safe means that the number of pixmap formats might
+   *     diverge, but all internal pixmap formats must
+   *     also be included in the external pixmap formats
+   *     array. This is recommended, because it allows
+   *     clients with more pixmap formats to still connect,
+   *     but not lose functionality.
+   *   - Risky means that the internal pixmap formats array is
+   *     allowed to be smaller than the external pixmap
+   *     formats array, but at least one pixmap format must
+   *     be included in both. This is potentially unsafe.
+   *   - Bypass or higher means that all of these checks are
+   *     essentially deactivated. This is a very bad idea.
+   */
 
-  compatible = 1;
+  const unsigned int tolerance = nxagentOption(ReconnectTolerance);
 
-  if (nxagentNumPixmapFormats != nxagentRemoteNumPixmapFormats)
+  if (ToleranceChecksBypass <= tolerance)
   {
-    #ifdef DEBUG
-    fprintf(stderr, "nxagentCheckForPixmapFormatsCompatibility: WARNING! Number of internal pixmap formats [%d] "
-                "doesn't match with remote formats [%d].\n", nxagentNumPixmapFormats,
-                    nxagentRemoteNumPixmapFormats);
+    #ifdef WARNING
+    fprintf(stderr, "nxagentCheckForPixmapFormatsCompatibility: WARNING! Not proceeding with any checks, "
+                    "because tolerance [%u] higher than or equal [%u]. Number of internally available "
+                    "pixmap formats is [%d], number of externally available pixmap formats is [%d].\n",
+            tolerance, ToleranceChecksBypass, nxagentNumPixmapFormats, nxagentRemoteNumPixmapFormats);
     #endif
+
+    return 1;
   }
 
-  for (i = 0; i < nxagentNumPixmapFormats; i++)
+  if ((ToleranceChecksStrict == tolerance) && (nxagentNumPixmapFormats != nxagentRemoteNumPixmapFormats))
   {
-    matched = 0;
+    #ifdef DEBUG
+    fprintf(stderr, "nxagentCheckForPixmapFormatsCompatibility: WARNING! No tolerance allowed and number "
+                    "of internal pixmap formats [%d] doesn't match with number of remote formats [%d].\n",
+            nxagentNumPixmapFormats, nxagentRemoteNumPixmapFormats);
+    #endif
 
-    for (j = 0; j < nxagentRemoteNumPixmapFormats; j++)
+    return 0;
+  }
+
+  if ((ToleranceChecksSafe == tolerance) && (nxagentNumPixmapFormats > nxagentRemoteNumPixmapFormats))
+  {
+    #ifdef DEBUG
+    fprintf(stderr, "nxagentCheckForPixmapFormatsCompatibility: WARNING! Tolerance [%u] too low "
+                    "and number of internal pixmap formats [%d] higher than number of external formats [%d].\n",
+            tolerance, nxagentNumPixmapFormats, nxagentRemoteNumPixmapFormats);
+    #endif
+
+    return 0;
+  }
+
+  /*
+   * By now the tolerance is either:
+   *   - Strict
+   *   - Safe and:
+   *     o the number of internal and external pixmap formats
+   *       matches exactly, or
+   *     o the number of external pixmap formats is higher than
+   *       the number of internal pixmap formats,
+   *   - Risky
+   */
+
+  bool compatible = true;
+  bool one_match = false;
+  bool matched = false;
+  int total_matches = 0;
+
+  for (int i = 0; i < nxagentNumPixmapFormats; ++i)
+  {
+    matched = false;
+
+    for (int j = 0; j < nxagentRemoteNumPixmapFormats; ++j)
     {
       if (nxagentPixmapFormats[i].depth == nxagentRemotePixmapFormats[j].depth &&
           nxagentPixmapFormats[i].bits_per_pixel == nxagentRemotePixmapFormats[j].bits_per_pixel &&
           nxagentPixmapFormats[i].scanline_pad == nxagentRemotePixmapFormats[j].scanline_pad)
       {
-        matched = 1;
+        matched = true;
+        one_match = true;
+        ++total_matches;
 
         break;
       }
     }
 
-    if (matched == 0)
+    if ((ToleranceChecksRisky > tolerance) && (!matched))
     {
       #ifdef WARNING
-      fprintf(stderr, "nxagentCheckForPixmapFormatsCompatibility: WARNING! Failed to match internal "
-                  "pixmap format depth [%d] bpp [%d] pad [%d].\n", nxagentPixmapFormats[i].depth,
-                      nxagentPixmapFormats[i].bits_per_pixel, nxagentPixmapFormats[i].scanline_pad);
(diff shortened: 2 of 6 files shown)


Ionic:
Referencing Mario's bug report #132.

Ionic:
Also, while working on this, I found two other functions which could benefit from being ported, namely `nxagentInitAndCheckVisuals()` and `nxagentCheckForColormapsCompatibility()`. These weren't listed by Mario, but maybe we want more loose checks on these as well? I haven't taken a good look at these yet, but they seem to be special (for instance, Visuals are handled somewhat differently based on their type and the colormaps are not explicitly evaluated, but merely the number of colormaps compared.)

What do you think?

uli42:
"The known values are Strict (= 0), Safe (= 1), Risky (= 2) and None (= 3). "

So we have a "strict" tolerance for value 0, which sounds a bit clumsy. And for 3 we have "no" tolerance. Which is simply wrong.

So I we need new names, e.g.:
0 none
1 low
2 middle
3 high

But as we lose the ability to insert further levels later with that (what to call a level between low and middle) I don't like that, either. We should drop the keywords. Proposal:
tolerance is a value between 0 and 255. 0-63 is what is "strict" now, 64-127 is safe, 128-191 is risky and 192-255 is none. I there's a new level to come we can insert it at an arbitrary position.

nxagentInitAndCheckVisuals() and nxagentCheckForColormapsCompatibility() are definitely called at reconnect, so they should be included in this patch

Ionic:
But leaving out the symbolic names means working with "magic numbers". That sounds bad, too (for maintainability.)

I could introduce symbols like `none_start = 0`, `none_end = 63`, ... at the same time, but if a finer-grained level is to be inserted, should it start from the middle to one end (e.g., a new level `tiny` from 32 to 63) or fixed to a specific value?

We have to be aware that whatever we do, a change in levels will mean breakage. In the above example, a level of 32 would be plainly mapped to the `none` behavior in older nx-libs versions, while in newer the behavior will be different. There's probably no way to avoid that, though?

Regarding `nxagentInitAndCheckVisuals()` and `nxagentCheckForColormapsCompatibility()`: in theory, yes. However, since Mario probably didn't experience any problems in version-heterogenous environments(?) with these checks and they weren't marked (indicating a lower priority), I'm asking for @sunweaver's opinion in this case.

uli42:
What about introducing a percentage here. It defines the number if checks that have to pass. 100% mean "all checks must pass". Of course a check is added this will also be affected.

Or another approach: A bitfield or keyword list indicating what checks need to pass. E.g. reconnect-checks=visuals,depth (can we categorize the checks like this?)

Vom Smartphone gesendet.

Von: "Mihai Moldovan" notifications@github.com
Gesendet: ‎11.‎07.‎2016 07:55
Cc: "Ulrich Sibiller" uli42@gmx.de; "Assign" assign@noreply.github.com
Betreff: Re: [ArcticaProject/nx-libs] Feature/reconnect checks (#183)

But leaving out the symbolic names means working with "magic numbers". That sounds bad, too (for maintainability.)
I could introduce symbols like none_start = 0, none_end = 63, ... at the same time, but if a finer-grained level is to be inserted, should it start from the middle to one end (e.g., a new level tiny from 32 to 63) or fixed to a specific value?
We have to be aware that whatever we do, a change in levels will mean breakage. In the above example, a level of 32 would be plainly mapped to the none behavior in older nx-libs versions, while in newer the behavior will be different. There's probably no way to avoid that, though?
Regarding nxagentInitAndCheckVisuals() and nxagentCheckForColormapsCompatibility(): in theory, yes. However, since Mario probably didn't experience any problems in version-heterogenous environments(?) with these checks and they weren't marked (indicating a lower priority), I'm asking for @sunweaver's opinion in this case.

Ionic:
Accepting failing checks (or rather saying which ones to skip) might be worth doing in a separate iteration, but not here.

The general problem is that some of these checks are unnecessarily strict, so we want to add some flexibility in what is allowed. For instance, if the new depth value is bigger than the old one, that probably won't create problems and we can allow reconnecting. Currently nxagent will refuse to reconnect.

An even worse example of this is the depth-list check. If there are MORE depths available than were previously used, nxagent currently refuses to reconnect, which is probably just stupid.

So for now, I want fine-grained levels of flexibility within the checks themselves.

If we stick to symbolic names only, inserting new levels probably won't cause a lot of breakage (minus passing an invalid "new" option to an older nxagent binary that won't understand it.)

As for the symbolic names, I'd like to go with `None`, `Strict`, `Risky` and `Bypass`. I realize that "bypass tolerance" sounds clumsy, but it's a fitting description of what that level does. :)

So maybe accepting an integer value isn't a good idea to begin with? We could at least work around potential problems when adding new levels if I disallow numeric levels.

sunweaver:
Hi Mihai, hi Uli,

thanks for your effort on this!!!

On  Do 14 Jul 2016 08:34:13 CEST, Mihai Moldovan wrote:

> Accepting failing checks (or rather saying which ones to skip) might  
> be worth doing in a separate iteration, but not here.
> 
> The general problem is that some of these checks are unnecessarily  
> strict, so we want to add some flexibility in what is allowed. For  
> instance, if the new depth value is bigger than the old one, that  
> probably won't create problems and we can allow reconnecting.  
> Currently nxagent will refuse to reconnect.
> 
> An even worse example of this is the depth-list check. If there are  
> MORE depths available than were previously used, nxagent currently  
> refuses to reconnect, which is probably just stupid.
> 
> So for now, I want fine-grained levels of flexibility within the  
> checks themselves.
> 
> If we stick to symbolic names only, inserting new levels probably  
> won't cause a lot of breakage (minus passing an invalid "new" option  
> to an older nxagent binary that won't understand it.)
> 
> As for the symbolic names, I'd like to go with `None`, `Strict`,  
> `Risky` and `Bypass`. I realize that "bypass tolerance" sounds  
> clumsy, but it's a fitting description of what that level does. :)
> 
> So maybe accepting an integer value isn't a good idea to begin with?  
> We could at least work around potential problems when adding new  
> levels if I disallow numeric levels.

Wouldn't a bitwise approach (one of Ulrich's suggestions) solve all the above?

We have a number of different "tolerance" checks and you can  
enable/disable them bit-by-bit. If there are checks that don't make  
sense with another check, then some bits might need auto-enabling.

???

Mike

## 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

Ionic:
That would work, if we were to create a new level for every function. `#functions * levels` could be very very costly costly though, do we really want to do that? (+ sanity functions checks, i.e., enabling both `none` tolerance (level 0) + `safe` (level 1) wouldn't make a lot of sense, so nxagent should either error out or only enable the most restrictive check, which would be `none` (level 0) in this case. That would need to be carried out for _every_ single check.)

Ionic:
Additionally, passing bitfields via `nx/nx` options for each check individually will blow up the command line/options file and users have to be very carefully to now forget a check... Not sure what to do. Comments?

So far I'd rename the levels and be done, but you don't seem to like the current solution too much?

uli42:
I am lacking time currently. Maybe a simple switch on/off is sufficient,   too

Vom Smartphone gesendet.

Von: "Mihai Moldovan" notifications@github.com
Gesendet: ‎23.‎07.‎2016 05:28
Cc: "Ulrich Sibiller" uli42@gmx.de; "Assign" assign@noreply.github.com
Betreff: Re: [ArcticaProject/nx-libs] Feature/reconnect checks (#183)

Additionally, passing bitfields via nx/nx options for each check individually will blow up the command line/options file and users have to be very carefully to now forget a check... Not sure what to do. Comments?
So far I'd rename the levels and be done, but you don't seem to like the current solution too much?

uli42:
On Sat, Jul 23, 2016 at 5:28 AM, Mihai Moldovan

> That would work, if we were to create a new level for every function. #functions \* levels
> could be very very costly costly though, do we really want to do that? (+ sanity functions
> checks, i.e., enabling both none tolerance (level 0) + safe (level 1) wouldn't make a lot of
> sense, so nxagent should either error out or only enable the most restrictive check,
> which would be none (level 0) in this case. That would need to be carried out for every single check.)

First, adding bitfields would mean dropping the tolerance levels. So
no #function \* levels. If we'd keep the levels the bitfields coudd be
used inside the nxagent but not be available outside.

> Additionally, passing bitfields via nx/nx options for each check
> individually will blow up the command line/options file and users have to be
> very carefully to now forget a check... Not sure what to do. Comments?

I do not see why anyone would need the tolerance level in the display
string or the options file. This is something that can be handled
differently at any reconnect. So it should be a client (proxy) option.
The agent must of course be informed about the desired checks, but
there's no need to store them anywhere. And even if there was a need
nxagent could handle that internally using a bitfield variable.

Let's say we have 6 checks and want to be able to deactivate them
individually. So we'd need 6 bits (if a set bit would indicate
deactivation or activation could be discussed). The party issuing the
reconnect could set any of these bits and get _one_ resulting number,
which has to be passed to nxagent. How the user may choose among the
possible checks is a different topic not of interest from the nxagent
point of view.

> So far I'd rename the levels and be done, but you don't seem to like the
> current solution too much?

Well, I'd like some fine tuning on the wording. But my main concern is
that introducing a solution that is clumsy from the beginning might
lead to having to support that for some time to not break
compatibility. So it should at least be well thought.

Finally: I am not sure about the effects of ignoring all checks. Will
this break a session beyond repair? Or will it just trigger some weird
problems for the current reconnect and work fine (again) for the next
reconnect from a client supporting more depths/visuals/whatever? If
so, a simple switch disabling all checks might be the solution.

Uli

uli42:
Looking at Reconnect.c I just stumbled over this:

```
void nxagentInitializeRecLossyLevel()
{
  *(int *)reconnectLossyLevel[DISPLAY_STEP]    = 0;
  *(int *)reconnectLossyLevel[SCREEN_STEP]     = 0;
  *(int *)reconnectLossyLevel[FONT_STEP]       = 0;
  *(int *)reconnectLossyLevel[PIXMAP_STEP]     = 0;
  *(int *)reconnectLossyLevel[GC_STEP]         = 0;
  *(int *)reconnectLossyLevel[CURSOR_STEP]     = 0;
  *(int *)reconnectLossyLevel[COLORMAP_STEP]   = 0;
  *(int *)reconnectLossyLevel[WINDOW_STEP]     = 0;
  *(int *)reconnectLossyLevel[GLYPHSET_STEP]   = 0;
  *(int *)reconnectLossyLevel[PICTFORMAT_STEP] = 0;
  *(int *)reconnectLossyLevel[PICTURE_STEP]    = 0;
}
```

These are passed to all functions called at reconnect but most of time they are not used at all (only exception is the font handling; and there it is called reconnectFlexibility ;-) ). Maybe we should try to extend/reuse them.

Ionic:
> First, adding bitfields would mean dropping the tolerance levels. So no #function \* levels. If we'd keep the levels the bitfields coudd be used inside the nxagent but not be available outside.

I wouldn't drop the levels... only disabling or enabling checks doesn't seem right. Also...

> Finally: I am not sure about the effects of ignoring all checks. Will this break a session beyond repair? Or will it just trigger some weird problems for the current reconnect and work fine (again) for the next reconnect from a client supporting more depths/visuals/whatever?

To be perfectly fair, I have no idea what happens if checks are completely disabled (which is also why I _want_ different levels of flexibility/tolerances for each specific check and not just blindly disabling everything.) It sounds risky to me and I hardly doubt this should be a supported use case, but if users wish to do so, they should be able to.

> I do not see why anyone would need the tolerance level in the display string or the options file. This is something that can be handled differently at any reconnect. So it should be a client (proxy) option. The agent must of course be informed about the desired checks, but there's no need to store them anywhere. And even if there was a need nxagent could handle that internally using a bitfield variable.

Uhm, yeah, that won't work too well. Generally, the proxy is not exchanging options with the agent. Instead, agent options are handled via the `DISPLAY` variable or `options` file in `nxagent`. My initial idea for the sleep option (previous PR) also was to be processed by the proxy and passed to the agent, but this turned out to be very, very difficult. A few options are passed to the agent, but these only affect cache sizes etc. This passing is done via a unique X request and then handled by the agent. Because X requests are limited in size to 32 bytes and this request already is 32 bytes in size, adding more data would be difficult. Additional data can be provided, but only via "additional data hacks". Anywho, back then when I complained about this situation to @sunweaver, he advised to scrap all that and implement it as an `nxagent` `nx/nx` option. Eventually, it really is an agent option only semantically, just like the checks here are only agent-dependent and any options should be agent-only options.
This said, my concern with the bitfield still stands. Even if represented via a bitfield internally, we'd need to accept different options for every single check. You already found `reconnectLossyLevel`, so... that would mean a unique option for every check/"step" (as they call it) in there. Of course we could add another pseudo-check/step called "ALL" or "OTHER" or "DEFAULT" which is used if no explicit check/step value has been provided (i.e., the check/step value was not explicitly overridden), but that also sounds weird.

> Well, I'd like some fine tuning on the wording.

What exactly, other than renaming the levels?

> Looking at Reconnect.c I just stumbled over this:

Yeah, that's what I've been using in my first approach (remember the `flexibility` parameter some - but not all - checking functions had?)
As some checks didn't use it at all and always passing a parameter feels somewhat clumsy, @sunweaver suggested reimplementing this as an `nxagentOption`, which could be queried from any function directly instead of needing to be passed as a parameter. I liked this and rebased my code like that.
I probably could have dropped the `reconnectLossyLevel` bits as well, but I'd like to split up the cleanup into another PR or do it afterwards.
`reconnectLossyLevel` seems to be just an artifact anyway. None of the functions that are of interest to us ever ued it and it was never properly implemented - showcased by the fact that the value is hardcoded. Changing the level would mean to change and recompile the source code which is a no-go.

uli42:
On Mon, Jul 25, 2016 at 7:23 AM, Mihai Moldovan

> First, adding bitfields would mean dropping the tolerance levels. So no #function \* levels. If we'd keep the levels the bitfields coudd be used inside the nxagent but not be available outside.
> 
> I wouldn't drop the levels... only disabling or enabling checks doesn't seem right. Also...
> 
> Finally: I am not sure about the effects of ignoring all checks. Will this break a session beyond repair? Or will it just trigger some weird problems for the current reconnect and work fine (again) for the next reconnect from a client supporting more depths/visuals/whatever?
> 
> To be perfectly fair, I have no idea what happens if checks are completely disabled (which is also why I want different levels of flexibility/tolerances for each specific check and not just blindly disabling everything.) It sounds risky to me and I hardly doubt this should be a supported use case, but if users wish to do so, they should be able to.

Ok, so honestly these stages are a workaround for us not fully
understanding the effects of the software/X11 ;-). If we'd fully
understood all that we could say what could happen. So these levels
make sense.

> I do not see why anyone would need the tolerance level in the display string or the options file. This is something that can be handled differently at any reconnect. So it should be a client (proxy) option. The agent must of course be informed about the desired checks, but there's no need to store them anywhere. And even if there was a need nxagent could handle that internally using a bitfield variable.
> 
> Uhm, yeah, that won't work too well. Generally, the proxy is not exchanging options with the agent. Instead, agent options are handled via the DISPLAY variable or options file in nxagent. My initial idea for the sleep option (previous PR) also was to be processed by the proxy and passed to the agent, but this turned out to be very, very difficult. A few options are passed to the agent, but these only affect cache sizes etc. This passing is done via an unique X request and then handled by the agent. Because X requests are limited in size to 32 bytes and this request already is 32 bytes in size, adding more data would be difficult. Additional data can be provided, but only via "additional data hacks". Anywho, back then when I complained about this situation to @sunweaver, he advised to scrap all that and implement it as an nxagent nx/nx option. Eventually, it really is an agent option only semantically, just like the checks here are only agent-dependent and any options should be agent-only options.

Ok, understood, I have not been aware of that.

> Well, I'd like some fine tuning on the wording.
> 
> What exactly, other than renaming the levels?

What I already mentioned: tolerance=bypass or tolerance=none are
pointing in the opposite direction. Maybe
tolerancechecks=strict/safe/risky/bypass (or none) would be better.

> As some checks didn't use it at all and always passing a parameter feels somewhat clumsy, @sunweaver suggested reimplementing this as an nxagentOption, which could be queried from any function directly instead of needing to be passed as a parameter. I liked this and rebased my code like that.
> I probably could have dropped the reconnectLossyLevel bits as well, but I'd like to split up the cleanup into another PR or do it afterwards.
> reconnectLossyLevel seems to be just an artifact anyway. None of the functions that are of interest to us ever ued it and it was never properly implemented - showcased by the fact that the value is hardcoded. Changing the level would mean to change and recompile the source code which is a no-go.

Ah, ok. Sure, that code should vanish anyway.

Uli

Ionic:
Adaptions:
- nx/nx parameter now called `reconnectchecks`
- within code, still use `ReconnectTolerance`, but rename the modes enum to `ToleranceChecksMode` and the values to `ToleranceChecks{Strict,...}`
- rename `None` to `Bypass` everywhere.

Ionic:
And rebased against current `3.6.x`, too.

uli42:
I am happy ;-)

Vom Smartphone gesendet.

Von: "Mihai Moldovan" notifications@github.com
Gesendet: ‎28.‎07.‎2016 05:20
Cc: "Ulrich Sibiller" uli42@gmx.de; "Assign" assign@noreply.github.com
Betreff: Re: [ArcticaProject/nx-libs] Feature/reconnect checks (#183)

Adaptions:
nx/nx parameter now called reconnectchecks 
within code, still use ReconnectTolerance, but rename the modes enum to ToleranceChecksMode and the values to ToleranceChecks{Strict,...} 
rename None to Bypass everywhere.

sunweaver:
I am just taking a look over it and rebase in some minutes if no flaw pops up...

sunweaver:
s/rebase/merge/

PR #194, opened 2016-08-21
Fix failure on first paste from CLIPBOARD

uli42:
Pasting the clipboard did not work directly after session start.

I am a bit what other effects might be introduced by leaving out this code. Maybe we should find out why it was introduced in the first place...

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
@@ -1583,6 +1583,14 @@ int nxagentInitClipboard(WindowPtr pWin)
     nxagentXFixesInfo.Initialized = 1;
   }
 
+  /*
+     The first paste from CLIPBOARD did not work directly after
+     session start. Removing this code makes it work. It is unsure why
+     it was introduced in the first place so it is possible that we
+     see other effects by leaving out this code.
+
+     Fixes X2Go bug #952, see https://bugs.x2go.org/952 for details .
+
   if (nxagentSessionId[0])
   {
     #ifdef TEST
@@ -1595,6 +1603,7 @@ int nxagentInitClipboard(WindowPtr pWin)
     pWin -> eventMask |= PropertyChangeMask;
     nxagentChangeWindowAttributes(pWin, CWEventMask);
   }
+  */
 
   if (nxagentReconnectTrap)
   {


nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
```diff
@@ -1583,6 +1583,9 @@ int nxagentInitClipboard(WindowPtr pWin)
     nxagentXFixesInfo.Initialized = 1;
   }
 
+  /*
+     Removing this seems to solve xg2o issue #952 (no paste directly after session start)
+
```
sunweaver, line 1593: Thanks for providing this patch. Please be less X2Go BTS centric in the complete PR. Describe what your patch fixes:
- in the PR name (please rename)
- in the comment above
- in the commit msg

sunweaver:
Please mention your concerns as given in the PR text in the comment above the commented out code passage. So we know, that we might expect other weird behaviour from commenting out the discovered code passage. Please fix typos/grammar in that text, too.

In the merge, I will reference this PR number and the X2Go BTS number (in a machine readable way). No need to mention the X2Go BTS all over the place. Please mention the X2Go BTS number in the comment-in-the-code, but only at the end of your textual explanation.

E.g.: This passage breaks first paste on Windows, whereas being unsure what other effects might be introduced by leaving out this code. (See https://bugs.x2go.org/952 for details).

Thanks!
Mike

uli42:
OK, I have updated the commit as requested.

I also did some research and found that this code had emerged between nxagent 1.4.0-33 and 1.4.0-47 (I do not have any intermediate releases). However, the changelog only contains two commits regarding clipboard:

```
nxagent-1.4.0-35

- Implemented a new method to comunicate to nxclient,
  the raise of the agent root window, taking the ownership
  of the selection "A-{MD5 of session}".
  Used the same selection to let nxclient comunicate to agent
  by changing the property on the same string, when the user
  choose by the administrator to terminate or suspend the
  session.
```

and

```
nxagent-1.4.0-44

- Fixed a problem in the clipboard procedure.
  Now when we get a request of the selection
  from an internal client we have to, if the 
  owner is on the X server, forward the request 
  to X, otherwise we have to pass the request 
  to our internal client.
  But for a problem in this procedure we passed, 
  in some situation, the request to the internal 
  client even if the owner was on the other side.
```

None of them seems to address the lines I have commented in this commit...

PR #196, opened 2016-08-24
Update X11's XKeysymDB

sunweaver:
- lib/X11/XKeysymDB: Update to last version found in X.org (between 1.3.5 and 1.3.6).
- fix-up commit from the libX11 pkg in Fedora 14.

### DIFF
--- a/nx-X11/lib/X11/XKeysymDB
+++ b/nx-X11/lib/X11/XKeysymDB
@@ -1,4 +1,3 @@
-! $Xorg: XKeysymDB,v 1.3 2000/08/17 19:45:04 cpqbld Exp $
 ! Copyright 1993 Massachusetts Institute of Technology
 !
 ! Permission to use, copy, modify, distribute, and sell this software and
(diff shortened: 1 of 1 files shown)


sunweaver:
Interestingly, we should also consider this commit in libX11 (X.org):

```
commit 50b4b07073bd46cbef3627c160a240bd2a2b99d5
Author: Daniel Stone <daniel@fooishbar.org>
Date:   Tue Jun 15 18:49:43 2010 +0100

    Delete now-redundant XKeysymDB

    Since XStringToKeysym now supports all the vendor keysyms, just delete
    our XKeysymDB, which was incomplete at best, misleading at worst, and
    always an annoyance.

    Signed-off-by: Daniel Stone <daniel@fooishbar.org>
    Reviewed-by: Alan Coopersmith <alan.coopersmith@oracle.com>

```

sunweaver:
@uli42: Even if 1.3.5 introduced forced usage of XCB, there may be many more parts of libX11 that can be upgraded even further than 1.3.4. (However, 1.3.4 is a good starting point for this!!!)

uli42:
I agree.

PR #197, opened 2016-08-27
Upgrade lib x11 and other stuff

uli42:
This is not a real pull request. I just made it to inform you about my current state before leaving to vacation. This is a bit rough and not well tested..

It is based on the other upgrade_libX11 pull request. I have added lots of files I had simply forgotten previously (it is never good when you interrupt such work for some days). This lead to xserver no longer compiling because it uses softlinks to some files from libX11. I have not bothered to debug that. Instead I solved it like upstream does: use own copies. Doing that I discovered that xkbfile is also integrated using softlinks, so I worked on and could drop xkbfile completely by adding some of its files to the server tree (xkb dir) and some to the include dir, just as upstream does, too.
And while working on that I made xserver more independent of the rest by replacing external includes with own copies, just as upstream does.

Intermediate problems arose with makekeys which had needed to be called differently in the Makefile now.

Some fixes from Debian libX11 1.3.3 are also included, but should be added in a separate PR and removed from here. There are quite a couple of patches in debian's libX11 1.3.3 which I have identified but not included yet...

The parts where libX11 and xserver are decoupled should also be made separate PRs.

libX11 1.3.4 has a nls tree that is completely missing in NX. I am not sure if we should/need to port this. I have done nothing here yet.

I plan to clean this mess and make clean PR(s) when I am back but this will need some time...

So feel free to use this, clean it up, extract parts of it, whatever.


uli42:
superseeded by #214 and #215 as well as some other commits.

PR #198, opened 2016-09-21
move xkblib to xserver

uli42:
Upstream does not use xkbfile but has own copies of the required
files. It was not used elsewhere.


uli42:
is part of PR #199

Issue #200, opened 2016-09-22
Server build still referring to X11 includes


uli42:
Imake places dependencies to include files in the resulting Makefile. There you can see that the compilation refers to files in /usr/include/X11 even if we have our own copies of most of them:

```
for i in `find . -name Makefile`; do grep /usr/include/X11 $i; done | cut -d: -f2- | xargs -n1 | sort | uniq -c | grep /usr/include/X11 | sort -rn
    131 /usr/include/X11/Xmd.h
    129 /usr/include/X11/fonts/font.h
    102 /usr/include/X11/fonts/fsmasks.h
     76 /usr/include/X11/Xfuncproto.h
     61 /usr/include/X11/extensions/shapeconst.h
     46 /usr/include/X11/Xproto.h
     45 /usr/include/X11/extensions/shape.h
     40 /usr/include/X11/Xutil.h
     27 /usr/include/X11/Xdefs.h
     27 /usr/include/X11/fonts/fontstruct.h
     27 /usr/include/X11/fonts/fontproto.h
     26 /usr/include/X11/X.h
     22 /usr/include/X11/Xdmcp.h
     11 /usr/include/X11/extensions/shapeproto.h
      7 /usr/include/X11/extensions/xfixeswire.h
      6 /usr/include/X11/extensions/xfixesproto.h
      6 /usr/include/X11/extensions/damagewire.h
      4 /usr/include/X11/Xprotostr.h
      3 /usr/include/X11/extensions/shmproto.h
      3 /usr/include/X11/extensions/shm.h
      1 /usr/include/X11/extensions/Xext.h
      1 /usr/include/X11/extensions/shmstr.h
      1 /usr/include/X11/extensions/extutil.h
      1 /usr/include/X11/extensions/composite.h
```

We need to clarify why that is and what we can do against that as it might cause hard to track problems...

In Agent.h I have included a workaround but I do not think we can add something like that everywhere:

```
/*
  <X11/extensions/shape.h> includes <X11/Xutil.h> but we need
  <nx-X11/Xutil.h>. As both use the same header guard we can first
  include <nx-X11/Xutil.h> and be ok.
*/
#include <nx-X11/Xutil.h>

#include <X11/extensions/shape.h>
```

PR #201, opened 2016-09-24
ensure that <X11/..> headers references use nx-X11 files if existing

uli42:
System headers including other headers via <X11/..> now use the
nx-X11 variant (if existing).

This fixes Arctica issue #200. But we should think about dropping the
<nx-X11/..> namespace now for includes.

### DIFF
--- a/nx-X11/include/Imakefile
+++ b/nx-X11/include/Imakefile
@@ -56,6 +56,12 @@ all::
 
 BuildIncludes($(HEADERS),nx-X11,..)
 
+includes::
+	-$(RM) $(BUILDINCDIR)/X11
+	-$(LN) nx-X11 $(BUILDINCDIR)/X11
+
+
+
 #if BuildLibraries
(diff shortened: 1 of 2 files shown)


sunweaver (changes requested):
Thanks, Uli, for this PR.

Please do two formal things.
1. Provide the proposed changes in two commits (one per file, with two different messages)
2. The revert on Agent.h is a revert, so please use git revert for undoing the changes previously committed via 9a4bb50af54ced8d53112b8fa3b5b9dc3b481224. This provides more meta info.
3. The change to the Imakefile, please ship with a proper "fixes" statement (that works with the Github issue tracker).  I normally put an extra line below the commit message:

```
Fixes ArcticaProject/nx-libs#200.
```

Thanks!

sunweaver:
... actually _three_ formal things...

uli42:
Changes applied. Ok?

sunweaver (approved):

PR #208, opened 2016-09-29
Pr/decouple xserver

uli42:
- Use own copies of xkb headers just as upstream does.
- refer to Xtrans via header file xstrans.c instead of softlinking in the makefile
- backport fixes for some compiler warnings that became obvious after the xstrans switch. These happened to be CVEs...

### DIFF
--- a/nx-X11/programs/Xserver/dix/devices.c
+++ b/nx-X11/programs/Xserver/dix/devices.c
@@ -61,7 +61,6 @@ SOFTWARE.
 #include "cursorstr.h"
 #include "dixstruct.h"
 #include "site.h"
-#define	XKB_IN_SERVER
 #ifdef XKB
 #include <xkbsrv.h>
 #endif
--- a/nx-X11/programs/Xserver/dix/devices.c
+++ b/nx-X11/programs/Xserver/dix/devices.c
@@ -62,7 +62,7 @@ SOFTWARE.
 #include "dixstruct.h"
 #include "site.h"
 #ifdef XKB
-#include <xkbsrv.h>
+#include "xkbsrv.h"
 #endif
 #ifdef XCSECURITY
 #define _SECURITY_SERVER
--- a/nx-X11/programs/Xserver/dix/devices.c
+++ b/nx-X11/programs/Xserver/dix/devices.c
@@ -63,7 +63,7 @@ SOFTWARE.
 #include "site.h"
 #define	XKB_IN_SERVER
 #ifdef XKB
-#include <nx-X11/extensions/XKBsrv.h>
+#include <xkbsrv.h>
 #endif
 #ifdef XCSECURITY
 #define _SECURITY_SERVER
--- a/nx-X11/programs/Xserver/dix/dispatch.c
+++ b/nx-X11/programs/Xserver/dix/dispatch.c
@@ -110,7 +110,7 @@ int ProcInitialConnection();
 #endif
 #ifdef XKB
 #include "inputstr.h"
-#include <xkbsrv.h>
+#include "xkbsrv.h"
 #endif
 
 #define mskcnt ((MAXCLIENTS + 31) / 32)
--- a/nx-X11/programs/Xserver/dix/dispatch.c
+++ b/nx-X11/programs/Xserver/dix/dispatch.c
@@ -109,7 +109,6 @@ int ProcInitialConnection();
 #include <nx-X11/extensions/security.h>
 #endif
 #ifdef XKB
-#define XKB_IN_SERVER
 #include "inputstr.h"
 #include <xkbsrv.h>
 #endif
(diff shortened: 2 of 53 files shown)


nx-X11/programs/Xserver/dix/devices.c
```diff
@@ -63,7 +63,7 @@ SOFTWARE.
 #include "site.h"
 #define	XKB_IN_SERVER
 #ifdef XKB
-#include <nx-X11/extensions/XKBsrv.h>
+#include <xkbsrv.h>
```
sunweaver: Please use "xkbsrv.h" instead here.

nx-X11/programs/Xserver/dix/dispatch.c
```diff
@@ -111,7 +111,7 @@ int ProcInitialConnection();
 #ifdef XKB
 #define XKB_IN_SERVER
 #include "inputstr.h"
-#include <nx-X11/extensions/XKBsrv.h>
+#include <xkbsrv.h>
```
sunweaver: same here... (quotation marks instead of pointed brackets)

nx-X11/programs/Xserver/dix/events.c
```diff
@@ -129,7 +129,7 @@ of the copyright holder.
 #include "globals.h"
 
 #ifdef XKB
-#include <nx-X11/extensions/XKBsrv.h>
+#include <xkbsrv.h>
```
sunweaver: same here...

nx-X11/programs/Xserver/dix/devices.c
sunweaver: is it correct to have no captial letters in this filename? XKBsrv.h vs. xkbsrv.h?

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -75,6 +75,7 @@ is" without express or implied warranty.
 
 #define XKBSRV_NEED_FILE_FUNCS
 #include <nx-X11/extensions/XKBsrv.h>
+#include <nx-X11/extensions/XKBstr.h>
```
sunweaver: and why XKBsrv.h here? As I understand,the XKBsrv.h version is libX11 whereas the xkbsrv.h is Xserver. Which one is actually needed here? (I feel, it should be xkbsrv.h here, but that is just a gut feeling).

nx-X11/programs/Xserver/os/utils.c
```diff
@@ -141,7 +141,7 @@ OR PERFORMANCE OF THIS SOFTWARE.
 #endif
 
 #ifdef XKB
-#include <nx-X11/extensions/XKBsrv.h>
+#include <xkbsrv.h>
```
sunweaver: same here... (I won't mark any more of these, please make sure to fix all occurrences)

sunweaver (changes requested):

sunweaver:
uli42: thanks for filing this PR. Please go through my individual comments. When done, please make sure you rebase this PR branch against 3.6.x. There are some commits that are already available on the 3.6.x branch, so they should not appear in this PR anymore.

THANKS!
Mike

uli42:
Regarding quotes vs brackets: I wanted to stay as close to upstream as possible. Maybe I overlooked some places, will re-check.

Also the naming of the header files is from upstream. 

The two files with XKBsrv.h need the client side headers. The reason for this is that on 64bit libX11 delivers structures that differ from the ones the identically named Xserver functions deliver. E.g. we use XkbGetKeyboard(), which is only existing in libX11 and we use XkbFreeKeyboard(), which is available both from libX11 and Xserver. So we must ensure to call the matching Free() function. If we use the server header file nxagent crashes immediately (since XKBFreeKeyboard() from the server is called) 

Interestingly upstream Xnest does use the server header file which I consider a bug. I have not yet understood why Xnest does not crash here. Maybe the free functions are more tolerant in upstream nowadays.

sunweaver:
> Regarding quotes vs brackets: I wanted to stay as close to upstream as possible. Maybe I overlooked 
> some places, will re-check.

Pointed brackets really look incoherent here. Maybe we should not adopt that from X.org.

> Also the naming of the header files is from upstream.

This is ok then. I just wanted to make sure filenames are not mixed up.

> [...]

I trust your judement on the rest of your comment.

Thanks!
Mike

sunweaver:
Make sure, you really rebase your branch against latest 3.6.x. I pushed some merges and close some PRs.

nx-X11/programs/Xserver/dix/dispatch.c
uli42: I have now checked includes and adapted to upstream. While you are probably right in respect to quotes against brackets I prefer to stay as close as possible to upstream. Upstream is not really coherent in that respect, too...

uli42:
I have now checked includes and adapted to upstream. While you are probably rightregarding quotes against brackets I prefer to stay as close as possible to upstream (this helps a lot when upgrading/backporting). Upstream is not really coherent in that respect, too...

sunweaver (approved):

PR #210, opened 2016-10-01
Fix protoheader mess (randr, damage)

### LINKED ISSUE #205: randrproto.h references X11/ not nx-X11 - why?
nx-X11/include/extensions/randrproto.h includes randr.h - but from the X11 namespace, not the nx-X11 namespace. #200 changed behaviour regarding X11/.. paths and we have randr.h in the nx-X11 namespace, too and that one will be taken now.

So the question arises: Is that a bug/mistake/flaw or is this on purpose?

`#include <X11/extensions/randr.h>`

sunweaver:
Ulrich Sibiller discovered that nx-libs includes some proto headers from X.org where we actually should use our own header versions. This PR fixes two issues of that kind (one for the RandR extension, one for the Damage extension).

This fixes ArcticaProject/nx-libs#205. (Make sure to mention this bug closure in the merge comment of this PR).


sunweaver:
@uli42: does this PR fix the brought-up concerns described in #205?

uli42:
Yes, this looks good, except for one thing, see my other comment.

nx-X11/include/extensions/randrproto.h
```diff
@@ -1,35 +1,38 @@
...
+ *	    Keith Packard, Intel Corporation
  */
 
 /* note that RANDR 1.0 is incompatible with version 0.0, or 0.1 */
 /* V1.0 removes depth switching from the protocol */
 #ifndef _XRANDRP_H_
 #define _XRANDRP_H_
 
-#include <X11/extensions/randr.h>
+#include "randr.h"
```
uli42: Why not `<nx-X11/extensions/randr.h>`?

uli42 (changes requested):

nx-X11/include/extensions/randrproto.h
sunweaver: Indeed!

sunweaver: Will fix this before merging.

PR #218, opened 2016-10-07
Fix unixpath crash

vatral:
Fixes bug #217


sunweaver (changes requested):
Please fix the commit history. It seems you are actually only proposing a single commit, but the PRs history contains two merges and another commit (fix missing #endif).

Please propose a PR that contains a clean Git history and that is properly rebased against the 3.6.x branch.

If you commit closes any of the open bugs, please reference the bug closure in the commit message (opposed to mentioning the closure in the PR text).

sunweaver:
vatral: actually I cherry-picked your commit and pushed directly to 3.6.x. No more work needed for #218.

PR #214, opened 2016-10-05
Pr/upgrade lib x11

uli42:
New attempt to upgrade to libX11 1.3.4

This is the result of splitting up and reworking #195. This pr only concentrates on libX11 where the previous attempt also changed parts of xserver.

PR #221, opened 2016-10-09
Pr/minor stuff

uli42:
Some small commits I came across while working on other things


.gitignore
```diff
@@ -0,0 +1,387 @@
...
+nx-X11/programs/Xserver/hw/nxagent/miinitext.c
+nx-X11/programs/Xserver/hw/nxagent/stubs.c
+nx-X11/programs/Xserver/hw/nxagent/xpstubs.c
+nx-X11/programs/Xserver/include/compiler.h
+nx-X11/programs/Xserver/include/osdep.h
+nx-X11/programs/Xserver/include/xf86Version.h
+nx-X11/programs/Xserver/include/xf86_ansic.h
+nx-X11/programs/Xserver/include/xf86_libc.h
+nx-X11/programs/Xserver/include/xkbfile.h
+nx-X11/programs/Xserver/xkb/modinit.h
```
sunweaver, line 387: @uli42: personally, I am not a fan of .gitignore files at all. I think .gitignore hides our flaws. I much more prefer a good make clean target.

vatral also offered a .gitignore file that he personally uses with nx-libs. If you guys feel like we should have one, I'd bow to that request (and remove it on my local working copies).

nx-X11/lib/X11/Pending.c
```diff
@@ -24,7 +24,9 @@ in this Software without prior written authorization from The Open Group.
 
 */
 
+#if defined(NX_TRANS_SOCKET) && defined(NX_TRANS_TEST)
 #include <stdio.h>
+#endif
```
sunweaver, line 29: Shouldn't this go into the libX11 -> 1.3.4 PR? If so, please add it there with our current iteration.

nxcomp/Misc.cpp
```diff
@@ -326,7 +326,7 @@ defer=n\n\
 tile=s\n\
 menu=n\n\
 sleep=n\n\
-tolerancecehcks=s\n\
+tolerancechecks=s\n\
```
sunweaver, line 329: Good catch.

sunweaver (changes requested):
Thanks for that bundle PR.

Let's wait for @vatral and @Ionic to comment on the usefulness of .gitignore.

Then please move the commit on X11/Pending.c into #214.

Then please file the typo fix and the .gitignore (if we agree on having it) as two different PRs.

May this PR be for .gitignore, and a new PR for the typo fix.

/me personally prefers PRs to be more atomic / topic based.

nx-X11/lib/X11/Pending.c
uli42: No, this has nothing to do with libX11 upgrade, it is just a clearer separation of NX changes from upstream changes.

.gitignore
uli42: As stated in IRC It is very helpful for the way I work AND it is NOT covering flaws in make clean/distclean!

Ionic: @sunweaver's point is that after running `clean/distclean`, all stray files should be gone and running `git status` is a way to list anything untracked. If this doesn't come up empty, we've forgotten some stuff.

While that might be true, I'm more often than not also using `git status` in an unclean state and having a `.gitignore` file certainly helps with the huge output...

Btw, did you know that every (sub-)directory can have its own `.gitignore` file? Instead of putting everything into the "root" file, I think I'll split it up.

uli42:
OK, splitting it up makes sense.

sunweaver:
@uli42: I have just cherry-picked the two non-gitignore commits from this PR. Thus closing this PR. For the split-up .gitignore files, please file a separate new PR. Thanks.

PR #232, opened 2016-10-18
Slave channel implementation

vatral:
This is the slave channel implementation we use at QVD. 

When the channel is enabled, and the NX_SLAVE_CMD environment variable is set to a command, the agent or the proxy listens on a port, and the connection is directed to the opposite side, where the command is invoked and a bidirectional channel is created to it.

### DIFF
--- a/nxcomp/Misc.cpp
+++ b/nxcomp/Misc.cpp
@@ -358,6 +358,13 @@ tolerancechecks=s\n\
                ory compared to the other programs, to make easier\n\
                for the user to execute the program from the shell.\n\
 \n\
+  NX_SLAVE_CMD The full path to the slave channel handler. When the\n\
+               slave channel is enabled, the agent will listen on a\n\
+               port and forward the connection to the NX_SLAVE_CMD\n\
+               program. This can be used to implement agent/proxy\n\
+               communication for applications such as serial port and\n\
+               USB forwarding.\n\
+\n\
   Shell environment:\n\
 \n\
   HOME         The variable is checked in the case NX_HOME is not\n\
--- a/nxcomp/Proxy.cpp
+++ b/nxcomp/Proxy.cpp
@@ -29,6 +29,8 @@
 #include <string.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <signal.h>
+
 #ifdef ANDROID
 #include <netinet/in.h>
 #include <netinet/ip.h>
@@ -91,6 +93,8 @@ struct sockaddr_un
 //
 
 extern void CleanupListeners();
+
+extern int HandleChild(int);
 
 //
 // Default size of string buffers.
@@ -165,6 +169,7 @@ Proxy::Proxy(int fd)
 
     fdMap_[channelId]      = nothing;
     channelMap_[channelId] = nothing;
+    slavePidMap_[channelId] = nothing;
   }
 
   inputChannel_  = nothing;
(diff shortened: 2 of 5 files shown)


uli42:
On Tue, Oct 18, 2016 at 9:29 AM, Vadim Troshchinskiy

> This is the slave channel implementation we use at QVD.
> 
> When the channel is enabled, and the NX_SLAVE_CMD environment variable is set to a command, the agent or the proxy listens on a port, and the connection is directed to the opposite side, where the command is invoked and a bidirectional channel is created to it.

What are you using it for?

Uli

nxcomp/Proxy.cpp
```diff
@@ -6308,42 +6311,94 @@ int Proxy::handleNewGenericConnectionFromProxyUnix(int channelId, T_channel_type
...
 
-  #ifdef WARNING
-  *logofs << "Proxy: Refusing new slave connection for "
-          << "channel ID#" << channelId << "\n"
-          << logofs_flush;
-  #endif
+  cerr << "Info" << ": New slave connection on "
+       << "channel ID#" << channelId << "\n";
+
+  char *qvd_slave_cmd = getenv("NX_SLAVE_CMD");
```
sunweaver: No QVD, X2Go, Arctica specific var names, please. Let's keep nx-libs independent (regarding namespacing) from our products.

sunweaver (changes requested):
Hi Vadim, I'd also be interested in how you use this patch. I don't see any problems with merging it except from the variable naming commented on above.

sunweaver:
@vatral: You also had a merge commit in your PR. Please provide a "clean" PR branch! Rebased against latest 3.6.x branch and without branching cruft. Thanks!

nxcomp/Proxy.cpp
vatral: Oops. I fixed the environment var, but forgot about the variable name.

By the way, does using an environment variable sound good to you?

vatral:
@uli42, @sunweaver 

We use it for communication between the VM and the client side. A connection can be established either way, client to VM, or VM to client. The handler we use is a small HTTP server that implements multiple commands.

It's being used for USB forwarding, TCP port forwarding, serial port forwarding, file sharing over sshfs, checking the version on the other side, and has a few optional debug commands (equivalents of ping, echo, discard and chargen) for network testing.

We recently found an issue with session suspend, so I will be updating this with a fix soon.

sunweaver:
@vatral: please clean up the PR branch (no merges on the branch as part of the PR).

And also, as an idea... please add some documentation of the NX_SLAVE_CMD env var to the nxproxy man page and the --help output of nxproxy. Thanks.

vatral:
@sunweaver Rebase done, manpage edited.

sunweaver:
@vatral: please rebase once more against latest 3.6.x branch. This PR is next in my queue...

sunweaver:
Hi Vadim,

I looked through this PR once more. Do I see it correctly, that you can attach a slave cmd to each channel of an nxcomp based application (i.e., proxy / agent)? Is it then sufficient to have one NX_SLAVE_CMD env variable?

This seems that NX_SLAVE_CMD must be a multi plexer, right? How does the NX_SLAVE_CMD then know to what channel it is attached?

Can you add an example script to testscripts that illustrates this feature?

nxcomp/Misc.cpp
```diff
@@ -358,6 +358,13 @@ tolerancechecks=s\n\
                ory compared to the other programs, to make easier\n\
                for the user to execute the program from the shell.\n\
 \n\
+  NX_SLAVE_CMD The full path to the slave channel handler. When the\n\
+               slave channel is enabled, the agent will listen on a\n\
+               port and forward the connection to the NX_SLAVE_CMD\n\
```
sunweaver, line 363: Shouldn't we replace "port" by "socket" here?  The word port may be misleading and people assume, that there will be some IP port being opened when using this NX_SLAVE_CMD feature.

vatral: No, it's actually a TCP port

vatral: And yes, the command is a multiplexer. In our particular usage, it's an HTTP server. So it works like this:

GET /ping
Pong!

POST /tcp/connect/4000
(slave handler execs socat, now the connection starts talking to port 4000)

I made a little test script that works on the same principle, just without HTTP and without using our libraries. I will add it to testscripts.

PR #236, opened 2016-10-19
nxcomp: Support the Generic Event Extension.

sunweaver:
I have extracted the Generic Event Extension part from the patch provided via Debian BTS: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=766299

However, I don't know how we can actually test that it works. Any clues?

@Ionic: @uli42: @vatral: @mbecroft: ?

Thanks for feedback!
Mike

### DIFF
--- a/nxcomp/ClientChannel.cpp
+++ b/nxcomp/ClientChannel.cpp
@@ -4808,6 +4808,12 @@ FIXME: Recover the sequence number if the proxy
           break;
         default:
           {
+            // BEWARE: not only inputOpcode == GenericEvent but also
+            // others not handled above, at least:
+            //   GraphicsExpose    13
+            //   MapRequest        20
+            //   ConfigureRequest  23
+            // and any beyond LASTEvent.
             #ifdef TEST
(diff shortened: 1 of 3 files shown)


sunweaver:
Adding more devs for feedback: @woglinde: ?, @salva: ?

Ionic:
No idea, really, other than finding an application that spews out generic events by itself or masked via xcb/Xlib.

uli42:
https://www.x.org/releases/X11R7.6/doc/xextproto/geproto.html ist the specificiation of GenericEvent. Maybe we can ask its author Peter Hutterer for some example and/or testing code. I have not found anything like that.

uli42:
I just noticed that the debian BTS link above contains hints that Nautilus (in the version current back then) uses the GenericEventsExtension (but without properly checking if it's there before using it).

uli42:
@sunweaver Your branch only contains changes to nxcomp. But I think we also need to backport the Generic Events Extension to the xserver for this to be relevant at all.

uli42:
The extension can be tested with atom, which seems to require it.

Issue #239, opened 2016-10-25
keyboard config file


uli42:
in `hw/nxagent/Keyboard.c` a file containing the keyboard config of the real X server is created in the agent's session dir, called `keyboard`. This file can be fed to `setxkmap` using the `-config` option:

```
/usr/bin/setxkbmap -config keyboard
```

Currently this does not work because setxkbmap is a bit picky about the correct format and will issue a `Couldn't find configuration file "/tmp/keyboard"` error message (which really means `Couldn't parse configuration file`...). 

This is how my file looks:

```
rules=base
model=evdev
layout=de
variant=nodeadkeys
options=
```

The cuprit is the last line: it does not have a right side. Simply adding double quotes helps:

```
rules=base
model=evdev
layout=de
variant=nodeadkeys
options=""
```

We should enhance the code to always contain double quotes for every line.

Update: http://unix.stackexchange.com/questions/167131/how-to-use-the-config-option-of-setxkbmap indicates that empty options will not reset the current setting and you need to add a single comma to achieve that. So the above example should become:

```
rules="base"
model="evdev"
layout="de"
variant="nodeadkeys"
options=","
```

I have checked that and the option are getting reset using this syntax.

(BTW: setxkbmap does not complain when the double quotes are in the file. But the keyboard is still not working correctly for me on RHEL6 using the keyboard definitions in `/usr/share/X11/xkb`. I am still investigating that issue)

PR #243, opened 2016-10-27
Pr/keyboard fixes

uli42:
some cleanups and a fix for the (never working) nxagent.keyboard config file.

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
@@ -173,8 +173,6 @@ extern        Status        XkbGetControls(
 #define XKB_DFLT_KB_OPTIONS  NULL
 #endif
 
-#define NXAGENT_KEYMAP_DIR_FILE "keymap.dir"
-
 extern int XkbDfltRepeatDelay;
 extern int XkbDfltRepeatInterval;
 
--- a/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
@@ -697,10 +697,6 @@ int nxagentKeyboardProc(DeviceIntPtr pDev, int onoff)
   int free_model = 0, free_layout = 0;
   XkbDescPtr xkb = NULL;
 
-  #ifdef TEST
-  int ret;
-  #endif
-
   switch (onoff)
   {
     case DEVICE_INIT:
@@ -857,13 +853,15 @@ XkbError:
              (char *) values.auto_repeats, sizeof(values.auto_repeats));
 
       #ifdef TEST
-      ret =
+      {
+        int ret =
       #endif
-      InitKeyboardDeviceStruct((DevicePtr) pDev, &keySyms, modmap,
+        InitKeyboardDeviceStruct((DevicePtr) pDev, &keySyms, modmap,
                                nxagentBell, nxagentChangeKeyboardControl);
 
       #ifdef TEST
-      fprintf(stderr, "nxagentKeyboardProc: InitKeyboardDeviceStruct returns [%d].\n", ret);
+        fprintf(stderr, "nxagentKeyboardProc: InitKeyboardDeviceStruct returns [%d].\n", ret);
+      }
       #endif
 
 #ifdef XKB
--- a/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
@@ -1863,10 +1863,9 @@ void nxagentKeycodeConversionSetup(void)
     fprintf(stderr, "Failed to create the keyboard file\n");
   }
 
-  if (nxagentOption(ClientOs) == ClientOsLinux &&
-            drules != NULL && dmodel != NULL &&
-                (strcmp(drules, "evdev") == 0 ||
-                    strcmp(dmodel, "evdev") == 0))
+  if (drules != NULL && dmodel != NULL &&
+      (strcmp(drules, "evdev") == 0 ||
+       strcmp(dmodel, "evdev") == 0))
   {
     #ifdef DEBUG
     fprintf(stderr, "nxagentKeycodeConversionSetup: "
--- a/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
@@ -1841,15 +1841,15 @@ void nxagentKeycodeConversionSetup(void)
       strcat(keyboard_file_path, "/keyboard");
       if ((keyboard_file = fopen(keyboard_file_path, "w")) != NULL) {
         if ( drules != NULL )
-          fprintf(keyboard_file, "rules=%s\n", drules);
+          fprintf(keyboard_file, "rules=\"%s\"\n", drules[0] == '\0' ? "," : drules);
         if ( dmodel != NULL )
-          fprintf(keyboard_file, "model=%s\n", dmodel);
+          fprintf(keyboard_file, "model=\"%s\"\n", dmodel[0] == '\0' ? "," : dmodel);
         if ( dlayout != NULL )
-          fprintf(keyboard_file, "layout=%s\n", dlayout);
+          fprintf(keyboard_file, "layout=\"%s\"\n", dlayout[0] == '\0' ? "," : dlayout);
         if ( dvariant != NULL )
-          fprintf(keyboard_file, "variant=%s\n", dvariant);
+          fprintf(keyboard_file, "variant=\"%s\"\n", dvariant[0] == '\0' ? "," : dvariant);
         if ( doptions != NULL )
-          fprintf(keyboard_file, "options=%s\n", doptions);
+          fprintf(keyboard_file, "options=\"%s\"\n", doptions[0] == '\0' ? "," : doptions);
         fclose(keyboard_file);
         fprintf(stderr, "keyboard file created\n");
       }
--- a/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
@@ -871,8 +871,6 @@ XkbError:
         FILE *file;
         XkbConfigRtrnRec config;
 
-        int nxagentXkbConfigFilePathSize;
-
         char *nxagentXkbConfigFilePath;
 
         XkbComponentNamesRec names;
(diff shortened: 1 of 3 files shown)


nx-X11/programs/Xserver/hw/nxagent/Reconnect.c
```diff
@@ -438,22 +438,10 @@ Bool nxagentReconnectSession(void)
 
   if (nxagentKeyboard != NULL)
   {
-    int size;
+    nxagentOldKeyboard = strndup(nxagentKeyboard, strlen(nxagentKeyboard));
+    /* FIXME: check if nxagentOldKeyboard == NULL and abort */
```
Ionic: The old code hasn't made any effort to validate `nxagentOldKeyboard` in the first place, but while we're at it anyway, care to remove the `FIXME` and do what you described there?

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1863,10 +1863,9 @@ void nxagentKeycodeConversionSetup(void)
     fprintf(stderr, "Failed to create the keyboard file\n");
   }
 
-  if (nxagentOption(ClientOs) == ClientOsLinux &&
-            drules != NULL && dmodel != NULL &&
-                (strcmp(drules, "evdev") == 0 ||
-                    strcmp(dmodel, "evdev") == 0))
+  if (drules != NULL && dmodel != NULL &&
```
Ionic, line 1866: Mh, careful, at least when it comes to the commit message. `evdev` will also be supported on FreeBSD once a new version is released, so we can't classify it as "Linux-only". This said, there's no `clientOs` value for *BSD currently, so I assume it will default to `UNDEFINED (-1)` on that platform.

Ionic: Even worse, while the server machine might have `rules` and `model` defined to `evdev` (because we currently really only support Linux as a server OS only in the first place), this change will also enable keycode conversion for OS X and Windows **clients** (i.e., the remote counterpart) which differs from the old behavior. That may or may not be a good thing...

nx-X11/programs/Xserver/hw/nxagent/Reconnect.c
uli42: Well, I have not found a quick way to abort cleanly here. Besides: This is the reconnect part, do we really want to kill the session in that case?

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: I had googled for evdev and not seen any FreeBSD match. And do we support FreeBSD? And keycode conversion might still be needed for FreeBSD.

Note: This change will only activate keycode conversion if the _client_ side has evdev. And I doubt that OS X or Windows offer that.

Speaking of that: I have a branch that contains a POC that automatically clones the settings of the client. And that disables keycode conversion if both server and client use evdev. It would be interesting if that works on OS X (I can test Windows myself, but not OS X). Here's the branch (I call it POC because the code is ugly and has to be improved/redone): https://github.com/uli42/nx-libs/tree/autokeyboard You need to call it with `-keyboard auto` or `keyboard=auto` in options.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1841,15 +1841,15 @@ void nxagentKeycodeConversionSetup(void)
...
         if ( dlayout != NULL )
-          fprintf(keyboard_file, "layout=%s\n", dlayout);
+          fprintf(keyboard_file, "layout=\"%s\"\n", dlayout[0] == '\0' ? "," : dlayout);
         if ( dvariant != NULL )
-          fprintf(keyboard_file, "variant=%s\n", dvariant);
+          fprintf(keyboard_file, "variant=\"%s\"\n", dvariant[0] == '\0' ? "," : dvariant);
         if ( doptions != NULL )
-          fprintf(keyboard_file, "options=%s\n", doptions);
+          fprintf(keyboard_file, "options=\"%s\"\n", doptions[0] == '\0' ? "," : doptions);
         fclose(keyboard_file);
```
sunweaver, line 1846: Is this really what we want here?

```
if d<param>[0] is '\0' -> then write "," (i.e. a komma) as value for d<param> to the keyboard config file, otherwise d<param>.
```

Shouldn't it rather be...

```
if d<param>[0] is '\0' -> then write "\"\"" (i.e. two quotes with nothing inbetween) as value for d<param> to the keyboard config file, otherwise d<param>.
```

??? Or am I getting somethign wrong here?

sunweaver: Ah... no... you even write:

```
d<param>=","
```

to the config file... Why the komma???

nx-X11/programs/Xserver/xkb/ddxLoad.c
```diff
@@ -767,36 +767,6 @@ char tmpname[PATH_MAX];
...
-                    sprintf(name,"%s%s.xkm", xkm_output_dir, keymap);
-		for (i = 0; i < 10; i++) {
-	            if (access(name, 0) == 0) break;
-		    usleep(500000);
-		}
-#ifdef DEBUG
-		if (i) ErrorF(">>>> Waited %d times for %s\n", i, name);
-#endif
-	    }
-#endif
```
sunweaver, line 799: @uli42: Please add Backported-to-nx-libs... to the git commit msg. Thanks.

nx-X11/programs/Xserver/hw/nxagent/Reconnect.c
sunweaver: nxagentOldKeyboard is a local var. And it gets set in the code section above. nxagentKeyboard comes in via the cmdline (see Args.c). It indeed never gets properly sanitized. But that is something for another PR. And it should be done in Args.c, right?

sunweaver: My intention is... nxagentOldKeyboard does not required to be sanitized. NULL checks are neither required (see the code below in Reconnect.c).

uli42:
You are getting it wrong. If the param is empty (two quotes with nothing inbetween) this has no effect, exusting settings will stay! But we want to set it to nothing, t.i _reset_the existing settings, so we need the comma. See my description in the issue text.

Vom Smartphone gesendet.

Von: "Mike Gabriel" <notifications@github.com>
Gesendet: ‎28.‎10.‎2016 13:41
Betreff: Re: [ArcticaProject/nx-libs] Pr/keyboard fixes (#243)

@sunweaver commented on this pull request.



In nx-X11/programs/Xserver/hw/nxagent/Keyboard.c:
>          if ( doptions != NULL )
-          fprintf(keyboard_file, "options=%s\n", doptions);
+          fprintf(keyboard_file, "options=\"%s\"\n", doptions[0] == '\0' ? "," : doptions);
         fclose(keyboard_file);
Is this really what we want here?
if d<param>[0] is '\0' -> then write "," (i.e. a komma) as value for d<param> to the keyboard config file, otherwise d<param>.
Shouldn't it rather be...
if d<param>[0] is '\0' -> then write "\"\"" (i.e. two quotes with nothing inbetween) as value for d<param> to the keyboard config file, otherwise d<param>.
??? Or am I getting somethign wrong here?

uli42:
No, what this is about is strdup(). It performs an malloc. If that fails strdup will return Null and this should be checked.

Vom Smartphone gesendet.

Von: "Mike Gabriel" <notifications@github.com>
Gesendet: ‎28.‎10.‎2016 13:58
Cc: "Ulrich Sibiller" <uli42@gmx.de>; "Mention" <mention@noreply.github.com>
Betreff: Re: [ArcticaProject/nx-libs] Pr/keyboard fixes (#243)

@sunweaver commented on this pull request.



In nx-X11/programs/Xserver/hw/nxagent/Reconnect.c:
> @@ -438,22 +438,10 @@ Bool nxagentReconnectSession(void)
 
   if (nxagentKeyboard != NULL)
   {
-    int size;
+    nxagentOldKeyboard = strndup(nxagentKeyboard, strlen(nxagentKeyboard));
+    /* FIXME: check if nxagentOldKeyboard == NULL and abort */
nxagentOldKeyboard is a local var. And it gets set in the code section above. nxagentKeyboard comes in via the cmdline (see Args.c). It indeed never gets properly sanitized. But that is something for another PR. And it should be done in Args.c, right?

sunweaver:
@uli42: Don't forget to close https://github.com/ArcticaProject/nx-libs/issues/239 with one of the commits in this PR.

nx-X11/programs/Xserver/hw/nxagent/Reconnect.c
uli42: No, what this is about is strndup(). It performs an malloc. If that fails strndup will return NULL and _this_ should be checked.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: You are getting it wrong. If a param in the config file is empty (two quotes with nothing inbetween) this has no effect, existing settings will stay! But we want to set it to nothing, t.i. _reset_the existing settings, so we need the comma. See my explanation in #243 (lower part).

nx-X11/programs/Xserver/xkb/ddxLoad.c
uli42: done

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: One more comment: _if_ FreeBSD uses evdev it might be compatible to the Linux version (at least i'd expect this for identically named components). So again no reason to check the remote OS.

sunweaver:
@uli42: where are we regarding PR #243 (keyboard fixes). Can you rebase and check and hand it over for a second review? There was a FIXME Ionic commented on that you might want to fix / address / discuss further?

uli42:
I am still waiting for a further comment by @Ionic. Regarding evdev os check and regarding NOT checking nxagentOldKeyboard

nx-X11/programs/Xserver/hw/nxagent/Reconnect.c
Ionic: Just use

``` C
  FatalError ("strndup failed for the keyboard settings");
```

or the like. A failing `strndup ()` is pretty much a death sentence anyway, unless the string to be copied is unreasonable big. Which, frankly, would be a problem on its own...

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
Ionic: Yeah, it should be a compatible implementation. I missed the rules and model being the client-side values, so... without proper `evdev` support in the kernel, we're likely not going to see `evdev` anytime soon as these key's values any time soon. I guess it's fine then.

nx-X11/programs/Xserver/hw/nxagent/Reconnect.c
uli42: I am still hesitating since the low memory situation may be temporary. Your suggestion would kill the whole session which is not desirable when we are reconnecting!

We might rather jump to `nxagentReconnectError` but then we'd have to add another `failedStep` value. And check if that has to be handled at different locations, too.

sunweaver:
Hi,

On  Do 03 Nov 2016 11:11:56 CET, Ulrich Sibiller wrote:

> 
> > @@ -438,22 +438,10 @@ Bool nxagentReconnectSession(void)
> 
>    if (nxagentKeyboard != NULL)
>    {
> -    int size;
> -    nxagentOldKeyboard = strndup(nxagentKeyboard, strlen(nxagentKeyboard));
> -    /\* FIXME: check if nxagentOldKeyboard == NULL and abort */
> 
> I am still hesitating since the low memory situation may be  
> temporary. Your suggestion would kill the whole session which is not  
> desirable when we are reconnecting!

Maybe, just abort the session resumption process and don't touch the  
running/suspended nxagent.

## Mike

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

nx-X11/programs/Xserver/hw/nxagent/Reconnect.c
Ionic: _shrug_ then keep it as-is. In any case, allocation failures are likely to not go away and cascade, if you don't free other data first. An allocation failure need not mean that no more memory is available, it's also possible that an application broke the allocator by writing random data to the stack or the heap. I have no idea what happens on allocation failures (for instance, at what point nxagent dies), never actually tested it.

uli42: I have just tried to add another failedStep and then saw, that this keyboard stuff is done before the first failedStep code. Inbetween there are further initialisations which are referred in  `nxagentReconnectError` so we cannot simply jump there. I have (for now?) simpy added a `return 0` (the same is done at the end of `nxagentReconnectError`)

uli42:
from my POV this can be merged.

Ionic:
Please rebase against current 3.6.x and I'll merge it.

uli42:
done

PR #253, opened 2016-10-29
drop Xserver/fb/fbmmx.[ch]

sunweaver:
This PR drops MMX code that upstream replaced by pixman_\* calls. We don't use pixman_\* calls, but we did not use MMX in Xserver/fb/ either (USE_MMX has never been defined, so far).

This PR drops the fbmmx.[ch] related code now, later on we should consider bumping the Xserver/fb code to latest X.org version status.


sunweaver:
@Ionic: can you review this PR, please (#253). Thanks!

nx-X11/programs/Xserver/fb/Imakefile
```diff
@@ -1,22 +1,3 @@
-/*
```
Ionic, line 1: Leave out "(, yet)" in your commit message, you already included a yet earlier. :)

nx-X11/programs/Xserver/fb/fbpict.c
```diff
@@ -901,11 +885,6 @@ fbComposite (CARD8      op,
 			switch (pDst->format) {
 			case PICT_r5g6b5:
 			case PICT_b5g6r5:
-#ifdef USE_MMX
-			    if (fbHaveMMX())
-				func = fbCompositeSolidMask_nx8x0565mmx;
-			    else
-#endif
 				func = fbCompositeSolidMask_nx8x0565;
```
Ionic: Theoretically, you could move the the call following the `USE_MMX` block one indentation level to the left, but that's really just an optical thing.

Would apply to all the following change blocks as well.

Ionic (commented):
Otherwise looks good.

nx-X11/programs/Xserver/fb/Imakefile
sunweaver: > Leave out "(, yet)" in your commit message, you already included a yet earlier. :)

Done.

nx-X11/programs/Xserver/fb/fbpict.c
sunweaver: > Theoretically, you could move the the call following the USE_MMX block one indentation level to the left, but that's really just an optical thing.

Done for all such blocks.

sunweaver:
@Ionic: I force-pushed my changes to the PR branch. Ready for merging. Thanks!

PR #270, opened 2016-11-01
Drop Xserver/GL/include.

sunweaver:
Replacing functions from glx_ansic.h by same functions as provided in stdlib.h and string.h. Also drop xf86glx.h.


sunweaver:
@Ionic: Can you please review #270 as paid work? Thanks!

nx-X11/programs/Xserver/GL/glxmodule.c
```diff
@@ -780,7 +780,7 @@ static const char *glcoreSymbols[] = {
     "__glXFogfvReqSize",
     "__glXFogivReqSize",
     "__glXForceCurrent",
-    "__glXFree",
+    "free",
```
uli42: glxmodule.c is dropped in my unused_files commit anyway...

uli42 (commented):
LGTM

nx-X11/programs/Xserver/GL/glxmodule.c
sunweaver: Yeah. Actually a find+replace bug. But anyway.

nx-X11/programs/Xserver/GL/glx/glxcmds.c
```diff
@@ -48,7 +48,6 @@
 #include "glximports.h"
```
Ionic, line 51: The original commit adds 

``` C
#include <assert.h>
```

and already has an include for `string.h`. We'd probably need this to, since the code is using `assert ()` calls.

nx-X11/programs/Xserver/GL/glx/glxcmds.c
```diff
@@ -2139,7 +2139,7 @@ int __glXQueryServerString(__GLXclientState *cl, GLbyte *pc)
     if ((buf = (char *) malloc(length << 2)) == NULL) {
         return BadAlloc;
     }
-    __glXStrncpy(buf, ptr, n);
```
Ionic, line 2144: The original commit switched to `memcpy ()` here.

nx-X11/programs/Xserver/GL/glx/glxdrawable.h
```diff
@@ -93,7 +93,7 @@ struct __GLXdrawablePrivateRec {
     /*
     ** "methods" that the drawble should be able to respond to.
     */
-    void (*freeBuffers)(struct __GLXdrawablePrivateRec *);
```
Ionic, line 96: This change doesn't make sense. The other function pointers are not renamed to `__glX*`, so why this one?

In the end, it's likely a moot point as this code will go away with another cleanup round sooner or later, as upstream dropped it, too, but why make a useless change?

nx-X11/programs/Xserver/GL/glx/glximports.c
```diff
@@ -40,7 +40,6 @@
 #include "glxserver.h"
```
Ionic, line 44: Missing headers, like `string.h`, `stdlib.h`, `stdio.h`?

nx-X11/programs/Xserver/GL/include/GL/xf86glx.h
```diff
@@ -1,42 +0,0 @@
-/**************************************************************************
```
Ionic, line 1: Within this commit, we could probably drop `-Iinclude` from `nx-X11/programs/Xserver/GL/Imakefile`, as this directory is gone now.

Ionic (commented):
That's all for now.

PR #279, opened 2016-11-04
LibX11: drop unused xlibi18n

uli42:
remove support for _loadable_ `xlibli18n`

As i18n is used by the application and we are the ones controlling the application (nxagent) it might also be possible to drop that i18n stuff completely! I have not investigated further, though.


nx-X11/config/cf/X11.tmpl
```diff
@@ -126,9 +126,6 @@ XORGRELSTRING = XorgManVersionString
 #define BuildX11Lib		BuildLibraries
 #endif
 
-#ifndef BuildLoadableXlibI18n
-#define BuildLoadableXlibI18n	(HasDlopen && SharedLibX11)
-#endif
```
sunweaver, line 131: I just grepped for more places where xlibi18n gets mentioned:

```
 [mike@minobo nx-X11 (uli42-pr/remove_xlibi18n)]$ grep -ri xlibi18n *
config/cf/lnxLib.tmpl:#define SharedXlibi18nReqs $(LDPRELIB) $(XONLYLIB)
config/cf/X11.tmpl: * This disables XLOCALEDIR.  Set it if you're using BuildLoadableXlibI18n,
```

sunweaver: Please amend those and include in this commit, too.

sunweaver (changes requested):
I found two more occurences of xlibi18n (grep -i). Please handle those via this PR, too.

Except from that, this PR is fine.

nx-X11/config/cf/X11.tmpl
uli42: rebased and added the two requested changes

sunweaver (approved):

PR #283, opened 2016-11-07
testscripts: Use non-NoMachine file headers for our testscripts.

sunweaver:
@Ionic: @uli42: @vatral: please check if you are ok with changing file headers in our testscripts to the headers that we use for all other files.

Esp. I wonder if we should leave the trademark information for the NX name in the headers... Ideas?


sunweaver:
Not having seen a veto on this, so merging...

PR #299, opened 2016-12-13
Make nxagent only bind to loopback when requested

vatral:
The current code always binds to localhost, though by default binding to localhost is disabled.

This was breaking QVD.


sunweaver (approved):
Looks good to me.

Please rebase against latest 3.6.x and I'll merge.

vatral:
@sunweaver  Done.

PR #302, opened 2016-12-22
Logging changes for Loop.cpp

vatral:
As requested, the changes have been separated out.

### DIFF
--- a/m4/ax_cxx_compile_stdcxx.m4
+++ b/m4/ax_cxx_compile_stdcxx.m4
@@ -1,0 +1,982 @@
+# ===========================================================================
+#  https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx.html
+# ===========================================================================
+#
+# SYNOPSIS
+#
+#   AX_CXX_COMPILE_STDCXX(VERSION, [ext|noext], [mandatory|optional])
+#
+# DESCRIPTION
+#
+#   Check for baseline language coverage in the compiler for the specified
+#   version of the C++ standard.  If necessary, add switches to CXX and
+#   CXXCPP to enable support.  VERSION may be '11' (for the C++11 standard)
+#   or '14' (for the C++14 standard).
+#
+#   The second argument, if specified, indicates whether you insist on an
+#   extended mode (e.g. -std=gnu++11) or a strict conformance mode (e.g.
+#   -std=c++11).  If neither is specified, you get whatever works, with
+#   preference for an extended mode.
+#
+#   The third argument, if specified 'mandatory' or if left unspecified,
+#   indicates that baseline support for the specified C++ standard is
+#   required and that the macro should error out if no mode with that
+#   support is found.  If specified 'optional', then configuration proceeds
+#   regardless, after defining HAVE_CXX${VERSION} if and only if a
+#   supporting mode is found.
+#
+# LICENSE
+#
+#   Copyright (c) 2008 Benjamin Kosnik <bkoz@redhat.com>
+#   Copyright (c) 2012 Zack Weinberg <zackw@panix.com>
+#   Copyright (c) 2013 Roy Stogner <roystgnr@ices.utexas.edu>
+#   Copyright (c) 2014, 2015 Google Inc.; contributed by Alexey Sokolov <sokolov@google.com>
+#   Copyright (c) 2015 Paul Norman <penorman@mac.com>
+#   Copyright (c) 2015 Moritz Klammler <moritz@klammler.eu>
+#   Copyright (c) 2016 Krzesimir Nowak <qdlacz@gmail.com>
+#
+#   Copying and distribution of this file, with or without modification, are
+#   permitted in any medium without royalty provided the copyright notice
+#   and this notice are preserved.  This file is offered as-is, without any
+#   warranty.
+
+#serial 7
+
+dnl  This macro is based on the code from the AX_CXX_COMPILE_STDCXX_11 macro
+dnl  (serial version number 13).
+
+AX_REQUIRE_DEFINED([AC_MSG_WARN])
+AC_DEFUN([AX_CXX_COMPILE_STDCXX], [dnl
+  m4_if([$1], [11], [ax_cxx_compile_alternatives="11 0x"],
+        [$1], [14], [ax_cxx_compile_alternatives="14 1y"],
+        [$1], [17], [ax_cxx_compile_alternatives="17 1z"],
+        [m4_fatal([invalid first argument `$1' to AX_CXX_COMPILE_STDCXX])])dnl
+  m4_if([$2], [], [],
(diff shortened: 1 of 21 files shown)


nxcomp/Loop.cpp
```diff
@@ -671,15 +671,11 @@ static int ReopenLogFile(char *name, ostream *&stream, int limit);
 static void handleCheckSessionInLoop();
 static void handleCheckBitrateInLoop();
 
-#if defined(TEST) || defined(INFO)
 static void handleCheckSelectInLoop(int &setFDs, fd_set &readSet,
-                                        fd_set &writeSet, T_timestamp selectTs);
-static void handleCheckResultInLoop(int &resultFDs, int &errorFDs, int &setFDs, fd_set &readSet,
+                                        fd_set &writeSet, T_timestamp selectTs);static void handleCheckResultInLoop(int &resultFDs, int &errorFDs, int &setFDs, fd_set &readSet,
```
sunweaver: missing line break here!

nxcomp/Loop.cpp
```diff
@@ -671,15 +671,11 @@ static int ReopenLogFile(char *name, ostream *&stream, int limit);
...
                                         fd_set &writeSet, struct timeval &selectTs,
                                             struct timeval &startTs);
-static void handleCheckStateInLoop(int &setFDs);
-#endif
-
+	static void handleCheckStateInLoop(int &setFDs);
```
sunweaver: don't indent here, make commits as clean as possible, esp. such big ones...

nxcomp/Loop.cpp
```diff
@@ -671,15 +671,11 @@ static int ReopenLogFile(char *name, ostream *&stream, int limit);
 static void handleCheckSessionInLoop();
 static void handleCheckBitrateInLoop();
 
-#if defined(TEST) || defined(INFO)
```
sunweaver: why do you drop ifdef TEST/INFO here?

sunweaver (changes requested):
Please revisit the code passages I commented on. I did not really look deeply at the actual logging implementation, because that has been done by @Ionic on PR #193.

nxcomp/Loop.cpp
Ionic: Because the logging code is not guarded by these tests anymore, so the functions must always be defined to make the code compile.

nxcomp/Loop.cpp
```diff
@@ -14943,27 +13944,20 @@ static void handleCheckBitrateInLoop()
...
 
 static void handleCheckStateInLoop(int &setFDs)
-{
-  int fdLength;
-  int fdPending;
-  int fdSplits;
-
+{  
+  int fdLength;  int fdPending;  int fdSplits;
```
Ionic: Is that change intentional?

nxcomp/Loop.cpp
```diff
@@ -2669,18 +2538,15 @@ int NXTransAlert(int code, int local)
...
     return 1;
   }
-  #if defined(DUMP) || defined(INFO)
   else
   {
     if (logofs == NULL)
     {
-      logofs = &cerr;
-    }
+      logofs = &cerr;    }
```
Ionic: Manual cleanup here.

nxcomp/Loop.cpp
```diff
@@ -3389,33 +3189,27 @@ int SetupProxyConnection()
...
     listenSocket.getSpec(&socketUri);
-    *logofs << "Loop: listenSocket is "<< ( listenSocket.enabled() ? "enabled" : "disabled") << ". "
-            << "The socket URI is '"<< ( socketUri != NULL ? socketUri : "<unset>") << "'.\n" << logofs_flush;
-    free(socketUri);
-    socketUri = NULL;
-    #endif
+    nxinfo << "Loop: listenSocket is "<< ( listenSocket.enabled() ? "enabled" : "disabled") << ". "
+           << "The socket URI is '"<< ( socketUri != NULL ? socketUri : "<unset>") << "'.\n" << std::flush;
 
+    free(socketUri);    socketUri = NULL;
```
Ionic: Manual cleanup here.

nxcomp/Loop.cpp
```diff
@@ -9225,36 +8536,30 @@ int ParseCommandLineOptions(int argc, const char **argv)
...
-  #ifdef TEST
 
-  *logofs << "Loop: Argc is " << argc << ".\n" << logofs_flush;
+  nxinfo << "Loop: Argc is " << argc << ".\n" << std::flush;
 
   for (int argi = 0; argi < argc; argi++)
-  {
-    *logofs << "Loop: Argv[" << argi << "] is " << argv[argi]
-            << ".\n" << logofs_flush;
+  {
```
Ionic: Stray whitespace after curly brace.

nxcomp/Loop.cpp
```diff
@@ -13954,19 +13021,15 @@ void PrintProcessInfo()
...
   {
-    cerr << "Info" << ": Using errors file '" << errorsFileName << "'.\n";
-  }
+    cerr << "Info" << ": Using errors file '" << errorsFileName << "'.\n";  }
 
   if (*statsFileName != '\0')
   {
-    cerr << "Info" << ": Using stats file '" << statsFileName << "'.\n";
-  }
+    cerr << "Info" << ": Using stats file '" << statsFileName << "'.\n";  }
```
Ionic: The last two changes only remove correct newlines.

Because it's outputting directly to `cerr`, we should keep these lines in mind though. More on that later.

nxcomp/Loop.cpp
```diff
@@ -14394,25 +13443,23 @@ int ParseArg(const char *type, const char *name, const char *value)
...
+         << "' from string '" << string << "' with base set to ";
 
   switch (tolower(*id))
   {
     case 'k':
     case 'm':
     case 'g':
-    {
-      *logofs << (char) toupper(*id);
+    {
```
Ionic: Stray whitespace after curly brace.

Ionic:
While looking through your changes, I noticed that some logging output was duplicated by NoMachine - once outputted via their `logofs` infrastructure (which will be replaced by your new work) and once via `cerr`.

Do we want to keep this duplication or remove it? Also, what about logging output that only ever goes to `cerr` but was never written via `logofs`? Do we want to keep it as-is, or make it use your new logging capabilities?

vatral:
@Ionic I've looked into the cerr/logofs bits. Many can be removed, and I've written some code to try to detect the duplicates, but in the end didn't enable it for this iteration because of the amount of changes that need careful checking.

My plan was to follow up with removing the duplication.

Ionic (approved):

Ionic:
Rebased against current `improved-logging-loop` branch.

I also fixed all whitespace errors that I could find.

sunweaver (approved):
LGTM

Issue #287, opened 2016-11-15
RHEL7: still linking libX11


uli42:
On RHEL7 the nxagent wrapper script calls nxagent like this:
```
NX_LIBS=/usr/lib64/nx

# make sure nxagent starts properly with pam_tmpdir.so being in use
NX_TEMP=${NX_TEMP:-/tmp}
export NX_TEMP
export LD_LIBRARY_PATH=/usr/lib64/nx-X11/

exec $NX_LIBS/bin/${NXAPP:-"nxagent"} "$@"
```

 Unfortunately this does not work. I can still see libX11 from the system in the linker output:
```
# LD_LIBRARY_PATH=/usr/lib64/nx-X11 ldd -v /usr/lib64/nx/bin/nxagent | less
        linux-vdso.so.1 =>  (0x00007ffe76bd8000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f9261978000)
        libXpm.so.4 => /lib64/libXpm.so.4 (0x00007f9261765000)
        libxml2.so.2 => /lib64/libxml2.so.2 (0x00007f92613fb000)
        libpixman-1.so.0 => /lib64/libpixman-1.so.0 (0x00007f926114e000)
        libXcomp.so.3 => /lib64/libXcomp.so.3 (0x00007f9260e60000)
        libXcompext.so.3 => /lib64/libXcompext.so.3 (0x00007f9260c4e000)
        libXcompshad.so.3 => /lib64/libXcompshad.so.3 (0x00007f9260a42000)
        libXrender.so.1 => /lib64/libXrender.so.1 (0x00007f9260837000)
        libXfixes.so.3 => /lib64/libXfixes.so.3 (0x00007f9260631000)
        libXfont.so.1 => /lib64/libXfont.so.1 (0x00007f92603fd000)
        libXcomposite.so.1 => /lib64/libXcomposite.so.1 (0x00007f92601f9000)
        libXinerama.so.1 => /lib64/libXinerama.so.1 (0x00007f925fff6000)
        libXdmcp.so.6 => /lib64/libXdmcp.so.6 (0x00007f925fdf0000)
-->        libNX_X11.so.6 => /lib64/libNX_X11.so.6 (0x00007f925faac000)
        libXext.so.6 => /lib64/libXext.so.6 (0x00007f925f89a000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f925f4d8000)
[...]

PR #307, opened 2017-01-11
Install missing libNX_X11.so.6 symlink - fixes gh-287

opoplawski:


uli42:
Looks good to me. (Adding linkable reference to #287 hereby)

PR #313, opened 2017-02-04
use rpath for nxagent (instead of LD_LIBRARY_PATH hack)

sunweaver:
Along with @uli42's suggestion, here comes a PR that makes the LD_LIBRARY_PATH in the /usr/bin/nxagent wrapper script obsolete.

@Ionic: can you please review this PR as paid work. Thanks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -186,14 +186,14 @@ uninstall-lite:
 	$(RM_DIR) $(DESTDIR)$(NXLIBDIR)/bin/
 	$(RM_FILE) $(DESTDIR)$(PREFIX)/share/man/man1/*.1
 	$(RM_FILE) $(DESTDIR)$(PREFIX)/share/nx/VERSION.nxproxy
-	$(RM_DIR) $(DESTDIR)$(NXLIBDIR)/share/nx/
+	$(RM_DIR) $(DESTDIR)$(PREFIX)/share/nx/
 
 uninstall-full:
 	for f in nxagent; do \
 	    $(RM_FILE) $(DESTDIR)$(BINDIR)/$$f; done
 
 	$(RM_FILE) $(DESTDIR)$(PREFIX)/share/nx/VERSION.nxagent
-	$(RM_DIR) $(DESTDIR)$(NXLIBDIR)/share/nx/
+	$(RM_DIR) $(DESTDIR)$(PREFIX)/share/nx/
 
 	if test -d nx-X11; then \
 	    if test -f nxcompshad/Makefile; then ${MAKE} -C nxcompshad $@; fi; \
--- a/Makefile
+++ b/Makefile
@@ -9,15 +9,15 @@ COPY_DEREFERENCED=cp -RH
 RM_FILE=rm -f
 RM_DIR=rmdir -p --ignore-fail-on-non-empty
 
-ETCDIR_NX   ?= /etc/nxagent
-PREFIX      ?= /usr/local
-BINDIR      ?= $(PREFIX)/bin
-LIBDIR      ?= $(PREFIX)/lib
-SHLIBDIR    ?= $(LIBDIR)
-NXLIBDIR    ?= $(SHLIBDIR)/nx
-USRLIBDIR   ?= $(NXLIBDIR)/X11
-INCLUDEDIR  ?= $(PREFIX)/include
-CONFIGURE   ?= ./configure
+ETCDIR_NX       ?= /etc/nxagent
+PREFIX          ?= /usr/local
+BINDIR          ?= $(PREFIX)/bin
+LIBDIR          ?= $(PREFIX)/lib
+SHLIBDIR        ?= $(LIBDIR)
+NXLIBDIR        ?= $(SHLIBDIR)/nx
+USRLIBDIR       ?= $(NXLIBDIR)/X11
+INCLUDEDIR      ?= $(PREFIX)/include
+CONFIGURE       ?= ./configure
 
 NX_VERSION_MAJOR=$(shell ./version.sh 1)
 NX_VERSION_MINOR=$(shell ./version.sh 2)
--- a/Makefile
+++ b/Makefile
@@ -13,9 +13,10 @@ ETCDIR_NX   ?= /etc/nxagent
 PREFIX      ?= /usr/local
 BINDIR      ?= $(PREFIX)/bin
 LIBDIR      ?= $(PREFIX)/lib
-USRLIBDIR   ?= $(LIBDIR)
+SHLIBDIR    ?= $(LIBDIR)
+NXLIBDIR    ?= $(SHLIBDIR)/nx
+USRLIBDIR   ?= $(NXLIBDIR)/X11
 INCLUDEDIR  ?= $(PREFIX)/include
-NXLIBDIR    ?= $(LIBDIR)/nx
 CONFIGURE   ?= ./configure
 
 NX_VERSION_MAJOR=$(shell ./version.sh 1)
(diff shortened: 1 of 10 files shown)


sunweaver:
@uli42: I'd be happy if you could also take a look. However, there have been various (boring, but many) Makefile changes. Unfortunately, @Ionic is our most meticulous reviewer, so I assigned him for doing the actual code review.

Makefile
```diff
@@ -143,13 +144,13 @@ install-full:
...
-	$(COPY_SYMLINK) nx-X11/.build-exports/lib/libNX_X11.so.6 $(DESTDIR)$(USRLIBDIR)/
-	$(COPY_DEREFERENCED) nx-X11/.build-exports/lib/libNX_X11.so.6.2 $(DESTDIR)$(USRLIBDIR)/
-	$(INSTALL_DIR) $(DESTDIR)$(USRLIBDIR)/nx-X11
-	$(INSTALL_SYMLINK) ../libNX_X11.so.6 $(DESTDIR)$(USRLIBDIR)/nx-X11/libX11.so.6
-	$(INSTALL_SYMLINK) ../libNX_X11.so.6.2 $(DESTDIR)$(USRLIBDIR)/nx-X11/libX11.so.6.2
+	$(INSTALL_DIR) $(DESTDIR)$(SHLIBDIR)
+	$(COPY_SYMLINK) nx-X11/.build-exports/lib/libNX_X11.so $(DESTDIR)$(SHLIBDIR)/
+	$(COPY_SYMLINK) nx-X11/.build-exports/lib/libNX_X11.so.6 $(DESTDIR)$(SHLIBDIR)/
+	$(COPY_DEREFERENCED) nx-X11/.build-exports/lib/libNX_X11.so.6.2 $(DESTDIR)$(SHLIBDIR)/
+	$(INSTALL_DIR) $(DESTDIR)$(SHLIBDIR)/nx-X11
```
Ionic: Make that `$(INSTALL_DIR) $(DESTDIR)$(USRLIBDIR)` if we're changing it anyway.

Ionic: Likewise in the two lines below.

Makefile
```diff
@@ -181,23 +182,23 @@ uninstall-lite:
...
 	# FIXME: don't use uninstall rule in nxproxy/Makefile.in, let's do
         # it on our own for now...
-	$(RM_FILE) $(DESTDIR)$(NXLIBDIR)/bin/nxproxy
-	$(RM_DIR) $(DESTDIR)$(NXLIBDIR)/bin/
+	$(RM_FILE) $(DESTDIR)$(NXDIR)/bin/nxproxy
+	$(RM_DIR) $(DESTDIR)$(NXDIR)/bin/
 	$(RM_FILE) $(DESTDIR)$(PREFIX)/share/man/man1/*.1
 	$(RM_FILE) $(DESTDIR)$(PREFIX)/share/nx/VERSION.nxproxy
-	$(RM_DIR) $(DESTDIR)$(NXLIBDIR)/share/nx/
+	$(RM_DIR) $(DESTDIR)$(NXDIR)/share/nx/
```
Ionic: I'm only noticing this now, but shouldn't that be `$(RM_DIR) $(DESTDIR)$(PREFIX)/share/nx/` here?

Makefile
```diff
@@ -181,23 +182,23 @@ uninstall-lite:
...
 
 uninstall-full:
 	for f in nxagent; do \
 	    $(RM_FILE) $(DESTDIR)$(BINDIR)/$$f; done
 
 	$(RM_FILE) $(DESTDIR)$(PREFIX)/share/nx/VERSION.nxagent
-	$(RM_DIR) $(DESTDIR)$(NXLIBDIR)/share/nx/
+	$(RM_DIR) $(DESTDIR)$(NXDIR)/share/nx/
```
Ionic: Also `PREFIX` here?

Makefile
```diff
@@ -71,7 +72,7 @@ build-full:
 
 	cd nxcompshad && autoconf && (${CONFIGURE}) && ${MAKE}
 
-	cd nx-X11 && ${MAKE} World
+	cd nx-X11 && ${MAKE} World USRLIBDIR=$(USRLIBDIR) SHLIBDIR=$(SHLIBDIR)
```
Ionic, line 75: Pointing over `USRLIBDIR` to the `nx-X11` path might help with setting the RPATH correctly (well, to some degree, but more on that later...), but will also have side-effects like installing lint libraries there or generally any library that is installed using `InstallLibrary()`.

Are you sure this is well-tested? Do you have the same files in binary packages before and after that PR?

debian/patches/016_nx-X11_install-location.debian.patch
```diff
@@ -6,16 +6,16 @@ Author: Mike Gabriel <mike.gabriel@das-netzwerkteam.de>
 Last-Update: 2012-12-31
```
Ionic: Don't forget to update the comment here, which is also... kind of wrong. We manually pull files out of the staging dir and library and binary directories differ. Not to mention that life is even more complicated due to multilib...

nx-X11/config/cf/lnxLib.rules
```diff
@@ -9,7 +9,7 @@
 #endif
 
 #ifndef UseRpath
-#define UseRpath NO
+#define UseRpath YES
```
Ionic, line 12: And that... could fail spectacularly.

It looks as if the RPATH is set conditionally based upon whether `UseInstalled` is defined or not. If it is, `USRLIBDIRPATH` is being used (which is kind-of what we want), otherwise `BUILDLIBDIR`.

I don't see `UseInstalled` being defined anywhere, so I guess we'll be ending up with `BUILDLIBDIR` being used, which is set to `$(TOP)/exports/lib`.

So do we **actually** get stuff RPATH'd to the correct location, or will `nxagent` just fail to execute? Did you test that?

That being said, even IF `USRLIBDIRPATH` is being used, I suspect that it could potentially be set to `/usr/lib/nx-X11` as defined in `nx-X11/config/cf/Imake.tmpl`, overriding the overridden value. If not (for instance because values passed directly to make normally override all variable definitions inside of a Makefile - but I don't know how imake behaves in this respect), we might be in the clear.

Ionic (changes requested):
Marking that as "Request changes" for now. Some comments really request a change, others are marking potential pitfalls and requesting discussion...

Makefile
sunweaver: We only have libNX_X11 left. And libNX_X11 is intalled into SHLIBDIR by our main Makefile. We don't use InstallLibrary() for nx-libs.

This will go away with the autoconf/automake migration, so let's bother about this here. If needed, I could remove the InstallLibrary() code from the imake stuff.

Makefile
sunweaver: Done.

Makefile
sunweaver: > I'm only noticing this now, but shouldn't that be $(RM_DIR) $(DESTDIR)$(PREFIX)/share/nx/ here?

Oh right. Good catch. Injecting an extra-commit for this.

Makefile
sunweaver: Yep, fixed.

nx-X11/config/cf/lnxLib.rules
sunweaver: I have been wondering for a while, what you mean by the above.

Indeed, it is so, that nxagent is linked with -Wl,-rpath-link,$(BUILDLIBDIR) (because UseInstalled is not defined at build time for our nx-libs). This is for the linker of nxagent, so that it can find all necessary libs for linking nxagent at build time.

And(!), nxagent is liniked with  -Wl,-rpath,$(USRLIBDIRPATH), this is also what we want. This path gets hard-coded into the nxagent executable, so that at runtime the library can be found.

The latter RPATH setting is not conditional to UseInstalled or any other imake variable.

sunweaver:
@Ionic: ok. Next round.

Changes:

  * USRLIBDIR -> /usr/<lib>/nx/X11 (not nx-X11)
  * USRLIBDIRPATH overridable in main Makefile
  * $(PREFIX)/share/nx fixes in main Makefile
  * update of 016 debian/patch and its comment header
  * reviewed rpath-link vs. rpath build options -> no change needed IMHO

Makefile
Ionic: Right, we never call any `install` rule but pull out needed stuff manually from the build dir.

nx-X11/config/cf/lnxLib.rules
Ionic: Yep, I failed to see that `ExtraLoadFlags` does *always* include `RpathLoadFlags` unconditionally, which may or may not be set based upon `UseRpath`, so that's actually fine.

Must have skipped `RpathLoadFlags` mentally and only saw the `-Wl,-rpath-link,...` part.

Makefile
```diff
@@ -9,14 +9,16 @@ COPY_DEREFERENCED=cp -RH
...
-NXLIBDIR    ?= $(LIBDIR)/nx
-CONFIGURE   ?= ./configure
+ETCDIR_NX       ?= /etc/nxagent
+PREFIX          ?= /usr/local
+BINDIR          ?= $(PREFIX)/bin
+LIBDIR          ?= $(PREFIX)/lib
+SHLIBDIR        ?= $(LIBDIR)
+NXLIBDIR        ?= $(SHLIBDIR)/nx
+USRLIBDIR       ?= $(NXLIBDIR)/X11
+USRLIBDIRPATH   ?= $(USRLIBDIR)
```
Ionic: We probably don't need to override `USRLIBDIRPATH` if `USRLIBDIR` gets properly overridden. Have you tried building this branch? Is the RPATH set correctly? (Also respecting multilib.)

If yes, and leaving out `USRLIBDIRPATH` in this Makefile also builds and leads to the correct result, we can drop it here for simplicity.

sunweaver: USRLIBDIRPATH override has been removed now. The update branch has been rebased against latest 3.6.x. Please merge then.

Ionic (approved):

Issue #51, opened 2015-06-19
empty nx/nx options disables nxcomp backend


sunweaver:
Using an empty option string in the display variable, disables NX compression...

Not ok: nx/nx:50
Ok: nx/nx,foo=bar:50
Ok: nx/nx,options=<file>:50
Not ok: nx/nx,options=/dev/null:50

Spotted by https://github.com/salva and reported during X2Go Gathering 2015.

sunweaver:
> Not ok: nx/nx,options=/dev/null:50

Any path starting with /dev/, /proc/, or /sys/ should be ignored. Providing a patch for that.

sunweaver:
Fix provided as PR #314.

PR #314, opened 2017-02-05
Allow empty nx/nx options string

sunweaver:
This PR fixes issue #51, so please close that in your merge commit msg.

### DIFF
--- a/nxcomp/Loop.cpp
+++ b/nxcomp/Loop.cpp
@@ -9138,7 +9138,7 @@ int ParseEnvironmentOptions(const char *env, int force)
           << env << "'.\n" << logofs_flush;
   #endif
 
-  if (*fileOptions != '\0')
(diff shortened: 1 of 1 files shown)


sunweaver:
@vatral: @salva: I took some time to fix a bug that @salva reported in 2015 at the X2Go Gathering. Please review and approve, if appropriate. Thanks.

Ionic (approved):
Looks good. I guess we don't *need* to exclude `/sys/`, `/proc/` and `/dev/`, but it looks like the current code just fails if something like `options=/dev/null` is passed? So just skipping over that might be better. Let's just hope no one sane comes along and actually tries to use something in `/dev`, `/proc` or `/sys` as a real options file.

sunweaver:
Rebased against 3.6.x. Happy merging... Thanks for the review.

PR #316, opened 2017-02-05
Support built-in fonts

sunweaver:
This PR, once and for all, fixes the "Could not open default font 'fixed'" issue in nxagent.

Important for distribution maintainers:

  symlink /usr/share/nx/fonts -> <your-distro's-X11-fonts-dir>
  drop hard dependency on xfonts-base, xorg-x11-fonts-misc (not required anymore as a must)

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Font.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Font.c
@@ -77,26 +77,22 @@ is" without express or implied warranty.
 #define NXAGENT_DEFAULT_FONT_PATH  \
 "/usr/share/nx/fonts/misc/,/usr/share/nx/fonts/Speedo/,\
 /usr/share/nx/fonts/Type1/,/usr/share/nx/fonts/75dpi/,\
-/usr/share/nx/fonts/100dpi/,/usr/share/nx/fonts/TTF/,\
-/usr/NX/share/fonts/base"
+/usr/share/nx/fonts/100dpi/,/usr/share/nx/fonts/TTF/"
 
 #define NXAGENT_ALTERNATE_FONT_PATH  \
 "/usr/share/X11/fonts/misc/,/usr/share/X11/fonts/Speedo/,\
 /usr/share/X11/fonts/Type1/,/usr/share/X11/fonts/75dpi/,\
-/usr/share/X11/fonts/100dpi/,/usr/share/X11/fonts/TTF/,\
-/usr/NX/share/fonts/base"
+/usr/share/X11/fonts/100dpi/,/usr/share/X11/fonts/TTF/"
 
 #define NXAGENT_ALTERNATE_FONT_PATH_2  \
 "/usr/share/fonts/X11/misc/,/usr/share/fonts/X11/Speedo/,\
 /usr/share/fonts/X11/Type1/,/usr/share/fonts/X11/75dpi/,\
-/usr/share/fonts/X11/100dpi/,/usr/share/fonts/X11/TTF/,\
-/usr/NX/share/fonts/base"
+/usr/share/fonts/X11/100dpi/,/usr/share/fonts/X11/TTF/"
 
 #define NXAGENT_ALTERNATE_FONT_PATH_3  \
 "/usr/X11R6/lib/X11/fonts/misc/,/usr/X11R6/lib/X11/fonts/Speedo/,\
 /usr/X11R6/lib/X11/fonts/Type1/,/usr/X11R6/lib/X11/fonts/75dpi/,\
-/usr/X11R6/lib/X11/fonts/100dpi/,/usr/X11R6/lib/X11/fonts/TTF/,\
-/usr/NX/share/fonts/base"
+/usr/X11R6/lib/X11/fonts/100dpi/,/usr/X11R6/lib/X11/fonts/TTF/"
 
 #undef NXAGENT_FONTCACHE_DEBUG
 #undef NXAGENT_RECONNECT_FONT_DEBUG
--- a/nx-X11/programs/Xserver/hw/nxagent/Font.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Font.c
@@ -73,7 +73,6 @@ is" without express or implied warranty.
 #define NXAGENT_ALTERNATE_FONT_DIR    "/usr/share/X11/fonts"
 #define NXAGENT_ALTERNATE_FONT_DIR_2  "/usr/share/fonts/X11"
 #define NXAGENT_ALTERNATE_FONT_DIR_3  "/usr/share/fonts"
-#define NXAGENT_ALTERNATE_FONT_DIR_4  "/usr/NX/share/fonts"
 
 #define NXAGENT_DEFAULT_FONT_PATH  \
 "/usr/X11R6/lib/X11/fonts/misc/,/usr/X11R6/lib/X11/fonts/Speedo/,\
@@ -98,9 +97,6 @@ is" without express or implied warranty.
 /usr/share/fonts/Type1/,/usr/share/fonts/75dpi/,\
 /usr/share/fonts/100dpi/,/usr/share/fonts/TTF/,\
 /usr/NX/share/fonts/base"
-
-#define NXAGENT_ALTERNATE_FONT_PATH_4  \
-"/usr/NX/share/fonts/base"
 
 #undef NXAGENT_FONTCACHE_DEBUG
 #undef NXAGENT_RECONNECT_FONT_DEBUG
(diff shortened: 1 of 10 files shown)


sunweaver:
@opoplawski: please take notice of my above comment in PR #316. Thanks.

sunweaver:
@opoplawski: I'd highly appreciate, if you could adapt the nx-libs.spec regarding the above described symlinking of the fonts paths. Thanks!

sunweaver:
@Ionic: and another review for you...

nx-X11/programs/Xserver/hw/nxagent/Font.c
```diff
@@ -1575,50 +1567,6 @@ void nxagentVerifyDefaultFontPath(void)
...
-    }
-    else
-    {
-      fontPath = realloc(fontPath, strlen(fontPath) + strlen(NXAGENT_ALTERNATE_FONT_PATH_4) + 1);
-    }
-
-    strcat(fontPath, NXAGENT_ALTERNATE_FONT_PATH_4);
-  }
-
-  if (*fontPath == '\0')
```
Ionic: Why are you removing this? A mistake?

nx-X11/programs/Xserver/hw/nxagent/NXdixfonts.c
```diff
@@ -936,19 +939,43 @@ SetDefaultFontPath(char *path)
...
+    else {
+	temp_path = strdup(path);
+    }
+    if (!temp_path)
+	return BadAlloc;
+
     /* get enough for string, plus values -- use up commas */
 #ifdef NX_TRANS_SOCKET
-    len = strlen(_NXGetFontPath(path)) + 1;
+    len = strlen(_NXGetFontPath(temp_path)) + 1;
```
Ionic: This will override `temp_path` if the `NX_FONT` env variable is set. Maybe we need logic like that in there (`_NXGetFontPath()`) as well to always have the `built-ins` string appended if needed?

nx-X11/programs/Xserver/hw/nxagent/NXdixfonts.c
```diff
@@ -936,19 +939,43 @@ SetDefaultFontPath(char *path)
...
 #endif
     nump = cp = newpath = (unsigned char *) ALLOCATE_LOCAL(len);
-    if (!newpath)
+    if (!newpath) {
+	DEALLOCATE_LOCAL(temp_path);
 	return BadAlloc;
+    }
 #ifdef NX_TRANS_SOCKET
-    pp = (unsigned char *) _NXGetFontPath(path);
+    pp = (unsigned char *) _NXGetFontPath(temp_path);
```
Ionic: Same here.

Ionic (commented):
What do you mean by "drop hard dependency on xfonts-base, xorg-x11-fonts-misc"? Are you referring to "built-ins"? Because otherwise, they are still needed very much... unless we were to let nxagent install its own sets of fonts, but that's stupid.

opoplawski:
If you want to symlink the fonts dir, then you would have something like this:
```
%package -n nxagent
# For /usr/share/X11/fonts
Requires: xorg-x11-font-utils

%install
ln -s ../X11/fonts %{buildroot}%{_datadir}/nx/fonts

%files -n nxagent
%{_datadir}/nx/fonts
```

sunweaver:
Hi,

On  Mo 06 Feb 2017 22:34:47 CET, Orion Poplawski wrote:

> If you want to symlink the fonts dir, then you would have something  
> like this:
> ```
> %package -n nxagent
> # For /usr/share/X11/fonts
> Requires: xorg-x11-font-utils
>
> %install
> ln -s ../X11/fonts %{buildroot}%{_datadir}/nx/fonts
>
> %files -n nxagent
> %{_datadir}/nx/fonts
> ```

@opoplawski:

Sort of like this?
https://github.com/ArcticaProject/nx-libs/pull/316/commits/b50d7c650a080461f7cd346033a1e053b92a294e

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

Ionic:
Also add a `BuildRequires` on `xorg-x11-font-utils` (and maybe the SUSE-equivalent, which we currently don't seem to have), as otherwise the build system/lint may issue warnings regarding links to non-existent directories. At least OBS probably checks for that as a QA measure.

nx-X11/programs/Xserver/hw/nxagent/NXdixfonts.c
sunweaver: > This will override temp_path if the NX_FONT env variable is set. Maybe we need logic like that 
> in there (_NXGetFontPath()) as well to always have the built-ins string appended if needed?

I think I found a better solution for this, I put...

```
start = _NXGetFontPath(path)
```

at the beginning of the function (instead of start = path).

sunweaver:
> Also add a BuildRequires on xorg-x11-font-utils (and maybe the SUSE-equivalent, which we currently 
> don't seem to have)...

For Fedora/RHEL: done. For SUSE, there is no such package (I saw SUSE 13.1 putting packages into /usr/share/fonts directly.

We really need an easy way to test-build against openSUSE, Fedora and alike... Gnaaa...

nx-X11/programs/Xserver/hw/nxagent/NXdixfonts.c
sunweaver: dito, see above

sunweaver:
@Ionic: second iteration for #316, please.

nx-X11/programs/Xserver/hw/nxagent/NXdixfonts.c
```diff
@@ -936,20 +939,41 @@ SetDefaultFontPath(char *path)
                 size = 0,
                 bad;
 
-    /* get enough for string, plus values -- use up commas */
 #ifdef NX_TRANS_SOCKET
-    len = strlen(_NXGetFontPath(path)) + 1;
+    start = (char *) _NXGetFontPath(path);
```
Ionic: Good idea, but wrong implementation:
```C
#ifdef NX_TRANS_SOCKET
    path = (char*)_NXGetFontPath(path);
#endif /* NX_TRANS_SOCKET */

    start = path;
```

Otherwise we'll be checking the NX font path against the originally passed value (`path`) and also write it to `temp_path` etc.

nx-X11/programs/Xserver/hw/nxagent/Font.c
Ionic: Ping?

It's probably a good idea to keep the last part.

nx-libs.spec
```diff
@@ -50,7 +50,7 @@ BuildRequires:  pixman-devel >= 0.13.2
 BuildRequires:  xorg-x11-libX11-devel
 BuildRequires:  xorg-x11-libXext-devel
 BuildRequires:  xorg-x11-libXpm-devel
-BuildRequires:  xorg-x11-libXfont-devel
+BuildRequires:  xorg-x11-libXfont-devel >= 1.4.2
 BuildRequires:  xorg-x11-libXdmcp-devel
 BuildRequires:  xorg-x11-libXdamage-devel
 BuildRequires:  xorg-x11-libXcomposite-devel
```
Ionic, line 60: In the general SUSE section, i.e., either after the next `%endif` or before `%if 0%{?suse_version} >= 1130` above, add `BuildRequires: filesystem`. (This is what provides `/usr/share/fonts`.)

sunweaver:
Hi @Ionic,

> What do you mean by "drop hard dependency on xfonts-base, xorg-x11-fonts-misc"? Are you
> referring to "built-ins"? Because otherwise, they are still needed very much... unless we were to
> let nxagent install its own sets of fonts, but that's stupid.

as it seems the fixed fonts are compiled into libXfont1 and libXfont2 nowadays. So, with this PR merged, nxagent can run without problems on systems where xfonts-base is not installed. Esp. where fonts under /usr/share/fonts/X11/misc are missing.

(Was that your question?)
Mike

nx-X11/programs/Xserver/hw/nxagent/Font.c
sunweaver: Oh, yes. No clue why this slipped out. Good catch!

nx-X11/programs/Xserver/hw/nxagent/NXdixfonts.c
sunweaver: > Good idea, but wrong implementation: ...

Indeed. Fixed now.

nx-libs.spec
sunweaver: Ack. Added as an extra commit to nx-libs.spec.

sunweaver:
@Ionic: this PR (#316) is ready for another iteration...

nx-X11/programs/Xserver/hw/nxagent/NXdixfonts.c
```diff
@@ -936,20 +939,41 @@ SetDefaultFontPath(char *path)
...
+    }
+    if (!temp_path)
+	return BadAlloc;
+
+    /* get enough for string, plus values -- use up commas */
+    len = strlen(temp_path) + 1;
     nump = cp = newpath = (unsigned char *) ALLOCATE_LOCAL(len);
-    if (!newpath)
+    if (!newpath) {
+	DEALLOCATE_LOCAL(temp_path);
```
uli42: I cannot see the corresponding ALLOCATE_LOCAL!?

Ionic: strdup

PR #317, opened 2017-02-05
nxproxy: Support communication over stdin. Allows nxcl to use standard ssh instead of nxssh.

sunweaver:
Patch from Gentoo. Ported 1:1.

### DIFF
--- a/nxproxy/Main.c
+++ b/nxproxy/Main.c
@@ -19,6 +19,8 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <errno.h>
+#include <limits.h>
 
 #include "NX.h"
 
@@ -26,6 +28,8 @@
 #define WARNING
 #undef  TEST
 #undef  DEBUG
+
+extern const char *__progname;
 
 /*
  * Entry point when running nxproxy stand-alone.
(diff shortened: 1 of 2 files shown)


sunweaver:
@voyageur: I am about to port one of your Gentoo patch to nx-libs / nxproxy. Could you please shortly visit us here and confirm that the patch is still valid and needed? Thanks!

voyageur:
Hmm, this Arctica project looks interesting!
This patch origin is here:
https://mail.kde.org/pipermail/freenx-knx/2008-March/006794.html
https://mail.kde.org/pipermail/freenx-knx/2008-April/006975.html

As far as I remember, only nxcl/qtnx needed it to work properly, other nx servers at the time did fine without the patch,  as still does x2go upstream I think). So it should not be useful to keep these days (I'll make a not for the next Gentoo package bump to test without it)

sunweaver:
Hi Bernhard,

On  Mo 06 Feb 2017 22:58:00 CET, Bernard Cafarelli wrote:

> Hmm, this Arctica project looks interesting!
> This patch origin is here:
> https://mail.kde.org/pipermail/freenx-knx/2008-March/006794.html
> https://mail.kde.org/pipermail/freenx-knx/2008-April/006975.html
>
> As far as I remember, only nxcl/qtnx needed it to work properly,  
> other nx servers at the time did fine without the patch,  as still  
> does x2go upstream I think). So it should not be useful to keep  
> these days (I'll make a not for the next Gentoo package bump to test  
> without it)

Thanks for your feedback. There is another use case of NX that comes  
from devs who like stdin communications. Let's see what they think!

@vatral: @salva: any use case for nxproxy -stdin support from your  
side? Otherwise, we will drop the Gentoo patch and close this issue.

Thanks+Greets,
Mike

-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

salva:
From the [QVD](https://github.com/theqvd/theqvd/) side, this is something we would really like to see incorporated into `nxproxy`, as we try to avoid opening ports on localhost as much as possible (at least when the OS allows it).

@sunweaver, feel free to assign the issue to @vatral if you thing it needs further work or just for reviewing/testing.

sunweaver:
@vatral: Please review this PR (#317) and if possible, try to provide a test script later. Thanks.

sunweaver:
@vatral: @salva: If you want to this PR #317 merged into nx-libs 3.6.x, I'd appreciate a code review. Ideally plus some tests. Thanks!

sunweaver:
@vatral: @salva: ping on #317. Code review is needed if you want to see this PR merged. Thanks!

vatral:
Hello,

At the moment I'm busy with an urgent matter, but I will take a look at that before the weekend.

nxproxy/Main.c
```diff
@@ -49,15 +51,33 @@ int main(int argc, const char **argv)
...
   }
 
-  /*
-   * This should not return...
-   */
+  if ((nx_commfd_str = getenv("NX_COMMFD")) != NULL)
+  {
+    int nx_commfd = atoi(nx_commfd_str);
+
+    if (nx_commfd)
```
vatral: STDIN is filehandle 0, and will evaluate as false in this comparison.

vatral (changes requested):
The stdin issue needs fixing.

I'm looking at how to do a test script for this.

sunweaver:
@vatral: next iteration. I actually did some testing of the patch myself now and had to rearrange some stuff in nxproxy's Main.c compared to the previous approach. Please look again. Thanks!

vatral (approved):

uli42:
I am wondering how to use this? Is there any example how to establish an connection to the nxagent with this method?

PR #318, opened 2017-02-05
Really enable MITSHM code sections in fboverlay c

sunweaver:


Ionic (commented):
Looks good in general... but shouldn't we do the same to `nx-X11/programs/Xserver/mi/Imakefile` for `miscrinit.c` if we enable it for `fboverlay.c`?

sunweaver:
@Ionic: No. Actually, good catch. But -DMITSHM (aka ShmDefines) is covered by ExtensionDefines from config/cf/X11.tmpl. This should be enough.

Update branch is rebased now and ready for merging.

Ionic (approved):

PR #319, opened 2017-02-06
Xserver: hard-code rgb colors

sunweaver:
Immitate what X.org has done a while ago. Drop rgb in text fromat or DBM format and simply hard-code it in Xserver/os/oscolor.c.


Ionic:
Is this one for me?

sunweaver:
Yes, go ahead.

On Sat Feb 18 10:53:32 2017 GMT+0100, Mihai Moldovan wrote:
> Is this one for me? 
> —
>

-- 

nx-X11/programs/Xserver/os/Imakefile
```diff
@@ -146,7 +142,7 @@ BOOTSTRAPCFLAGS =
 		  -I$(SERVERSRC)/Xext -I$(SERVERSRC)/render \
```
Ionic, line 142: `$(RGB_DEFINES)` three lines above this one.

nx-libs.spec
```diff
@@ -510,7 +510,6 @@ rm -r %{buildroot}%{_includedir}/nx-X11/Xtrans
 %dir %{_sysconfdir}/nxagent
 %config(noreplace) %{_sysconfdir}/nxagent/keystrokes.cfg
 %config(noreplace) %{_sysconfdir}/nxagent/nxagent.keyboard
-%config(noreplace) %{_sysconfdir}/nxagent/rgb
```
Ionic, line 513: @opoplawski is that enough to drop config files? I searched around, but it seems that RPM has no special concept of config file deprecation. Will it do the right thing automatically, or just leave the file around, even if unmodified?

nx-X11/programs/Xserver/os/oscolor.c
```diff
@@ -45,438 +44,1619 @@ SOFTWARE.
...
-#include <ndbm.h>
-#else
-#ifdef SVR4
-#include <rpcsvc/dbm.h>
-#else
-#include <dbm.h>
-#endif
-#endif
-#include "rgb.h"
+#include <X11/keysym.h>
```
Ionic: Why do we need that? Looks misplaced, keysyms shouldn't be used in `oscolor.c` at all.

Ionic (changes requested):
Everything else looks okay. The warning... is probably just gcc being too helpful. One could probably split the large arrays up and concat them together at run time, but if compilers are handling it just fine it's probably better to not add such a quirk. That's my interpretation of why upstream didn't do it in the first place.

nx-libs.spec
opoplawski: It will do the right thing - remove the file if unmodified - keep a copy as rgb.rpmsave if it was.

Ionic: Okay, thanks!

nx-X11/programs/Xserver/os/oscolor.c
sunweaver: This is true. I "blindly" copied over X.org's oscolor.c. This really feels like an unneeded include. Dropping it (and diverting from X.org).

sunweaver:
@Ionic: requested changes worked-in and rebased against 3.6.x branch. Please take another look and merge.

nx-X11/programs/Xserver/os/Imakefile
sunweaver: >  $(RGB_DEFINES) three lines above this one.

Yep. Removed now.

Ionic (approved):

Issue #325, opened 2017-02-14
ReduceCompositeOp (picture.c:1674): Conditional jump or move depends on uninitialised value(s)


sunweaver:
I regularly see these issues when running nxagent locally with valgrind:

```
==10999== Conditional jump or move depends on uninitialised value(s)
==10999==    at 0x170278: ReduceCompositeOp (picture.c:1674)
==10999==    by 0x170278: CompositePicture (picture.c:1777)
==10999==    by 0x1774E6: miGlyphs (NXmiglyph.c:228)
==10999==    by 0x17DF86: damageGlyphs (damage.c:618)
==10999==    by 0x16CE14: ProcRenderCompositeGlyphs (NXrender.c:1129)
==10999==    by 0x1673F7: ProcRenderDispatch (NXrender.c:1687)
==10999==    by 0x166E7E: Dispatch (NXdispatch.c:491)
==10999==    by 0x14A762: main (main.c:433)
```

sunweaver:
running valgrind with --track-origins=yes reveals this:

```
==13181==  Uninitialised value was created by a heap allocation
==13181==    at 0x4C2BBAF: malloc (vg_replace_malloc.c:299)
==13181==    by 0x16F197: createSourcePicture (NXpicture.c:329)
==13181==    by 0x170AA7: CreateSolidPicture (NXpicture.c:285)
==13181==    by 0x16788B: ProcRenderCreateSolidFill (NXrender.c:1530)
==13181==    by 0x1673F7: ProcRenderDispatch (NXrender.c:1687)
==13181==    by 0x166E7E: Dispatch (NXdispatch.c:491)
==13181==    by 0x14A762: main (main.c:433)
```

PR #327, opened 2017-02-14
render: picture.c avoid use of uninitialised values

sunweaver:
This is WIP, PR filed for documentation purposes only...

This PR attempts fixing #325...

### DIFF
--- a/nx-X11/programs/Xserver/Xext/panoramiXsrv.h
+++ b/nx-X11/programs/Xserver/Xext/panoramiXsrv.h
@@ -32,6 +32,7 @@ extern unsigned long XRT_WINDOW;
 extern unsigned long XRT_PIXMAP;
 extern unsigned long XRT_GC;
 extern unsigned long XRT_COLORMAP;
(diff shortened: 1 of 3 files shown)


Ionic:
Mh, that conflicts with my render update, but I guess it doesn't make a big difference the update isn't fully working anyway, at least currently.

sunweaver:
Hi Ionic

On Tue Feb 14 21:17:50 2017 GMT+0100, Mihai Moldovan wrote:
> Mh, that conflicts with my render update, but I guess it doesn't make a big difference the update isn't fully working anyway, at least currently. 

My question regarding your work on the Render extension is: would it make sense to backport individual commits already? Esp those not directly related to files in Xserver/render/

???
Mike 
-- 

Ionic:
Not really, I've tried to keep it render-centric as much as possible, only changing other parts when absolutely necessary. It's highly inter-dependent and will likely just fall apart if you pick separate commits instead of the full deal.

sunweaver:
Will be superseded by #440. Thus, closing...

PR #332, opened 2017-02-16
Fix Composite extension in NX

sunweaver:
This is a WIP branch, addressing the fix of Composite extension support in nxagent.


uli42:
What is the state of this? What is the state of composite anyway?

PR #340, opened 2017-02-20
Use Mesa from Mesa-Upstream Git via git-subtree 

sunweaver:
This PR pulls in a massive commit history from Mesa upstream. The mechanism used is "git-subtree". It may be discussed if Git submodule is more appropriate, but I sense that git-subtree is the more state-of-the-art approach for pulling in Vcs-based 3rd-party projects.

@Ionic: @vatral: @salva: @uli42: @mikedep333: Request for comments!!!

PR #354, opened 2017-02-22
GLX 1.4 with Mesa 7.0.3

sunweaver:
This PR documents the status of the GLX version bumping efforts. Current approach is switching to GLX from 21 May 2008 and the corresponding Mesa release (7.0.3). The proto version will then be GLX 1.4.


uli42:
Is 1.5 too far ahead?

uli42:
What's the state of this? Just curious...

frantracer:
Did you consider to perform a _**git submodule**_ instead of the **_git subtree_**? I think it could be clearer and there is not need to import all the files into the project, just a reference

sunweaver:
Hi Fran,

I'd like to have git subtree being used, actually.

Basic info and howto:
https://www.atlassian.com/blog/git/alternatives-to-git-submodule-git-subtree

Here is one reason, why I prefer Git subtree for nx-libs's Mesa:
https://www.quora.com/Is-a-Git-subtree-better-than-submodules

Some reason against submodule:
https://codingkilledthecat.wordpress.com/2012/04/28/why-your-company-shouldnt-use-git-submodules/

The point is that we will use the Mesa repo in read-only mode. We will not change the Mesa code. However, if needed, we will patch it (via nx-X11/extras/Mesa.patches). So git subtree with commit log squashing is our way to go IMHO.

Please note, that #340 needs to be worked on first.

sunweaver:
will be superceded by #534

Issue #144, opened 2016-06-20
Use XORG_TOO_OLD macros instead of NXAGENT_SERVER for missing Xserver features


sunweaver:
Throughout the Xserver code, we use defined(NXAGENT_SERVER) to check if building nxagent.

When backporting extensions, we need this check and it denotes that actually the underlying Xserver code (mainly in Xserver/dix/ and Xserver/mi/) is too old for recent extension ABIs.

We should use another macro name for indicating a too old Xserver version, e.g. XORG_TOO_OLD.

See #125, #137 for earlier discussions.

PR #356, opened 2017-02-24
Explicitly mark outdated Xserver ABI sections

sunweaver:
This PR introduces various macro names for tagging nx-libs's specifc deviations from X.org because of missing ABI implementations or missing extensions.

Macro names:

  XSERVER_LACKS_PRIVATES_ABI
  XSERVER_LACKS_XACE_EXT
  XSERVER_OLD_DEVICE_ABI
  XSERVER_OLD_RESOURCE_NAME_ABI
  XSERVER_OLD_CLOSE_SCREEN_ABI
  XSERVER_EVENT_STILL_NEEDS_SEQNO

Marco names can be subject to discussion. The list is probably not yet complete.

This PR needs to wait for #311 to be merged. It will close issue #144.


sunweaver:
@Ionic: This PR introduces explicit names for ancient ABIs in nxagent's Xserver code that we still have to work around. Whenever new backports are done, we should use such macro names. If needed, more macro names should be invented (instead of using ifdef NXAGENT_SERVER everywhere).

Can you review?

Ionic:
I think dix lookup and dix privates are very closely related. Not sure if splitting them up makes a lot of sense, as we cannot really have one without the other.

nx-X11/programs/Xserver/randr/randr.c
```diff
@@ -518,7 +518,7 @@ RRDeliverResourceEvent(ClientPtr client, WindowPtr pWin)
     xRRResourceChangeNotifyEvent re = {
         .type = RRNotify + RREventBase,
         .subCode = RRNotify_ResourceChange,
-#ifdef NXAGENT_SERVER
+#ifdef XSERVER_EVENT_STILL_NEEDS_SEQNO
```
Ionic, line 521: Used but not defined.

Ionic (changes requested):

sunweaver:
@Ionic: XSERVER_LACKS_DIXLOOKUP_ABI now dropped.  XSERVER_EVENT_STILL_NEEDS_SEQNO now defined.

sunweaver:
@Ionic: next iteration.

nx-X11/programs/Xserver/randr/rrscreen.c
```diff
@@ -976,13 +976,13 @@ ProcRRSetScreenConfig(ClientPtr client)
...
-#else                           /* !defined(NXAGENT_SERVER) */
+#else                           /* !defined(XSERVER_LACKS_PRIVATES_ABI) */
     pDraw =
         SecurityLookupDrawable(stuff->drawable, client, DixWriteAccess);
     rc = pDraw ? Success : BadDrawable;
-#endif                          /* !defined(NXAGENT_SERVER) */
+#endif                          /* !defined(XSERVER_LACKS_PRIVATES_ABI) */
 
     if (rc != Success)
         return rc;
```
Ionic, line 988: All Randr files still have `NXAGENT` ifdefs for... well, I'm not sure what to call it. Mostly it's for operations on a linked list, which we don't have support for yet, so `XSERVER_LACKS_LINKED_LIST` might sound appropriate at first. But the linked list is actually part of (newer) Screen structures, so maybe `XSERVER_LACKS_SCREEN_LINKED_LISTS`? Maybe you have a better idea for the name. Semantically it would make sense to include them in this PR, though.

There are also `Screen->isGPU` blocks, but they are so rare that I don't think it makes a lot of sense to assign them a special macro.

Ionic (commented):
Okay, took a real close look at all `NXAGENT_SERVER` places now and came up with this additional potential.

PR #362, opened 2017-02-27
Switch to libxfont2 API

sunweaver:
Optionally allow building nxagent against libXfont1 (old) or libXfont2.


sunweaver:
@Ionic: PR #362 is rebased.

nx-X11/programs/Xserver/hw/nxagent/Font.c
```diff
@@ -73,7 +73,7 @@ is" without express or implied warranty.
 #define NXAGENT_DEFAULT_FONT_DIR      "/usr/share/nx/fonts"
 #define NXAGENT_ALTERNATE_FONT_DIR    "/usr/share/X11/fonts"
 #define NXAGENT_ALTERNATE_FONT_DIR_2  "/usr/share/fonts/X11"
-#define NXAGENT_ALTERNATE_FONT_DIR_3  "/usr/X11R6/lib/X11/fonts"
```
Ionic: This commit isn't really doing what it describes. Hard to do so, anyway, because `NXAGENT_ALTERNATE_FONT_PATH_4` is long gone.

Dropping support for `/usr/X11R6` might have an impact on very old *SuSE platforms, though even SLE 11 uses X.Org 7.4, so we should be in the clear.

nx-X11/programs/Xserver/hw/nxagent/Font.c
```diff
@@ -91,9 +91,10 @@ is" without express or implied warranty.
 /usr/share/fonts/X11/100dpi/,/usr/share/fonts/X11/TTF/"
 
 #define NXAGENT_ALTERNATE_FONT_PATH_3  \
-"/usr/X11R6/lib/X11/fonts/misc/,/usr/X11R6/lib/X11/fonts/Speedo/,\
-/usr/X11R6/lib/X11/fonts/Type1/,/usr/X11R6/lib/X11/fonts/75dpi/,\
-/usr/X11R6/lib/X11/fonts/100dpi/,/usr/X11R6/lib/X11/fonts/TTF/"
+"/usr/share/fonts/misc/,/usr/share/fonts/Speedo/,\
+/usr/share/fonts/Type1/,/usr/share/fonts/75dpi/,\
+/usr/share/fonts/100dpi/,/usr/share/fonts/TTF/,\
+/usr/NX/share/fonts/base"
```
Ionic: And adding `/usr/NX/share/fonts/base` plain out contradicts the commit's message, unless I'm reading it wrong.

nx-X11/programs/Xserver/hw/nxagent/Font.c
```diff
@@ -1567,24 +1568,6 @@ void nxagentVerifyDefaultFontPath(void)
 
     strcat(fontPath, NXAGENT_ALTERNATE_FONT_PATH_3);
   }
-  if (*fontPath == '\0')
```
Ionic: I faintly remember that you tried to remove this hunk during the libXfont2 default font selection mechanism, too, and it generally shouldn't hurt if fontPath always at least contains `default`, but why take the risk? I'd leave that in...

nx-X11/programs/Xserver/Imakefile
```diff
@@ -313,33 +321,44 @@ NXAGENTNXLIBS = -L/usr/sfw/lib \
...
 #elif defined(cygwinArchitecture)
 NXAGENTNXLIBS = -L../../../nxcomp  \
                 -L../../../nx-X11/exports/lib \
                 -L../../../nxcompshad \
                 -lXcomp \
                 -lXcompshad \
-                -lXrender -lXfixes -lXfont -lXcomposite -lXdmcp \
-                -lNX_X11 -lXext
-#elif defined(OpenBSDArchitecture)
-NXAGENTNXLIBS = -L../../../nxcomp  \
```
Ionic: Uhm, you're dropping the `OpenBSDArchitecture` block, which **does** have special settings (like using `libkvm`.)

nx-X11/programs/Xserver/hw/nxagent/Image.c
```diff
@@ -98,9 +98,9 @@ int nxagentAlphaCompat  = 0;
  * displays having different byte order.
  */
 
-extern void BitOrderInvert(unsigned char *, int);
-extern void TwoByteSwap(unsigned char *, register int);
-extern void FourByteSwap(register unsigned char *, register int);
+extern void nxagentBitOrderInvert(unsigned char *, int);
```
Ionic, line 101: I think you have forgotten to include the new file `BitmapUtils.c` in this commit. Can't really comment on it until that's done...

nx-X11/config/cf/Imake.tmpl
```diff
@@ -564,6 +564,9 @@ X_BYTE_ORDER = ByteOrder
 #ifndef HasFfs
```
Ionic, line 564: Should probably mention

```
commit 05a793f5b3c40747d5a92a076def7f4fb673c7e7
Author: Keith Packard <keithp@keithp.com>
Date:   Tue Sep 1 18:50:55 2015 -0700

    dix: Switch to the libXfont2 API (v2)

    This new libXfont API eliminates exposing internal X server symbols to
    the font library, replacing those with a struct full of the entire API
    needed to use that library.

    v2: Use libXfont2 instead of libXfont_2

    Reviewed-by: Adam Jackson <ajax@redhat.com>
```

?

nx-X11/programs/Xserver/hw/nxagent/Imakefile
```diff
@@ -257,7 +261,11 @@ all:: $(OBJS)
 
 LinkSourceFile(stubs.c,$(SERVERSRC)/Xi)
 LinkSourceFile(xpstubs.c,$(SERVERSRC)/dix)
-SpecialCObjectRule(Init,$(ICONFIGFILES),$(_NOOP_))
+SpecialCObjectRule(Init,$(ICONFIGFILES),$(_NOOP_) $(FONT_DEFINES))
```
Ionic: Drop `$(_NOOP_)` here.

Btw, I wonder if we actually need the `$(NULL)` stuff if the undefined `$(_NOOP_)` also works, but it's the first time I saw this... too late to change everything now.

nx-X11/programs/Xserver/mi/Imakefile
```diff
@@ -179,6 +183,9 @@ NormalLibraryTarget(cbrt,cbrt.o)
 
 SpecialCObjectRule(miinitext,$(ICONFIGFILES),$(EXT_DEFINES))
 SpecialCObjectRule(miscrinit,$(ICONFIGFILES),$(EXT_DEFINES))
+SpecialCObjectRule(miglblt,$(ICONFIGFILES),$(FONT_DEFINES))
+SpecialCObjectRule(mipolytext,$(ICONFIGFILES),$(FONT_DEFINES))
```
Ionic, line 183: `mipolytext` is not really needed here, why add it?

Ionic: Nevermind, it's using `GetGlyphs()`, which is part of `libXfont1`.

Ionic: Hmm, no, it's not, in reality we're using a macro to redefine `GetGlyphs` to `dixGetGlyphs` so that we do not actually have to link to `libXfont`.

debian/rules
```diff
@@ -51,6 +51,10 @@ override_dh_auto_install:
 
 override_dh_auto_build:
 
+	if ! pkg-config --modversion xfont2 1>/dev/null 2>/dev/null; then \
+	    sed -e 's/#define.*HasXfont2.*NO/#define HasXfont2 YES/' -i nx-X11/config/cf/host.def \
+	fi
```
Ionic: That's indeed not a good way to handle this... more work required here.

Ionic (changes requested):

nx-X11/programs/Xserver/Imakefile
sunweaver: Indeed. Another rebasing flaw. Amended.

nx-X11/programs/Xserver/hw/nxagent/Font.c
sunweaver: This commit was all bogus and sneaked in while rebasing. Commit has been removed.

nx-X11/programs/Xserver/hw/nxagent/Font.c
sunweaver: dito (commit removed)

nx-X11/programs/Xserver/hw/nxagent/Font.c
sunweaver: dito (commit removed)

nx-X11/programs/Xserver/hw/nxagent/Image.c
sunweaver: Indeed. Now added. (Luckily, I had not done a git clean in my working copy since I wrote down the changes on this branch).

nx-X11/config/cf/Imake.tmpl
sunweaver: Yes. Good point. X.org commit reference added to Git history of nx-libs.

nx-X11/programs/Xserver/hw/nxagent/Imakefile
sunweaver: Yes. Thanks for the comment nonetheless.

nx-X11/programs/Xserver/mi/Imakefile
sunweaver: With mipolytext removed from Imakefile's SpecialCObjectRule(...) (i.e. not defining FONT_DEFINES, I get the build build failure:

```
mi/libmi.a(mipolytext.o): In function `miPolyText':
./nx-X11/programs/Xserver/mi/mipolytext.c:82: undefined reference to `dixGetGlyphs'
mi/libmi.a(mipolytext.o): In function `miPolyText8':
./nx-X11/programs/Xserver/mi/mipolytext.c:105: undefined reference to `dixGetGlyphs'
mi/libmi.a(mipolytext.o): In function `miPolyText16':
./nx-X11/programs/Xserver/mi/mipolytext.c:128: undefined reference to `dixGetGlyphs'
mi/libmi.a(mipolytext.o): In function `miImageText':
./nx-X11/programs/Xserver/mi/mipolytext.c:154: undefined reference to `dixGetGlyphs'
mi/libmi.a(mipolytext.o): In function `miImageText8':
./nx-X11/programs/Xserver/mi/mipolytext.c:176: undefined reference to `dixGetGlyphs'
mi/libmi.a(mipolytext.o):./nx-X11/programs/Xserver/mi/mipolytext.c:195: more undefined references to `dixGetGlyphs' follow
collect2: error: ld returned 1 exit status
```

So, it has to stay.

debian/rules
sunweaver: It actually is. There is no cleaner approach for building a Debian source package against various distro versions. I have added some clean-up stuff around it, but it basically needs to stay like this.

sunweaver:
@Ionic: next iteration, please. Thanks!

Ionic:
libXfont2 availability:

Fedora: 24+
OpenSUSE: none
CentOS/RHEL: none
SLE: none

nx-X11/programs/Xserver/Imakefile
```diff
@@ -313,33 +321,59 @@ NXAGENTNXLIBS = -L/usr/sfw/lib \
...
                 -L../../../nxcompshad \
                 -lkvm \
                 -lXcomp \
                 -lXcompshad \
-                -lXrender -lXfixes -lXfont -lXcomposite -lXinerama -lXdmcp \
-                -lNX_X11 -lXext
+                -lXrender \
+                -lXfixes \
+                -lXfont \
+                -lXcomposite \
```
Ionic: Missing `-lXinerama`.

debian/rules
```diff
@@ -51,8 +57,18 @@ override_dh_auto_install:
 
 override_dh_auto_build:
 
+	if ! pkg-config --modversion xfont2 1>/dev/null 2>/dev/null; then \
+	    cp nx-X11/config/cf/host.def nx-X11/config/cf/host.def.bak; \
+	    sed -e 's/#define.*HasXfont2.*YES/#define HasXfont2 NO/' -i nx-X11/config/cf/host.def; \
```
Ionic: I really don't like sedding the imake file.

Also, the -i option is non-portable and implementations behave differently. GNU sed expects a backup extension right after the -i flag without whitespace, while BSD sed expects the backup extension to be separated from the flag with whitespace.

Better don't use the in-place version; since you're already copying the file just run the command on that and overwrite the original file.

I'd still like to see a better way of handling this...

It would be easier to check `pkg-config --modversion xfont2` in the main Makefile and add pass `-DHasXfont2=\"YES\"`  or `-UHasXfont2` to the build process directly based on the return value. Will need some testing to see if it works, though. Especially since we don't want it to default to YES again if we're disabling it by passing `-UHasXfont2`.

nx-X11/programs/Xserver/hw/nxagent/Imakefile
Ionic: `$(_NOOP_)` still on that line.

nx-X11/programs/Xserver/mi/Imakefile
Ionic: In order to not rewrite `GetGlyphs()` to `dixGetGlyphs()` when using `libXfont2`, okay.

Ionic (changes requested):

debian/rules
sunweaver: @Ionic: I have dropped the sedding of host.def now completely and do the auto-detection in the main Makefile. The package maintainer can then choose what Xfont API to use by making Xfont2 available (or not). Implying that "or not" means, that Xfont(1) is available instead.

nx-X11/programs/Xserver/Imakefile
sunweaver: -lXinerama re-added now.

debian/rules
sunweaver: Agreeing here. See the latest approach just pushed.

 * FONT_DEFINES and XLIBFONT now get defined in Server.tmpl.
 * both get overridden in the main Makefile (there is the pkg-config based auto-detection code now)
 * debian/control simply build-depends on libxfont-dev (which is API 2 or 1 depending on what Debian version we build against)
 * nx-libs.spec has conditional BRs depending on what distro and distro version we build against

sunweaver:
> in nx-X11/programs/Xserver/hw/nxagent/Imakefile:
> $(_NOOP_) still on that line.

Fixed.

sunweaver:
@Ionic: next and hopefully last iteration.

Ionic:
I made a mistake... libXfont2 isn't available on OpenSUSE at all currently. Only in OpenSUSE Tumbleweed, which is roughly equivalent to Debian Unstable. So currently no Xfont2 builds on OpenSUSE.

nx-libs.spec
```diff
@@ -38,7 +38,11 @@ BuildRequires:  pkgconfig(pixman-1) >= 0.13.2
 BuildRequires:  pkgconfig(x11)
 BuildRequires:  pkgconfig(xext)
 BuildRequires:  pkgconfig(xpm)
+%if 0%{?suse_version} >= 4210
```
Ionic: Comment this out for now please and only use `pkgconfig(xfont) >= 1.4.2`.

Ionic (changes requested):

nx-libs.spec
sunweaver: If I look at RPM search, I get this:

```
OpenSuSE 	ftp5.gwdg.de/pub/opensuse/repositories/X11:/XOrg/openSUSE_Leap_42.1/x86_64/libXfont2-devel-2.0.1-1.1.x86_64.rpm
```

nx-X11/config/cf/host.def
```diff
@@ -390,6 +390,12 @@
 #define BuildXKBlib		YES
 
 /*
+ * If you want to build against libXfont2, rathen than libXfont(1), uncomment this.
```
Ionic: Typo, "rathen".

nx-X11/config/cf/Server.tmpl
```diff
@@ -70,3 +70,11 @@ GLX_DEFINES = GlxDefines
 #ifdef XorgVersion
 #include <xorg.tmpl>
 #endif
+
+#ifdef HasXfont2
```
Ionic: Do variables when passed directly when calling make really override any other definitions? Normally that's the case.

Yes, that'll probably work (since `FONT_DEFINES` and `XFONTLIB` are always overridden via our main Makefile), but it's difficult to understand that this section here essentially normally does nothing.

Maybe we should provide comments clarifying that this is just a fallback if these variables are not overridden manually.

Meh, but it gets worse: since we override the variables, `HasXfont2` is essentially useless and won't tell what library is actually being used.

This is why I originally suggested to do the check and use `-DHasXfont2=\"YES\"` in the main Makefile.

Ionic (changes requested):

nx-libs.spec
Ionic: This is what originally mislead me as well. It's not part of the official OpenSUSE repository though, but a custom one. Like our repositories for Debian...

nx-X11/config/cf/Server.tmpl
sunweaver: Without our main Makefile, the nx-libs build defaults to libXfont1. This will die anyway, once we drop imake. It is a hack due to the limitations of imake, so we should not spend to much energy on this, IMHO.

nx-libs.spec
sunweaver: > This is what originally mislead me as well. It's not part of the official OpenSUSE repository though, but a custom one. Like our repositories for Debian...

ouch. Ok. Commenting stuff out.

nx-X11/config/cf/host.def
sunweaver: Oh dear. Fixed.

Ionic (approved):

PR #331, opened 2017-02-15
Introduce SetNotifyFd ABI to nxagent Xserver

sunweaver:
This PR requires PR #330 to be merged.

The SetNotifyFd ABI is needed for building against libXfont2 (i.e. for a +/- clean cherry-pick from X.org).

PR #376, opened 2017-03-03
Various Xserver/os backports

sunweaver:
This PR requires #331 to be merged first.

### DIFF
--- a/nx-X11/programs/Xserver/Xext/shm.c
+++ b/nx-X11/programs/Xserver/Xext/shm.c
@@ -150,7 +150,6 @@ static ShmFuncs fbFuncs = {fbShmCreatePixmap, fbShmPutImage};
 
 
 #if defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__CYGWIN__)
-#include <sys/signal.h>
 
 static Bool badSysCall = FALSE;
 
@@ -167,7 +166,7 @@ static Bool CheckForShmSyscall()
     int shmid = -1;
 
     /* If no SHM support in the kernel, the bad syscall will generate SIGSYS */
-    oldHandler = signal(SIGSYS, SigSysHandler);
+    oldHandler = OsSignal(SIGSYS, SigSysHandler);
 
     badSysCall = FALSE;
     shmid = shmget(IPC_PRIVATE, 4096, IPC_CREAT);
@@ -182,7 +181,7 @@ static Bool CheckForShmSyscall()
         /* Allocation failed */
         badSysCall = TRUE;
     }
-    signal(SIGSYS, oldHandler);
+    OsSignal(SIGSYS, oldHandler);
     return(!badSysCall);
 }
 
--- a/nx-X11/programs/Xserver/include/dixstruct.h
+++ b/nx-X11/programs/Xserver/include/dixstruct.h
@@ -24,6 +24,7 @@ SOFTWARE.
 #ifndef DIXSTRUCT_H
 #define DIXSTRUCT_H
 
+#include "client.h"
 #include "dix.h"
 #include "resource.h"
 #include "cursor.h"
@@ -141,6 +142,8 @@ typedef struct _Client {
     long    smart_start_tick;
     long    smart_stop_tick;
     long    smart_check_tick;
+
+    ClientIdPtr  clientIds;
 }           ClientRec;
 
 /*
--- a/nx-X11/programs/Xserver/include/dixstruct.h
+++ b/nx-X11/programs/Xserver/include/dixstruct.h
@@ -95,6 +95,7 @@ typedef struct _Client {
     void        *requestBuffer;
     void        *osPrivate;	/* for OS layer, including scheduler */
     Bool        swapped;
+    Bool        local;
     ReplySwapPtr pSwapReplyFunc;
     XID         errorValue;
     int         sequence;
--- a/nx-X11/programs/Xserver/include/dixstruct.h
+++ b/nx-X11/programs/Xserver/include/dixstruct.h
@@ -150,18 +150,19 @@ extern long SmartScheduleTime;
 extern long SmartScheduleInterval;
 extern long SmartScheduleSlice;
 extern long SmartScheduleMaxSlice;
-extern unsigned long SmartScheduleIdleCount;
-extern Bool SmartScheduleDisable;
-extern Bool SmartScheduleIdle;
-extern Bool SmartScheduleTimerStopped;
-extern Bool SmartScheduleStartTimer(void);
-#ifdef NXAGENT_SERVER
-extern Bool SmartScheduleStopTimer(void);
+#ifdef HAVE_SETITIMER
+#if HAVE_SETITIMER
+extern Bool SmartScheduleSignalEnable;
+#else
+#define SmartScheduleSignalEnable FALSE
 #endif
+#endif
+extern void SmartScheduleStartTimer(void);
+extern void SmartScheduleStopTimer(void);
 #define SMART_MAX_PRIORITY  (20)
 #define SMART_MIN_PRIORITY  (-20)
 
-extern Bool SmartScheduleInit(void);
+extern void SmartScheduleInit(void);
 
 /* This prototype is used pervasively in Xext, dix */
 #define DISPATCH_PROC(func) int func(ClientPtr /* client */)
--- a/nx-X11/programs/Xserver/os/Imakefile
+++ b/nx-X11/programs/Xserver/os/Imakefile
@@ -128,6 +128,7 @@ TMEMCMP_OBJS = timingsafe_memcmp.o
 BOOTSTRAPCFLAGS = 
            SRCS = WaitFor.c		\
                   access.c		\
+                  client.c		\
                   connection.c		\
                   io.c			\
                   $(COLOR_SRCS)		\
@@ -151,6 +152,7 @@ BOOTSTRAPCFLAGS =
 
            OBJS = WaitFor.o		\
                   access.o		\
+                  client.o		\
                   connection.o		\
                   io.o			\
                   $(COLOR_OBJS)		\
--- a/nx-X11/programs/Xserver/os/Imakefile
+++ b/nx-X11/programs/Xserver/os/Imakefile
@@ -43,6 +43,7 @@ NX_DEFINES = -DNX_TRANS_SOCKET		\
              -DNX_TRANS_SLEEP		\
              -DNX_TRANS_EXIT		\
              -DNX_TRANS_WAKEUP=1000	\
+             -DNXAGENT_SERVER		\
              $(NULL)
 
 #            -DNX_TRANS_WARN		\
--- a/nx-X11/programs/Xserver/os/Imakefile
+++ b/nx-X11/programs/Xserver/os/Imakefile
@@ -25,6 +25,8 @@
 
 #include <Server.tmpl>
 
+NULL =
+
 /*
  * If you have any extra files to be put into the library, define them here.
  */
@@ -33,17 +35,18 @@
 
 NX_INCLUDES = -I../../../../nxcomp
 
-NX_DEFINES = -DNX_TRANS_SOCKET        \
-             -DNX_TRANS_AUTH          \
-             -DNX_TRANS_FOPEN         \
-             -DNX_TRANS_SLEEP         \
-             -DNX_TRANS_EXIT          \
-             -DNX_TRANS_WAKEUP=1000
+NX_DEFINES = -DNX_TRANS_SOCKET		\
+             -DNX_TRANS_AUTH		\
+             -DNX_TRANS_FOPEN		\
+             -DNX_TRANS_SLEEP		\
+             -DNX_TRANS_EXIT		\
+             -DNX_TRANS_WAKEUP=1000	\
+             $(NULL)
 
-#            -DNX_TRANS_WARN          \
-#            -DNX_TRANS_INFO          \
-#            -DNX_TRANS_TEST          \
-#            -DNX_TRANS_DEBUG         \
+#            -DNX_TRANS_WARN		\
+#            -DNX_TRANS_INFO		\
+#            -DNX_TRANS_TEST		\
+#            -DNX_TRANS_DEBUG		\
 
 #endif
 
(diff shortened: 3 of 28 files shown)


sunweaver:
@Ionic: here is the bulk backport pull request... Once this one is in, we can look at the libXfont2 PR.

uli42 (commented):
Hmm, this seems to affect ssh-X11 forwarding and might therefore have impact on the x2go setup (which, IIRC, uses ssh -X to trigger ssh optimizations)

Ionic:
Github does something weird with this PR. Allegedly, 7ad1f8cc17cbdd88f53bb767aa20b5b816d1aa3a is supposed to come before 495889247c6e07b21dc2f90fd38856c7d68031ab, but that's impossible because the later introduced the change that is then changed again by the former...

Ionic:
You somehow managed to break Github with this PR. Most commits seem to be shown out of order - but the git history looks fine. I'll create a github bug report for this...

Ionic:
Okay, seems to be a known issue and they won't do anything about it, c.f. https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/

nx-X11/programs/Xserver/os/connection.c
```diff
@@ -988,7 +988,7 @@ CloseDownConnection(ClientPtr client)
     OsCommPtr oc = (OsCommPtr)client->osPrivate;
 
     if (FlushCallback)
-	CallCallbacks(&FlushCallback, NULL);
+	CallCallbacks(&FlushCallback, client);
```
Ionic: This commit doesn't fit in, really... Semantically, when looking at the previous one, it does, but the commit message references a completely different upstream commit that has nothing to do with your changes.

Your changes match this upstream commit:

```
commit b380f3ac51f40ffefcde7d3db5c4c149f274246d
Author: Michel Dänzer <michel.daenzer@amd.com>
Date:   Tue Aug 2 17:53:01 2016 +0900

    dix: Pass ClientPtr to FlushCallback

    This change has two effects:

    1. Only calls FlushCallbacks when we're actually flushing data to a
       client. The unnecessary FlushCallback calls could cause significant
       performance degradation with compositing, which is significantly
       reduced even without any driver changes.

    2. By passing the ClientPtr to FlushCallbacks, drivers can completely
       eliminate unnecessary flushing of GPU commands by keeping track of
       whether we're flushing any XDamageNotify events to the client for
       which the corresponding rendering commands haven't been flushed to
       the GPU yet.

    Reviewed-by: Adam Jackson <ajax@redha.com>
    Signed-off-by: Michel Dänzer <michel.daenzer@amd.com>
```

Ionic: If you really meant to backport the commit you mentioned in this commit message, you will have to do that as well, but I don't know if it makes sense at this point in time, since the mentioned upstream commit fixes a regression that appeared after upstream changed to its new `ospoll`-/`epoll`-based implementation. We do not have that yet.

nx-X11/programs/Xserver/os/Imakefile
```diff
@@ -121,16 +124,51 @@ TMEMCMP_OBJS = timingsafe_memcmp.o
...
-           OBJS = WaitFor.o access.o connection.o io.o $(COLOR_OBJS) \
-                  osinit.o utils.o log.o auth.o mitauth.o secauth.o \
-                  $(XDMAUTHOBJS) $(RPCOBJS) xdmcp.o OtherObjects \
-                  xstrans.o $(SNPRINTF_OBJS) $(STRLCAT_OBJS) \
-		  $(REALLOCARRAY_OBJS) xprintf.o $(TMEMCMP_OBJS)
+           SRCS = WaitFor.c		\
+                  access.c		\
+                  connection.c		\
+                  io.c			\
+                  $(COLOR_SRCS)		\
```
Ionic: You've broken the escape character lineup with tabsize = 8 here and on several occasions below that line.

I guess you don't see it because you set your editor to a different tabsize, just mentioning it.

Also, the `pkg-config` line way down is too long even for tabsize = 8, so while you're at it, make sure the escape chars line up in that hunk?

nx-X11/programs/Xserver/os/Imakefile
```diff
@@ -121,16 +124,51 @@ TMEMCMP_OBJS = timingsafe_memcmp.o
...
+                  mitauth.c		\
+                  secauth.c		\
+                  $(XDMAUTHSRCS)	\
+                  $(RPCSRCS)		\
+                  xdmcp.c		\
+                  OtherSources		\
+                  xstrans.c		\
+                  $(SNPRINTF_SRCS)	\
+                  $(STRLCAT_SRCS)	\
+                  xprintf.c		\
```
Ionic: Try to keep the order. In the original string, `xprintf.c` comes after `$(REALLOCARRAY_SRCS)`. This isn't a problem per se, but in the OBJS definition, the ordering is different again: first `$(REALLOCARRAY_OBJS)`, then `$(TMEMCMP_OBJS)`, then `xprintf.o`. Please try to keep both definitions in synch, since humans are pretty good at pattern matching and would expect entries to be at the same relative location if all the others are. That is, I don't care about the actual order, but I'd like to have a consistent one.

In both cases, too, `$(REALLOCARRAY_SRCS)` is followed by spaces, whereas you converted all the other lines to use tabs.

nx-X11/programs/Xserver/os/access.c
```diff
@@ -496,7 +496,7 @@ DefineSelf (int fd)
 void
 DefineSelf (int fd)
 {
-#if !defined(TCPCONN) && !defined(UNIXCONN) && !defined(MNX_TCPCONN)
+#if !defined(TCPCONN) && !defined(UNIXCONN)
```
Ionic, line 495: Making a note: `MNX_TCPCONN` left over in `nx-X11/lib/X11/ConnDis.c`. It doesn't fit in with the `Xserver/os` changes, but we probably want to remove it outside of this PR as well.

nx-X11/programs/Xserver/os/access.c
```diff
@@ -1314,57 +1323,64 @@ LocalClientCredAndGroups(ClientPtr client, int *pUid, int *pGid,
...
-	    return -1;
-    if (pUid != NULL)
-	    *pUid = peercred.uid;
-    if (pGid != NULL)
-	    *pGid = peercred.gid;
+    if (getsockopt(fd, SOL_SOCKET, SO_PEERCRED, &peercred, &so_len) == -1) {
+	FreeLocalClientCreds(lcc);
+	return -1;
+    }
+     lcc->euid = peercred.uid;
```
Ionic: Unnecessary whitespace error here?

nx-X11/programs/Xserver/os/connection.c
```diff
@@ -534,8 +534,8 @@ AuthAudit (ClientPtr client, Bool letin,
     char addr[128];
     char *out = addr;
 
-    int client_uid;
     char client_uid_string[32];
```
Ionic: Hm, not an immediate problem with your backport, but it looks like `client_uid_string` had size 64 in the original code.

That change was part of

```
commit a7b944f0d96c3e0e15e75378a04def1ac96089fb
Author: Alan Coopersmith <alan.coopersmith@sun.com>
Date:   Wed Nov 1 16:17:49 2006 -0800

    If getpeerucred() is available, include pid & zoneid in audit messages too
```

Backporting the whole commit doesn't make sense since it introduces `#ifdef HAS_GETPEERUCRED` blocks which are then removed again with the commit you backported now, but at least cherry-picking that one change might.

nx-X11/programs/Xserver/Xext/shm.c
```diff
@@ -167,7 +166,7 @@ static Bool CheckForShmSyscall()
     int shmid = -1;
 
     /* If no SHM support in the kernel, the bad syscall will generate SIGSYS */
-    oldHandler = signal(SIGSYS, SigSysHandler);
+    oldHandler = OsSignal(SIGSYS, SigSysHandler);
```
Ionic, line 169: What about `nx-X11/programs/Xserver/os/utils.c`? Missing changes for `SIGCHLD` there.

nx-X11/programs/Xserver/os/connection.c
```diff
@@ -857,59 +859,44 @@ EstablishNewConnections(ClientPtr clientUnused, void * closure)
...
-	    ErrorConnMax(new_trans_conn);
-	}
-      }
-#ifndef WIN32
-    }
-#endif
     return TRUE;
 }
 
+static void
```
Ionic, line 899: Something in this commit looks odd...

The upstream commit 7b02f0b87ec2fa0cc5a65307a1fd55c671cec884 has this hunk:

```diff
@@ -1417,8 +1398,7 @@ ListenOnOpenFD(int fd, int noxauth)
     ListenTransConns[ListenTransCount] = ciptr;
     ListenTransFds[ListenTransCount] = fd;

-    FD_SET(fd, &WellKnownConnections);
-    FD_SET(fd, &AllSockets);
+    SetNotifyFd(fd, QueueNewConnections, X_NOTIFY_READ, NULL);

     /* Increment the count */
     ListenTransCount++;
```

which is missing in your backport.

While the function in question doesn't exist in our source tree, I hope we're not missing out on the `SetNotifyFd()` call. Probably not, since the function is not even used on anything but the Darwin and Wayland DDX'.

Ionic (changes requested):

nx-X11/programs/Xserver/os/connection.c
sunweaver: Oh dear. Sorry for this flaw. You are, of course, right on this. Amended locally already.

sunweaver: There actually was another issue with this commit. I backported the above commit from 2016 and then X.org commit c65f610e12f9df168d5639534ed3c2bd40afffc8 from 2010. I re-did this commit and the 2010 commit and wove them all into this one with extra backport commit message.

nx-X11/programs/Xserver/os/io.c
```diff
@@ -965,6 +965,10 @@ WriteToClient (ClientPtr who, int count, const void *__buf)
 	  CriticalOutputPending = FALSE;
 	  NewOutputPending = FALSE;
 	}
+
+	if (FlushCallback)
+	    CallCallbacks(&FlushCallback, NULL);
+
 	return FlushClient(who, oc, buf, count);
```
sunweaver: @Ionic: this part of the commit now becomes obsolete. Has been moved to FlushClient() function later in 2016.

sunweaver: This commit has been removed now and will appear as reference in  65b6a62bc726b5e4f4833ed1e27733e22fabcfa8.

nx-X11/programs/Xserver/os/Imakefile
sunweaver: > You've broken the escape character lineup with tabsize = 8 here and on several occasions below that line.
>
> I guess you don't see it because you set your editor to a different tabsize, just mentioning it.
> 
> Also, the pkg-config line way down is too long even for tabsize = 8, so while you're at it, make sure the escape chars line up in that hunk?

@Ionic:I use tab size 8 here and the breakage only appears in the diff output (because of the extra "+" at the beginning of the diff line. In the file itself, the tabbing is ok.

nx-X11/programs/Xserver/os/Imakefile
sunweaver: > Try to keep the order. In the original string, xprintf.c comes after $(REALLOCARRAY_SRCS). This isn't a problem per se, but in the OBJS definition, the ordering is different again: first $(REALLOCARRAY_OBJS), then $(TMEMCMP_OBJS), then xprintf.o. Please try to keep both definitions in synch, since humans are pretty good at pattern matching and would expect entries to be at the same relative location if all the others are. That is, I don't care about the actual order, but I'd like to have a consistent one.
>
> In both cases, too, $(REALLOCARRAY_SRCS) is followed by spaces, whereas you converted all the other lines to use tabs.

Nice catch! Fixed.

nx-X11/programs/Xserver/os/access.c
sunweaver: > Making a note: MNX_TCPCONN left over in nx-X11/lib/X11/ConnDis.c. It doesn't fit in with the Xserver/os changes, but we probably want to remove it outside of this PR as well.

See new issue #399...

nx-X11/programs/Xserver/os/connection.c
sunweaver: Good catch. Fixed.

nx-X11/programs/Xserver/Xext/shm.c
sunweaver: @Ionic:

> What about nx-X11/programs/Xserver/os/utils.c? Missing changes for SIGCHLD there.

Oh dear, this commit has been completely redone. Basically by pulling in all recent and ancient changes to the smart scheduler ABI.

nx-X11/programs/Xserver/include/dixstruct.h
```diff
@@ -160,7 +160,7 @@ extern Bool SmartScheduleIdle;
 extern Bool SmartScheduleTimerStopped;
 extern Bool SmartScheduleStartTimer(void);
 #ifdef NXAGENT_SERVER
-extern Bool SmartScheduleStopTimer(void);
+extern void SmartScheduleStopTimer(void);
```
sunweaver: This commit ("Xserver/include/dixstruct.h: Fix type mismatch for SmartScheduleStopTimer() prototype.") will get dropped with my next version of this PR branch.

nx-X11/programs/Xserver/os/connection.c
sunweaver: > [...]
>
> While the function in question doesn't exist in our source tree, I hope we're not missing out on the SetNotifyFd() call. Probably not, since the function is not even used on anything but the Darwin and Wayland DDX'.

The ListenOnOpenFD() function was specially added to the Xserver ABI for XWayland. It may be needed for nxagent later, if we want to include external socket handling support for nxagent. But for now, we should be fine.

sunweaver:
@Ionic: ok. I think I worked myself through all your notes / requests.

The signal() ->  OsSignal() commit got fully reworked now and ships many more backports in a bundle.

Also, I added some command line parsing commit for the "-x" option at the end of the commit series.

Before merging this PR, I'd like to do some more performance tests, as I am not 100% sure, the Signal and SmartScheduler changes won't bring us performance problems.

sunweaver:
@Ionic: ok, I did various tests. Performance seems to be slightly improve with this PR. However, it seems we need to keep an eye on drawing glitches, esp. when using shadow mode.

nx-X11/programs/Xserver/os/access.c
Ionic: Ping

nx-X11/programs/Xserver/os/Imakefile
Ionic: Okay, almost done. Line 203 has a space char after `-I.`.

nx-X11/programs/Xserver/Xext/shm.c
Ionic: Phew... this one got really huge... and is still missing a lot of commit references for quite some of the `itimer` changes.

Also, I noticed `SMART_SCHEDULE_DEFAULT_INTERVAL` and `SMART_SCHEDULE_MAX_SLICE` changed upstream. If we change the default values as well, this might influence performance. No idea if for the better or worse, though.

One commit that should have been mentioned:

```
commit e10ba9e4b52269b2ac75c4802dce4ca47d169657
Author: Keith Packard <keithp@keithp.com>
Date:   Wed Nov 11 22:02:01 2015 -0800

    Remove non-smart scheduler. Don't require setitimer.

    This allows the server to call GetTimeInMillis() after each request is
    processed to avoid needing setitimer. -dumbSched now turns off the
    setitimer.

    Reviewed-by: Adam Jackson <ajax@redhat.com>
```

It also looks like this hunk is missing from `nx-X11/programs/Xserver/hw/nxagent/NXdispatch.c`:

```diff
@@ -415,7 +418,10 @@ Dispatch(void)
                    result = BadLength;
                else
                    result = (* client->requestVector[MAJOROP])(client);
-
+
+               if (!SmartScheduleSignalEnable)
+                   SmartScheduleTime = GetTimeInMillis();
+
                if (result != Success)
                {
                    if (client->noClientException != Success)
```

But wait, there's more!

In `nx-X11/programs/Xserver/hw/nxagent/Display.h`:

```diff
 #define nxagentDisableTimer() \
 \
-    if (SmartScheduleTimerStopped == 0) \
-    { \
-      SmartScheduleStopTimer(); \
-    } \
-\
-    SmartScheduleDisable = 1;
+    SmartScheduleStopTimer(); \
+    SmartScheduleSignalEnable = FALSE;

 /*
  * File descriptor currently used by
```

`SmartScheduleSignalEnable` is set unguarded. This will go boom if `HAVE_SETITIMER` is undefined, because in that case `SmartScheduleSignalEnable` is macro'd to `FALSE` and we're then effectively setting `FALSE = FALSE`.

nx-X11/programs/Xserver/os/WaitFor.c
```diff
@@ -213,18 +213,10 @@ WaitForSomething(int *pClientsReady)
...
-		wt = &waittime;
-	    }
-	    else
-	    {
-		XFD_COPYSET (&ClientsWithInput, &clientsReadable);
-		break;
-	    }
+	    someReady = TRUE;
+	    waittime.tv_sec = 0;
+	     waittime.tv_usec = 0;
```
Ionic: Unnecessary whitespace error.

Ionic (changes requested):
Not just because of the whitespace error, but we've got a few other things that aren't perfect yet...

nx-X11/programs/Xserver/os/Imakefile
sunweaver: Ok, fixed now.

nx-X11/programs/Xserver/os/access.c
sunweaver: > Unnecessary whitespace error here?

Sorry, missed this one. Fixed now

nx-X11/programs/Xserver/os/WaitFor.c
sunweaver: > Unnecessary whitespace error. (in os/WaitFor.c)

Fixed.

nx-X11/programs/Xserver/Xext/shm.c
sunweaver: @Ionic: Display.h and NXdispatch.c have been amended.

sunweaver:
@ionic: next review iteration.

Ionic (approved):
Everything has been fixed and I wasn't able to come up with anything new while again looking at the `OsSignal` change.

PR #333, opened 2017-02-16
Collection of compiler warning fixes.

sunweaver:
WIP PR, collecting fixes for various compiler warnings here and there.

PR #378, opened 2017-03-03
Various nxagent and Xserver (compiler warning) fixes

uli42:
These are some fixes that I have done back in December. Some of them are pretty obvious an can be cherry-picked, I'd say.

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -1235,9 +1235,11 @@ static void nxagentParseOptions(char *name, char *value)
     return;
 #else
     if (PANORAMIX_DISABLED_COND && RRXINERAMA_DISABLED_COND)
+    {
       nxagentChangeOption(Xinerama, 0);
       fprintf(stderr, "Warning: XINERAMA extension has been disabled on %s startup.\n", __progname);
       return;
+    }
 
     if (!strcmp(value, "1"))
     {
--- a/nx-X11/programs/Xserver/hw/nxagent/Rootless.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Rootless.c
@@ -334,6 +334,7 @@ void nxagentRootlessRestack(unsigned long children[], unsigned int nchildren)
 
   if (!ntoplevel)
   {
+    free(toplevel);
     return;
   }
 
--- a/nx-X11/programs/Xserver/hw/nxagent/Rootless.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Rootless.c
@@ -965,7 +965,7 @@ void nxagentImportProperty(Window window,
       }
     }
 
-    if ((wmHints.flags & IconWindowHint) && (wmHints.icon_window =! None))
+    if ((wmHints.flags & IconWindowHint) && (wmHints.icon_window != None))
     {
       WindowPtr icon = nxagentWindowPtr(wmHints.icon_window);
 
@@ -986,7 +986,7 @@ void nxagentImportProperty(Window window,
       }
     }
 
-    if ((wmHints.flags & IconMaskHint) && (wmHints.icon_mask =! None))
+    if ((wmHints.flags & IconMaskHint) && (wmHints.icon_mask != None))
     {
       PixmapPtr icon = nxagentPixmapPtr(wmHints.icon_mask);
 
--- a/nx-X11/programs/Xserver/hw/nxagent/Rootless.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Rootless.c
@@ -351,14 +351,14 @@ void nxagentRootlessRestack(unsigned long children[], unsigned int nchildren)
 
   for (i = 0; i < ntoplevel; i++)
   {
-    fprintf(stderr, "[%p]\n", toplevel[i]);
+    fprintf(stderr, "nxagentRootlessRestack: [%p]\n", toplevel[i]);
   }
 
   fprintf(stderr, "nxagentRootlessRestack: Internal top level windows before restack:");
 
   for (pWin = screenInfo.screens[0]->root -> firstChild; pWin != NULL; pWin = pWin -> nextSib)
   {
-    fprintf(stderr, "[%p]\n", pWin);
+    fprintf(stderr, "nxagentRootlessRestack: [%p]\n", pWin);
   }
 
   #endif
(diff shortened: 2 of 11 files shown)


nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -1235,9 +1235,11 @@ static void nxagentParseOptions(char *name, char *value)
     return;
 #else
     if (PANORAMIX_DISABLED_COND && RRXINERAMA_DISABLED_COND)
+    {
       nxagentChangeOption(Xinerama, 0);
       fprintf(stderr, "Warning: XINERAMA extension has been disabled on %s startup.\n", __progname);
       return;
```
Ionic, line 1241: Hum, doesn't that mean that selecting did nothing so far? Good fix.

nx-X11/programs/Xserver/hw/nxagent/Rootless.c
```diff
@@ -334,6 +334,7 @@ void nxagentRootlessRestack(unsigned long children[], unsigned int nchildren)
 
   if (!ntoplevel)
   {
+    free(toplevel);
```
Ionic, line 344: Making a note here, `toplevel` could be ported to use `allocarray()`.

nx-X11/programs/Xserver/hw/nxagent/Rootless.c
```diff
@@ -497,6 +504,7 @@ int nxagentExportProperty(pWin, property, type, format, mode, nUnits, value)
   #ifdef _XSERVER64
   else if (strcmp(typeS, "CARDINAL") == 0 || strcmp(typeS, "WM_SIZE_HINTS") == 0)
   {
+    /* FIXME: is it okay here to ignore malloc fails? */
```
Ionic, line 507: Well given the code is only doing something when `malloc` succeeded... probably okay... If `export` isn't true, the function won't do much later on.

nx-X11/programs/Xserver/hw/nxagent/Rootless.c
```diff
@@ -1067,6 +1091,14 @@ void nxagentImportProperty(Window window,
     WindowPtr pWindow;
     int i;
 
+    if (!wind)
+    {
+      #ifdef WARNING
```
Ionic, line 1096: Maybe convert everything here to use `allocarray` as well?

nx-X11/programs/Xserver/hw/nxagent/Rootless.c
```diff
@@ -351,14 +351,14 @@ void nxagentRootlessRestack(unsigned long children[], unsigned int nchildren)
 
   for (i = 0; i < ntoplevel; i++)
   {
-    fprintf(stderr, "[%p]\n", toplevel[i]);
+    fprintf(stderr, "nxagentRootlessRestack: [%p]\n", toplevel[i]);
```
Ionic, line 354: I'd use `"%s: ...", __func__, ...` in this commit and especially in all new code.

nx-X11/programs/Xserver/hw/nxagent/Window.c
```diff
@@ -2684,8 +2684,6 @@ void nxagentDisconnectWindow(void * p0, XID x1, void * p2)
 
 Bool nxagentReconnectAllWindows(void *p0)
 {
-  int flexibility = *(int *) p0;
```
Ionic: NAK, don't remove that.

Add `(void) flexibility;` instead. Otherwise we will never know what this `p0` argument is supposed to be in the future.

Ionic (changes requested):

uli42:
Well, we could add an according comment...

Vom Smartphone gesendet.

Von: "Mihai Moldovan" <notifications@github.com>
Gesendet: ‎04.‎03.‎2017 09:10
Betreff: Re: [ArcticaProject/nx-libs] Pr/misc cherries (#378)

@Ionic requested changes on this pull request.



In nx-X11/programs/Xserver/hw/nxagent/Window.c:
> @@ -2684,8 +2684,6 @@ void nxagentDisconnectWindow(void * p0, XID x1, void * p2)
 
 Bool nxagentReconnectAllWindows(void *p0)
 {
-  int flexibility = *(int *) p0;
NAK, don't remove that.
Add (void) flexibility; instead. Otherwise we will never know what this p0 argument is supposed to be in the future.

Ionic:
Yeah, but comments are more easily and inadvertently removed. An unused stack variable won't kill us - and the compiler likely optimizes it out anyway.

uli42:
Hmm. What's the state of the flexibility stuff. I thought it is
finished? What is going to come here in future?  Aaannd: shouldn't we
remove p0 until we really need it?

Ionic:
Hm, good point, that's long done. Code can get a value via `nxagentOption(ReconnectTolerance)`, so there's probably no need to even keep this argument around.

Ionic:
I think the original idea in NX code was to provide different values to each function - but that was never actually used.

It wasn't a terrible idea to allow stricter checks for, say, depth and less strict ones for Pixmaps, but given that was never used, we could do away with it altogether and just re-fetch the value on demand if we ever want to make functions more dynamic in their handling.

uli42:
ack

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: I am not sure when exactly that code is active, But it definitely looked wrong ;-)

nx-X11/programs/Xserver/hw/nxagent/Rootless.c
uli42: there are _many_ places all over the code...

Ionic:
How about removing the `void *p0` parameter, then? Currently it would be unused and you're exchanging a compiler warning (`flexibility unused` for `argument unused`.)

sunweaver:
@uli42: @Ionic: let's get this PR in first, and then I revisit #333.

sunweaver:
@uli42: one reviewer on this PR is enough, please finish this up with Ionic. It feels like this should be the next PR to get merged, right?

nx-X11/programs/Xserver/hw/nxagent/Rootless.c
uli42: I don't want to integrate allocarray into this pr.

nx-X11/programs/Xserver/hw/nxagent/Rootless.c
uli42: Well, there are about 100000 locations all over the place where this could be done. But not in this commit.

uli42:
I will rebase this and take out the changes regarding flexibility. Cleaning up there is way more than just removing this variable and p0, so I will open a new issue for that.

uli42:
this is basically ready to merge. But as I originally had thought of this a a collection of cherrypicks the question is if the name of the branch should stay in case of a complete merge.

nx-X11/programs/Xserver/hw/nxagent/Rootless.c
Ionic: Just thought it would make sense if you're changing the lines to include the function anyway.

nx-X11/programs/Xserver/hw/nxagent/Rootless.c
Ionic: I didn't mean in this PR, that was a general comment.

Ionic:
It doesn't really make sense to cherry-pick single commits. There also isn't one specific topic all commits would fit under... if you can find a better name, rename the branch (although I'm not sure if GH supports that), otherwise I guess it's fine the way it is?

Ionic (approved):

PR #401, opened 2017-03-21
NXpicturestr_PictSolidFill.h: Drop file and revert db8705. Not an iss…

sunweaver:
…ue anymore nowadays.

The issue at the time of introducing this ABI diversion from X.org was that people could not see any fonts in GTK applications like libreoffice. Tested with various GTK-2 and GTK-3 application, this workaround is not necessary anymore more, these days.


sunweaver:
@vatral: can you test this patch with your old SUSE systems? Do see any missing glyphs with this PR applied? Thanks!

uli42 (commented):
I'm sorry, I am in no way into this. 

Technically it looks ok , but I cannot evaluate if it is okay from the way Xorg works.

Ionic:
I don't really know either. In my tests, having this workaround made no difference to how glyphs were shown - but I haven't tested stuff like LibreOffice and certainly not on old platforms.

PR #408, opened 2017-03-22
hw/nxagent: Drop support for MCOPGLOBALS atom needed by KDE 3's ancie…

sunweaver:
…nt ArtsD sound server.


sunweaver:
@uli42: Can you take a quick look at #408? The removal (reason) should be self-explanatory.

uli42 (commented):
Isn't that commit "remove NXAGENT_ARTSD support?

However, I am still using KDE3 (well, Trinity), so I'd rather not drop it. The reason I am using it is that konsole is much faster in KDE3 than in KDE4. On the other hand I am not using sound at all...

sunweaver:
Hi Uli,

On  Mi 22 Mär 2017 20:21:14 CET, Ulrich Sibiller wrote:

>
> Isn't that commit "remove NXAGENT_ARTSD support?
>
> However, I am still using KDE3 (well, Trinity), so I'd rather not  
> drop it. The reason I am using it is that konsole is much faster in  
> KDE3 than in KDE4. On the other hand I am not using sound at all...

And Trinity stil uses ArtsD? You are kidding... They should be on  
Pulseaudio meanwhile, too, aren't they???

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

uli42:
As said above I do not use sound for NX. But in the TDE repo there's some reference to artsd: http://mirror.git.trinitydesktop.org/cgit/tdemultimedia/tree/

Also see here: http://www.giustetti.net/wiki/index.php?title=En/TDE_arts

sunweaver:
Let's postpone this until a later point in time. MCOPGLOBALS code does not hurt, and we don't want to stress the Trinity users.

Issue #394, opened 2017-03-12
nxcomp 3.6.x FTBFS on Cygwin


mikedep333:
Using latest 32-bit Cygwin packages as of right now (2017-03-12).
Using latest nx-libs 3.6.x as of right now (dc8477f76372e685eb999a71afb4e5ec7e3176f1)

```
g++ -c  -O3 -fno-rtti -fno-exceptions -I../nx-X11/exports/include -Wmissing-declarations -DIN_ADDR_T=in_addr_t -DVERSION=\"3.5.99.4\"  -DPACKAGE_NAME=\"\" -DPACKAGE_TARNAME=\"\" -DPACKAGE_VERSION=\"\" -DPACKAGE_STRING=\"\" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -Wall -Wpointer-arith   Loop.cpp
Loop.cpp:108:8: error: redefinition of ‘struct sockaddr_un’
 struct sockaddr_un
        ^
In file included from Loop.cpp:47:0:
/usr/include/sys/un.h:18:8: error: previous definition of ‘struct sockaddr_un’
 struct sockaddr_un {
        ^
make: *** [Makefile:100: Loop.o] Error 1
```

[Full Log](http://pastebin.com/Ku31i2cu)

sunweaver:
Hi Mike,

On  So 12 Mär 2017 17:19:21 CET, Mike DePaulo wrote:

> Using latest 32-bit Cygwin packages as of right now (2017-03-12).
> Using latest nx-libs 3.6.x as of right now
>
> `
> g++ -c  -O3 -fno-rtti -fno-exceptions -I../nx-X11/exports/include  
> -Wmissing-declarations -DIN_ADDR_T=in_addr_t -DVERSION=\"3.5.99.4\"   
> -DPACKAGE_NAME=\"\" -DPACKAGE_TARNAME=\"\" -DPACKAGE_VERSION=\"\"  
> -DPACKAGE_STRING=\"\" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\"  
> -Wall -Wpointer-arith   Loop.cpp
> Loop.cpp:108:8: error: redefinition of ‘struct sockaddr_un’
>  struct sockaddr_un
>         ^
> In file included from Loop.cpp:47:0:
> /usr/include/sys/un.h:18:8: error: previous definition of ‘struct  
> sockaddr_un’
[...]

PR #416, opened 2017-03-29
nxcomp: Use Cygwin un.h & remove EMX code

mikedep333:
Cygwin has modified it over the years.
Fixes FTBFS #394 

Note that Loop.cpp now has this line defined outside of the #ifndef 
#include <sys/un.h>


mikedep333:
I am planning on submitting a follow-up PR to clean up the EMX code.
https://en.m.wikipedia.org/wiki/EMX_(programming_environment)

sunweaver (approved):

PR #419, opened 2017-04-07
Enhance -displayfd code path and display number auto detection

### LINKED ISSUE #418: Utilize -displayfd more genuinely in nxagent
More genuinely means:

  - allow :<displ> to be a starting point for scanning for the next available display number
  - if -displayfd X is given and X is stderr, then add some recognition text around the display number being printed to stderr.

sunweaver:
This PR fixes #418. It slightly modifies the behaviour of the new -displayfd cmdline option for nxagent:

  * -displayfd 2 -> this will pipe the display number to stderr and add some beautiful text around it.
  * -displayfd <X> : 50 -> this will let the nxagent auto-detect a free display number, starting at :50

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
+++ b/nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
@@ -180,8 +180,23 @@ causes the server to generate a core dump on fatal errors.
 specifies a file descriptor in the launching process.  Rather than specifying
 a display number, the X server will attempt to listen on successively higher
 display numbers, and upon finding a free one, will write the port number back
-on this file descriptor as a newline-terminated string.  The \-pn option is
-ignored when using \-displayfd.
+on this file descriptor as a newline-terminated string.  The \fB\-pn\fR option is
+ignored when using \fB\-displayfd\fR.
+
+nxagent specific:
+
+(1) Other than in X.org's Xserver, you can use \fB\-displayfd\fR in
+conjunction with an explicit display number. If the explicit display number
+is not available (i.e., already in use), nxagent tries to figure out the next
+available display number,
+
+e.g.:
+
+   \fBnxagent\fR \fI\-displayfd 2 :50\fR
+
+(2) If -displayfd <X> is given with <X> equaling 2 (STDERR), then the
+display number string written to STDERR is beautified with some human-readable
+(machine-parseable) text.
 .TP 8
 .B \-deferglyphs \fIwhichfonts\fP
 specifies the types of fonts for which the server should attempt to use
--- a/nx-X11/programs/Xserver/os/connection.c
+++ b/nx-X11/programs/Xserver/os/connection.c
@@ -427,7 +427,11 @@ CreateWellKnownSockets(void)
     if (NoListenAll) {
 	ListenTransCount = 0;
     }
+#ifndef NXAGENT_SERVER
     else if ((displayfd < 0) || explicit_display) {
+#else
+    else if (displayfd < 0) {
+#endif /* ! NXAGENT_SERVER */
         if (TryCreateSocket(atoi(display), &partial) &&
             ListenTransCount >= 1)
             if (!PartialNetwork && partial)
@@ -435,7 +439,14 @@ CreateWellKnownSockets(void)
     }
     else { /* -displayfd and no explicit display number */
 	Bool found = 0;
+#ifdef NXAGENT_SERVER
+	int i_offset = 0;
+	if (explicit_display)
+	     i_offset = atoi(display);
+	for (i = i_offset; i < 65536 - X_TCP_PORT; i++) {
+#else
 	for (i = 0; i < 65536 - X_TCP_PORT; i++) {
+#endif /* NXAGENT_SERVER */
 	    if (TryCreateSocket(i, &partial) && !partial) {
 		found = 1;
 		break;
--- a/nx-X11/programs/Xserver/os/connection.c
+++ b/nx-X11/programs/Xserver/os/connection.c
@@ -79,6 +79,7 @@ SOFTWARE.
 #include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <unistd.h>
 
 #ifndef WIN32
 #include <sys/socket.h>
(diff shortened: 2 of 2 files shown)


sunweaver:
@Ionic: #419 could be your next review...

nx-X11/programs/Xserver/os/connection.c
```diff
@@ -364,6 +365,13 @@ NotifyParentProcess(void)
 {
 #if !defined(WIN32)
     if (displayfd >= 0) {
+#ifdef NXAGENT_SERVER
+	if (displayfd == STDERR_FILENO)
+	{
+	    if (write(displayfd, "Auto-detected display number is: DISPLAY=:", 42) != 42)
```
Ionic: Hm, make that
```C
	    const char *msg = "Auto-detected display number is: DISPLAY=:";
	    if (write(displayfd, msg, strlen(msg)) != strlen(msg))
```

Since nobody likes counting chars...

(Btw, the whole code block here is bad IMHO, since `write()` might fail for other reasons, for instance with `EINTR`, `EWOULDBLOCK` or `EAGAIN`, which aren't fatal or hard errors. But that's upstream's mess to clean up, if they care to, I guess.)

Ionic:
Uhm... I'm having a hard time understanding what you're trying to do.

Shall `-displayfd :50` explicitly select the display number? Or be used as the starting point for auto-detection? Was `-displayfd : 50` in the original description a typo, or did you really mean that?

There's something wrong in your code, but I can only elaborate once I know what you're after.

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -182,8 +182,23 @@ causes the server to generate a core dump on fatal errors.
...
+nxagent specific:
+
+(1) Other than in X.org's Xserver, you can use \fB\-displayfd\fR in
+conjunction with an explicit display number. If the explicit display number
+is not available (i.e. already in use), nxagent tries to figure out the next
+available display number,
+
+e.g.:
+
+   \fBnxagent\fR \fI\-displayfd 2 :50\fR
```
Ionic, line 195: I don't think you actually mean passing both `2` and `:50` at the same time?

Ionic: Ah, no, that's fine, since `:50` is selecting the specific display number and 2 is `-displayfd`'s parameter.

nx-X11/programs/Xserver/os/connection.c
```diff
@@ -426,15 +426,26 @@ CreateWellKnownSockets(void)
     if (NoListenAll) {
 	ListenTransCount = 0;
     }
+#ifndef NXAGENT_SERVER
     else if ((displayfd < 0) || explicit_display) {
+#else
+    else if ((displayfd < 0) && explicit_display) {
+#endif /* ! NXAGENT_SERVER */
         if (TryCreateSocket(atoi(display), &partial) &&
```
Ionic: The problem I see is that if `-displayfd` is not given and no explicit display has been selected, the code will jump to the else-branch and do display number autodetection, which breaks the previous behavior (that defaulted to trying to open up display `0` once and fail if it couldn't.)

The condition should thus probably rather be something like

```C
    else if (displayfd < 0) {
```

since if we don't have an displayfd, we can't do autodetection (or rather not write out the autodected display number anywhere.)

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -182,8 +182,23 @@ causes the server to generate a core dump on fatal errors.
...
-on this file descriptor as a newline-terminated string.  The \-pn option is
-ignored when using \-displayfd.
+on this file descriptor as a newline-terminated string.  The \fB\-pn\fR option is
+ignored when using \fB\-displayfd\fR.
+
+nxagent specific:
+
+(1) Other than in X.org's Xserver, you can use \fB\-displayfd\fR in
+conjunction with an explicit display number. If the explicit display number
+is not available (i.e. already in use), nxagent tries to figure out the next
```
Ionic: AE mostly prefers a comma after i.e. and e.g.

Ionic (changes requested):

nx-X11/programs/Xserver/os/connection.c
sunweaver: > Since nobody likes counting chars...

Done.

nx-X11/programs/Xserver/os/connection.c
sunweaver: > The problem I see is that if -displayfd is not given and no explicit display has been selected, the 
> code will jump to the else-branch and do display number autodetection, which breaks the previous 
> behavior (that defaulted to trying to open up display 0 once and fail if it couldn't.)

Agreeing on this. Amended locally.

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
sunweaver: > AE mostly prefers a comma after i.e. and e.g.

Got that.

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
sunweaver: > Ah, no, that's fine, since :50 is selecting the specific display number and 2 is -displayfd's parameter.

Exactly. :50 is the offset for starting the free socket search. The "2" is -> stderr.

sunweaver:
@Ionic: I am about to push an amended version of the update branch.

About the functionality:

  nxagent 
  -> try to open display at :0 (doomed to failure in most scenarios, keeping for legacy reasons)

  nxagent :50
  -> open display at :50

  nxagent -displayfd 2
  -> Start detecting an available display, starting with ":0", then ":1", etc. Display number gets written
      with nice text around to stderr (fd #2)

  nxagent -displayfd <n>
  -> Start detecting an available display, starting with ":0", then ":1", etc. Display number gets written
      to a previously provided FD #<n>, only ":X\n" gets written to the FD.

  nxagent -displayfd 2 :50
  -> Start detecting an available display, starting with ":50", then ":51", etc. Display number gets written
      with nice text around to stderr (fd #2)

  etc.

uli42:
> nxagent -displayfd
> -> Start detecting an available display, starting with ":0", then ":1", etc. Display number gets written
> to a previously provided FD #, only ":X\n" gets written to the FD.

Where/when/how is this fd provided?

Ionic:
Good question, as far as I could see `-displayfd` always expects a parameter and won't work without it. Mike probably meant `nxagent -displayfd [integer other than 2]`.

Ionic (approved):

Issue #261, opened 2016-10-30
Remove install-sh  and mkinstalldirs


uli42:
`nx{comp{,ext,shad},proxy}` all contain `install-sh` and `mkinstalldir`. Both should not be part of the repo but come with autotools.

sunweaver:
@uli42: Are you working on this already?

sunweaver:
@uli42: I have assigned myself to this for now. Veto, if you are already on it.

uli42:
No

Vom Smartphone gesendet.

----- Ursprüngliche Nachricht -----
Von: "Mike Gabriel" notifications@github.com
Gesendet: ‎30.‎10.‎2016 18:20
An: "ArcticaProject/nx-libs" nx-libs@noreply.github.com
Cc: "Ulrich Sibiller" uli42@gmx.de; "Mention" mention@noreply.github.com
Betreff: Re: [ArcticaProject/nx-libs] Remove install-sh and mkinstalldirs(#261)

@uli42: Are you working on this already?

uli42:
isn't this obsolete now?

sunweaver:
yes, this issue got fixed by #473, #474, #480 and #498. Thus, closing...

PR #422, opened 2017-04-11
nx{comp*,proxy}: Drop install-sh and mkinstalldirs and symlink to gnu…

sunweaver:
…lib's versions of those script instead before build.

 Fixes ArcticaProject/nx-libs#261


Makefile
```diff
@@ -45,6 +45,8 @@ SHELL:=/bin/bash
 	    rm -f nx-X11/config/cf/nxversion.def; \
 	    rm -f bin/nxagent; \
 	    rm -f bin/nxproxy; \
+	    rm -f nx*/install-sh; \
```
Ionic, line 48: NAK, this PR is bogus.

We should:
  - be fixing `configure.in`:
    - rename it to `configure.ac`
    - fix `AC_INIT` (`AC_INIT(..., m4_esyscmd([tr -d '\n' < ../VERSION]))`
    - add proper automake dependencies to it (`AM_INIT_AUTOMAKE`)
    - correctly m4-quote all `AC_DEFUN()` calls (`AC_DEFUN([var], ...`)
  - use `autoreconf -fvi` as part of roll-tarball.sh to generate the build system for the tarballs (which should automatically install install-sh and mkinstalldirs as symlinks if they don't exist and autoreconf determined that the build system is using automake.)

Ionic (changes requested):

Makefile
Ionic: `mkinstalldirs` is tricky, we should be using `./install-sh -d` instead, since the former is deprecated.

sunweaver:
Will be obsoleted by PR #446, once it is ready...

PR #424, opened 2017-04-11
Various backports for Xserver/mi/miexpose (esp. miPaintWindow rewrite)

sunweaver:


nx-X11/programs/Xserver/Xext/panoramiXprocs.c
```diff
@@ -1117,8 +1117,7 @@ int PanoramiXCopyArea(ClientPtr client)
 		}
 	    }
 	    RegionValidate(&totalReg, &overlap);
-	    (*pScreen->SendGraphicsExpose)(
-		client, &totalReg, stuff->dstDrawable, X_CopyArea, 0);
+	    SendGraphicsExpose(client, &totalReg, stuff->dstDrawable, X_CopyArea, 0);
```
Ionic, line 1120: Any reason why you haven't formatted this as upstream did here and below?

nx-X11/programs/Xserver/include/dix.h
```diff
@@ -470,6 +470,12 @@ extern void ClientWakeup(
 extern Bool ClientIsAsleep(
     ClientPtr /*client*/);
 
+extern void SendGraphicsExpose(ClientPtr /*client */ ,
```
Ionic, line 473: Could also directly use the current style here as well.

nx-X11/programs/Xserver/mi/miexpose.c
```diff
@@ -610,8 +610,8 @@ miPaintWindow(WindowPtr pWin, RegionPtr prgn, int what)
 	draw_x_off = drawable->x;
 	draw_y_off = drawable->y;
 
-	tile_x_off = 0;
-	tile_y_off = 0;
+	tile_x_off = pWin->drawable.x - draw_x_off;
```
Ionic: This backport might be meant well, but won't actually fix anything. With the current code, the behavior will at best be unchanged (since the new expressions should evaluate to zero anyway) and in the worst case break completely due to underflows.

For this to actually work correctly, we'll need a partial or full backport of [e4d11e58ce349dfe6af2f73ff341317f9b39684c](https://cgit.freedesktop.org/xorg/xserver/commit/?id=e4d11e58ce) first - but that one is HUGE!

Ionic (changes requested):

uli42:
see also #440

PR #437, opened 2017-04-19
use source tree's  nxcomp{,shad}

sunweaver:
This PR includes some imake fixes.

  - one that assures that the source tree's nxcomp{,shad} shared libs get preferred to the system-wide installed nxcomp{,shad} shared libs at linking time
  - another that drops a local library path we don't need / want anymore
  - another that fixes a problem with some buggy zlib pkg-config file that got shipped with some zlib versions


nx-X11/programs/Xserver/Imakefile
```diff
@@ -292,7 +292,7 @@ NXAGENTOBJS = hw/nxagent/miinitext.o     \
 XPMLIB    =  -lXpm
 XMLLIB    =  `pkg-config --libs libxml-2.0`
 PIXMANLIB =  `pkg-config --libs pixman-1`
-ZLIB      =  `pkg-config --libs zlib`
+ZLIB      =  `pkg-config --libs-only-l zlib`
```
Ionic: I'm not convinced we really need this - as long as `-lXcomp*` comes before system libraries, which should be the case if I see the ordering in the Imakefile correctly.

`-L` options only have effect for subsequently listed libraries, so removing the multi-arch path should not make any difference for `libXcomp*`?

Ionic: After your linker options order change patch, at least.

I'm not sure if github gets the commit order right, but doesn't it work with the linker options order change first and this commit dropped completely?

nx-X11/programs/Xserver/Imakefile
```diff
@@ -316,7 +316,6 @@ $(NXAGENTOBJS) $(NXAGENTLIBS) $(NXAGENTSYSLIBS):: $(NXAGENTDIRS)
 #if defined(SunArchitecture)
 NXAGENTNXLIBS = -L/usr/sfw/lib \
                 -L../../../nxcomp  \
-                -L../../../nx-X11/exports/lib \
```
Ionic, line 319: `-L../../exports/lib` is being inserted by `LdPreLib`. Took me a while to figure that out.

Ionic (approved):

PR #439, opened 2017-04-19
Drop unused Xserver/miext/cw

sunweaver:
Xserver/miext/cw: Remove. The nxagent Xserver was never a consumer.

 This change lets nxagent segfault in fb/fbsolid.c if the Composite
 extension gets enabled (nxagent +extension Composite :<disp>).

 However, as Composite support is broken anyway, currently, let's
 approach Composite support without the deprecated CompositeWrapper.


Ionic (commented):
Theoretically okay, but do we want to ignore the segfault?

Issue #282, opened 2016-11-07
-Wpedantic does not work on RHEL6


uli42:
`-Wpedantic` is not supported on RHEL6`s gcc, I get 

```
gcc -o makekeys -g -O3 -fno-strict-aliasing -g -std=c99 -Wall -Wpedantic -Wpointer-arith -Wundef    -I../.. -I../../exports/include   -Dlinux -D__amd64__ -D_POSIX_C_SOURCE=199309L                               -D_POSIX_SOURCE -D_XOPEN_SOURCE                                 -D_DEFAULT_SOURCE                                 -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64                               -D_GNU_SOURCE                            -DFUNCPROTO=15 -DNARROWPROTO -DXTHREADS  -D_REENTRANT -DXUSE_MTSAFE_API    -DMALLOC_0_RETURNS_NULL  -DHAS_SNPRINTF -DLIBX11                     -DPOSTLOCALELIBDIR=\"lib\" -I../../../nxcomp -DNX_TRANS_SOCKET              -DNX_TRANS_EXIT -DXCMS -DUSE_XCB=0   util/makekeys.c  -lpthread   -Wl,-rpath-link,../../exports/lib
cc1: Fehler: nicht erkannte Kommandozeilenoption »-Wpedantic«
```

The provided compiler is

```
LANG=C gcc -v
Using built-in specs.
Target: x86_64-redhat-linux
Configured with: ../configure --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-bootstrap --enable-shared --enable-threads=posix --enable-checking=release --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-languages=c,c++,objc,obj-c++,java,fortran,ada --enable-java-awt=gtk --disable-dssi --with-java-home=/usr/lib/jvm/java-1.5.0-gcj-1.5.0.0/jre --enable-libgcj-multifile --enable-java-maintainer-mode --with-ecj-jar=/usr/share/java/eclipse-ecj.jar --disable-libjava-multilib --with-ppl --with-cloog --with-tune=generic --with-arch_32=i686 --build=x86_64-redhat-linux
Thread model: posix
gcc version 4.4.7 20120313 (Red Hat 4.4.7-17) (GCC)
```

[...]

PR #447, opened 2017-04-25
nx-libs.spec: fix FTBFS on RHEL6

uli42:
Fixes: ArcticaProject/nx-libs#282

### DIFF
--- a/nx-libs.spec
+++ b/nx-libs.spec
@@ -357,6 +357,10 @@ find -name LICENSE | xargs sed -i \
   -e 's/MA  02111-1307/MA  02110-1301/'
 # Fix source permissions
 find -type f -name '*.[hc]' | xargs chmod -x
+# Some systems do not know -Wpedantic
+%if ( 0%{?rhel} && 0%{?rhel} < 7 ) || ( 0%{?suse_version} && 0%{?suse_version} < 1310 )
+sed -i -e 's/Wpedantic/pedantic/g' nx-X11/config/cf/{{host,xorgsite}.def,xorg.cf}
+%endif
 
 # Bundled nx-X11/extras
 # Mesa - Used by the X server


sunweaver:
LGTM. @Ionic? Any concerns? Otherwise, I could merge it.

uli42:
ping!

nx-libs.spec
```diff
@@ -1,4 +1,8 @@
 %global _hardened_build 1
+# prevent "ERROR: No build ID note found" errors
+%if 0%{rhel} < 7
```
Ionic: Needs fixup: `%if 0%{?rhel} && 0%{?rhel} < 7`

(For once, missing the `?` char which will cause errors when the macro is not defined, secondly `0%{?rhel} < 7` will evaluate to true if `%{rhel}` is undefined [since `0%{?rhel} < 7` becomes `0 < 7` in that case and this is `true`], which is not what we want.)

nx-libs.spec
```diff
@@ -1,4 +1,8 @@
...
+%undefine _missing_build_ids_terminate_build
```
Ionic: Why exactly does this lead to a failure anyway?

According to [the RHEL6 documentation](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Developer_Guide/compiling-build-id.html) all binaries built on EL6+ should have buildid information (probably through the compiler.)

So, why is that failing for you?

nx-libs.spec
```diff
@@ -357,6 +357,10 @@ find -name LICENSE | xargs sed -i \
   -e 's/MA  02111-1307/MA  02110-1301/'
 # Fix source permissions
 find -type f -name '*.[hc]' | xargs chmod -x
+# RHEL6 and lower do not know -Wpedantic
+%if 0%{rhel} < 7
```
Ionic: Same problem as in the other commit (missing `?` char and the other check.)

While this may fix builds for EL6-, it would be good to also fix *SUSE builds if we're at it.

Looks like `-Wpedantic` is supported from GCC 4.8 onwards, and...

SLE 12 has GCC 4.8, SLE 11 has 4.3.

OpenSUSE 13.* has GCC 4.8, so that's fine. 12.* only has GCC 4.7, too old.

So how about:

```spec
%if ( 0%{?rhel} && 0%{?rhel} < 7 ) || ( 0%{?suse_version} && 0%{?suse_version} < 1310 )
```

?

Ionic (changes requested):

nx-libs.spec
uli42: you are completely right, I had thought of that later, too.

nx-libs.spec
uli42: We have a system with different compiler in the path and I had forgotten to update the path prior to compiling. Mock would have helped here....

nx-libs.spec
uli42: Ok, I have reworked the patch. It is building fine once I remove the other gcc from the PATH... Please have a second glance.

uli42:
Ok, I have reworked the patch. It is building fine once I remove the other gcc from the PATH... Please have a second glance.

Ionic (approved):
Okay, but go ahead and squash the changes since it's really the same semantic unit.

uli42:
Updated!

Ionic:
Please rebase against 3.6.x.

uli42:
rebased

PR #453, opened 2017-05-05
Fix FTBFS on Linux based HP-PARisc platforms

sunweaver:


uli42 (dismissed):
Only changing the path in the patch file should be removed from this patch

sunweaver:
What path did I change? The debian/patches/016 patch required rebasing due to line number changes. Anything, I am not really spotting here?

uli42:
You changed

```
a/nx-X11/config/cf/site.def
```
to
```
nx-libs.orig/nx-X11/config/cf/site.def
```

and similar at several places. This produces unneccessary lines in the patch.

sunweaver:
ahhh... I see. Right. Done.

sunweaver:
note... Helge Deller, a PA-Risc expert from the Debian context noted in a private comm that hppa is BIG ENDIAN, not LITTLE ENDIAN, so this got fixed on this branch.

nx-X11/programs/Xserver/include/servermd.h
```diff
@@ -48,308 +48,19 @@ SOFTWARE.
...
-# else
-#  define BITMAP_BIT_ORDER      LSBFirst
-# endif
-#endif
-
-#ifndef BITMAP_SCANLINE_UNIT
-# if defined(XF86MONOVGA) || defined(XF86VGA16)
-#  define BITMAP_SCANLINE_UNIT  8
-# endif
 #endif
```
Ionic: Missing else-catch with `#error "Too weird to live."`.

nx-X11/programs/Xserver/include/servermd.h
```diff
@@ -48,308 +48,19 @@ SOFTWARE.
...
-#if	(defined(SVR4) && defined(i386)) || \
-	defined(__alpha__) || defined(__alpha) || \
-	defined(__i386__) || defined(__i386) || \
-	defined(__s390x__) || defined(__s390__)
-  
-#ifndef IMAGE_BYTE_ORDER
-#define IMAGE_BYTE_ORDER	LSBFirst
-#endif
-
-#ifndef BITMAP_BIT_ORDER
```
Ionic, line 314: These removals (including `BITMAP_SCANLINE_UNIT`) are really part of [f6469f1910c72add358b7263d69bd116b4450ece](https://cgit.freedesktop.org/xorg/xserver/commit/?id=f6469f1910c72add358b7263d69bd116b4450ece).

More `XF86MONOVGA` and friends sections have been previously removed in [446fe9eecddd1337f9d5164dd7c301e1ba3dfe32](https://cgit.freedesktop.org/xorg/xserver/commit/?id=446fe9eecddd1337f9d5164dd7c301e1ba3dfe32) - including the `vax` one (which had a non-standard `GLYPHADDSIZE`.)

The other non-standard `GLYPHADDSIZE` section for old IBM systems has been removed in [5aaacb501ff26bad391664cc7367bfe5164f72c7](https://cgit.freedesktop.org/xorg/xserver/commit/?id=5aaacb501ff26bad391664cc7367bfe5164f72c7).

nx-X11/config/cf/Imake.cf
```diff
@@ -841,6 +841,12 @@ XCOMM Keep cpp from replacing path elements containing i486/i586/i686
 #  define Mc68020Architecture
```
Ionic: Doesn't this revert f6dcf2feb8e0f48811852de7f4148cbf7336db08 partially? It doesn't bring back support for HP-UX, but pulls in back some parts for HPPA. It also seems to be missing some hunks.

nx-X11/config/cf/linux.cf
```diff
@@ -677,6 +677,15 @@ XCOMM binutils:	(LinuxBinUtilsMajorVersion)
 # endif
 #endif /* AlphaArchitecture */
```
Ionic, line 691: Non-ELF and ELF definitions for `HPArchitecture` are missing in this file, but I guess this is okay since the defaults match more or less what would have been defined anyway. `AsmDefines` might be a problem, though...

Ionic (changes requested):

nx-X11/config/cf/Imake.cf
sunweaver: > Doesn't this revert f6dcf2f partially? It doesn't bring back support for HP-UX, but pulls in back some parts for HPPA. It also seems to be missing some hunks.

Yes, it actually does. Good point. I have now picked the __hppa__ related stuff back in while leaving the hpux stuff out. The commit now brings back the old code, so things should look similar as compared to the earlier state.

nx-X11/config/cf/linux.cf
sunweaver: > Non-ELF and ELF definitions for HPArchitecture are missing...

Yes, they are back now. Also the AsmDefines are as "before" when __hppa__ support was still present.

nx-X11/programs/Xserver/include/servermd.h
sunweaver: > Missing else-catch with #error "Too weird to live.".

Got that now. Superb!

sunweaver:
@Ionic: ready to go with the next iteration on #453.

Ionic:
Are you sure you pushed the correct changes? I have a feeling that something is missing, but I'm currently investigating the PR again.

nx-X11/config/cf/Imake.cf
```diff
@@ -887,6 +887,12 @@ XCOMM Keep cpp from replacing path elements containing i486/i586/i686
...
 # endif
+# ifdef __hppa__
+#  ifndef HPArchitecture
+#   define HPArchitecture
+#  endif
+#  undef __hppa__
+# endif /* __hppa__ */
 # ifdef __sh__
 #  ifndef SuperHArchitecture
 #   define SuperHArchitecture
```
Ionic, line 907: `__hppa__` section still missing in the `__OpenBSD__` section?

nx-X11/config/cf/linux.cf
```diff
@@ -537,6 +537,13 @@ XCOMM binutils:	(LinuxBinUtilsMajorVersion)
 #   define LdCmd		CcCmd -nostdlib -Wl,"-m alpha"
 #   define AsmDefines		-DUSE_GAS -U__ELF__
 #  endif /* AlphaArchitecture */
+#  ifdef HPArchitecture
+#   define CcCmd		gcc
+#   define CplusplusCmd		g++
+#   define AsCmd		CcCmd -c -x assembler
+#   define LdCmd		CcCmd -nostdlib
+#   define AsmDefines		-D__ELF__
```
Ionic: Shouldn't this be `-DUSE_GAS -U__ELF__`?

nx-X11/config/cf/linux.cf
```diff
@@ -575,6 +582,20 @@ XCOMM binutils:	(LinuxBinUtilsMajorVersion)
...
+#  ifdef HPArchitecture
+#   define CcCmd		gcc
+#   define CplusplusCmd		g++
+#   define AsCmd		CcCmd -c -x assembler
+#   define LdCmd		CcCmd -nostdlib
+#   define AsmDefines		-DUSE_GAS -U__ELF__
+#  endif /* HPArchitecture */
+#  ifdef HPArchitecture
+#   define CcCmd		gcc
+#   define CplusplusCmd		g++
```
Ionic: The first section here is bogus (should be part of the first hunk, really, given it includes `-DUSE_GAS -U__ELF__`), the second one actually belongs here.

nx-X11/programs/Xserver/include/servermd.h
Ionic: What about this one? I guess we should mention these commits as well, even though sometimes changes were just cherry-picked (contrary to fully applied.)

nx-X11/config/cf/linux.cf
sunweaver: > Shouldn't this be -DUSE_GAS -U__ELF__?

Right.

nx-X11/config/cf/Imake.cf
sunweaver: > __hppa__ section still missing in the __OpenBSD__ section?

Ack. I was focussing on linux only, but alas... Let's get that in again, too.

nx-X11/config/cf/linux.cf
sunweaver: > The first section here is bogus (should be part of the first hunk, really, given it includes -DUSE_GAS -U__ELF__), the second one actually belongs here.

Yes, right, too.

nx-X11/programs/Xserver/include/servermd.h
sunweaver: > What about this one? I guess we should mention these commits as well, even though sometimes changes were just cherry-picked (contrary to fully applied.)

Done. Other X.org commits now mentioned.

sunweaver:
@Ionic: next iteration... Thanks!

Ionic (approved):

PR #454, opened 2017-05-07
Fix FTBFS on mips64el architecture due to missing definition of the _…

sunweaver:
…XSERVER64 macro.

### DIFF
--- a/nx-X11/config/cf/linux.cf
+++ b/nx-X11/config/cf/linux.cf
@@ -790,10 +790,8 @@ XCOMM binutils:	(LinuxBinUtilsMajorVersion)
 # define ServerOSDefines	XFree86ServerOSDefines -DDDXTIME
 # define AsVISOption		-Av9a
 # ifdef Sparc64Architecture
-#  define AsOutputArchSize	64
 #  define ServerExtraDefines	-DGCCUSESGAS XFree86ServerDefines -D_XSERVER64
 # else
-#  define AsOutputArchSize	32
 #  define ServerExtraDefines	-DGCCUSESGAS XFree86ServerDefines
 # endif
 #endif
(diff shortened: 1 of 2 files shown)


sunweaver:
@Ionic: please take a look at this PR (#454). Tested against Debian experimental. See https://buildd.debian.org/status/package.php?p=nx-libs&suite=experimental#problem-1

nx-X11/config/cf/linux.cf
```diff
@@ -733,6 +733,21 @@ XCOMM binutils:	(LinuxBinUtilsMajorVersion)
 # define ServerExtraDefines	-DGCCUSESGAS XFree86ServerDefines
 #endif
 
+#ifdef MipselArchitecture
+# ifndef OptimizedCDebugFlags
+#  define OptimizedCDebugFlags	DefaultGcc2MipsOpt
+# endif
+# define LinuxMachineDefines	-D__MIPSEL__
+# define ServerOSDefines	XFree86ServerOSDefines -DDDXTIME
+# ifdef Mips64elArchitecture
```
Ionic, line 741: Did you take the `AsOutputArchSize` stuff from the `SparcArchitecture` block? It looks like we haven't ever used `AsOutputArchSize` in the first place, so may as well not include it in the Mipsel block...

Ionic: You could have probably merged this into the `Mips` block and checked for `Mips(64)elArchitecture` when needed within it. Would have made some stuff more compact, but maybe also more difficult to read. Your call.

sunweaver: > Did you take the AsOutputArchSize stuff from the SparcArchitecture block? It looks like we haven't ever used AsOutputArchSize in the first place, so may as well not include it in the Mipsel block...

You are right dropping AsOutputArchSize here, as it seems unused elsewhere. Also removing it from the Sparc part, as that was indeed the place I imitated.

sunweaver: > You could have probably merged this into the Mips block and checked for Mips(64)elArchitecture when needed within it. Would have made some stuff more compact, but maybe also more difficult to read. Your call.

Leaving it as is.

sunweaver:
@Ionic: #454 is ready for merging.

Issue #459, opened 2017-05-24
FTBFS on Fedora 25


mikedep333:
150cbc2 (the current version on the 3.6.x branch) is failing to build on Fedora 25.

I am using our upstream .spec file in Mock.

See this Mock [build.log](https://pastebin.com/agj5jFgT)

mikedep333:
Similar errors in the [build.log](https://pastebin.com/3PiEyAxj) on epel7.

mikedep333:
Similar errors in the [build.log](https://pastebin.com/1RjZqN5a) on epel6.

mikedep333:
I just found this line:
./mesa-quilt: line 32: quilt: command not found

I will submit a PR to add quilt to the spec file. EPEL, Fedora, and SUSE all have it.

uli42:
There's an open issue regarding this. I don't want epel requirements because I faced dependecy problems several times. I would prefer dropping quilt. This also leads to problems when switching branches...

Uli

Vom Smartphone gesendet.

----- Ursprüngliche Nachricht -----
Von: "Mike DePaulo" <notifications@github.com>
Gesendet: ‎25.‎05.‎2017 04:18
An: "ArcticaProject/nx-libs" <nx-libs@noreply.github.com>
Cc: "Subscribed" <subscribed@noreply.github.com>
Betreff: Re: [ArcticaProject/nx-libs] FTBFS on Fedora 25 (#459)

I just found this line:
./mesa-quilt: line 32: quilt: command not found
I will submit a patch to add quilt to the spec file. EPEL, Fedora, and SUSE all have it.

uli42:
See issue #444

sunweaver:
uli42: we won't drop quilt because it makes the nx-specific patch management very easy during  development. We simply need a different approach for applying patches for Mesa, if quilt is not available.

uli42:
On Wed, May 31, 2017 at 1:13 PM, Mike Gabriel <notifications@github.com> wrote:
>
> uli42: we won't drop quilt because it makes the nx-specific patch management very easy during development. We simply need a different approach for applying patches for Mesa, if quilt is not available.


What patches do you mean? I am only aware of patches in the debian dir
and the mesa dir. We can use quilt for debian, I think it will be easy
to add a dependency there. But for Redhat/Cetnos/Fedora/Suse quilt is
not in the distribution and should therefore not be required.

sunweaver:
[...]

PR #460, opened 2017-05-25
nx-libs.spec fix FTBFS by adding quilt for mesa patches (Fixes: #459)

mikedep333:


nx-libs.spec
```diff
@@ -19,6 +19,7 @@ BuildRequires:  gcc-c++
 BuildRequires:  libjpeg-devel
 BuildRequires:  pkgconfig
 BuildRequires:  imake
+BuildRequires:  quilt
```
sunweaver, line 22: Adding quilt as BR for all RPM-based packaging approaches will not work. RHEL6 and RHEL7 do not have it, for example. Similar issue with openSUSE Leap 42.2

So this patch should come with some if guards, e.g. if 0%{?fedora_version} ... %endif. What do you think?

sunweaver (changes requested):

nx-libs.spec
woglinde: Hi,

quilt is nice for creating and managing patches. but in the end there are patches. So why not apply the patches on rh and suse with the patch command directly. This will cause only some overhead.

uli42:
+1

sunweaver:
@mikedep333: ping

Ionic:
Will be obsoleted by ArcticaProject/nx-libs#470, closing up.

Issue #444, opened 2017-04-25
RPM-Building on Centos7 fails


uli42:
`rpmbuild -ta nx-libs-3.5.99-6.tar.gz` fails with mesa problems because the quilt call is missing...

```
./mesa-quilt push -a
./mesa-quilt: line 32: quilt: command not found
```

RHEL7/CentOS7 does not provide `quilt` out of the box, you need to take the one from EPEL. We had lots of problems at our customers when running EPEL software (mainly because of unmaintained dependencies), so it is generally not a good idea to depend on such packages. As I am experiencing other inconveniences with this mesa-quilt script when doing development I think we should find another approach for this.

sunweaver:
The only approach I see then his having a fallback in "mesa-quilt" that simply uses the patch command.

uli42:
Yea, something like that. As there's no huge patch collection
currently we could drop quilt completely. Not sure if the number will
grow large, though...

uli42:
Also, the quilt dependency is missing in the spec file.

uli42:
Maybe git-series could be used: https://github.com/git-series/git-series

> git-series tracks both a patch series and its evolution within the same git repository. git-series works entirely with existing git features, allowing git to push and pull a series to any git repository along with other branches and tags.

sunweaver:
From closed/duplicate bug #89... We need to make sure that such keystrokes also work in shadowed sessions.

From X2Go BTS:
http://bugs.x2go.org/cgi-bin/bugreport.cgi?bug=846
package: nxagent

Currently, there doesn't seem to be a way to cleanly terminate a
shadow session running in fullscreen mode.

Ctrl-Alt-T doesn't work, Alt-Tab gets sent to the remote end, so one
can only run x2goterminatesession on the remote end or kill the
x2goclient task on the local side.

PR #467, opened 2017-06-14
Use Mesa 6.4.1 from original source via git-subtree squash

frantracer:
This pull request contains a mechanism to include any Mesa version using git-subtree command and a symbolic link pointing to the desired project and patches that shall be used.

This pull request replaces #340 due to now the whole list of commits are squashed.


uli42:
Please see #444 regarding the quilt build requirement.

Makefile
```diff
@@ -81,7 +81,6 @@ build-full:
 	cd nxcompshad && autoconf && (${CONFIGURE}) && ${MAKE}
 
 	./mesa-quilt push -a
-
```
sunweaver: Avoid white space change here.

mesa-quilt
```diff
@@ -17,18 +17,18 @@
...
     export QUILT_REFRESH_ARGS="-p ab --no-timestamps --no-index"
     export QUILT_COLORS="diff_hdr=1;32:diff_add=1;34:diff_rem=1;31:diff_hunk=1;33:diff_ctx=35:diff_cctx=33"
-    if ! [ -d "$d/nx-X11/extras/Mesa.patches" ]; then mkdir "$d/nx-X11/extras/Mesa.patches"; fi
-    cd "$d/nx-X11/extras/Mesa/"
+    if ! [ -d $d/nx-X11/extras/Mesa.patches ]; then mkdir $d/nx-X11/extras/Mesa.patches; fi
+    cd $d/nx-X11/extras/Mesa/
 fi
 
-quilt "$@"
+quilt $@
```
sunweaver: Don't change the mesa-quilt script at all. The version you had on my old branch is outdated, the one in 3.6.x is better.

nx-X11/extras/Mesa.patches/series
```diff
@@ -1,2 +1,4 @@
 4001_Arctica_Fix-non-working-GLX-in-64bit-Xorg-7.0.patch
```
sunweaver: Drop the "Arctica" part from this patch name. Does not make sense, really. (I do stupid stuff sometimes, no need to copy me).

roll-tarballs.sh
```diff
@@ -111,7 +111,19 @@ if [ "x$MODE" = "xfull" ]; then
...
+    rm -f  "nx-X11/extras/Mesa/include/GL/internal/"{dri_interface.h,sarea.h}
+    rm -Rf "nx-X11/extras/Mesa/src/"{egl/,glu/,glut/,glw/,glx/mini/,Makefile,glx/Makefile,descrip.mms}
+    rm -Rf "nx-X11/extras/Mesa/src/mesa/"{ppc,sparc,tnl_dd,x86,x86-64,sources}
+    rm -Rf "nx-X11/extras/Mesa/src/mesa/drivers/"{allegro,beos,d3d,directfb,dos,fbdev,ggi,glide,osmesa,svga,windows}
+    rm -Rf "nx-X11/extras/Mesa/src/mesa/drivers/dri/"{ffb,i810,i915,mga,r200,radeon,savage,tdfx,unichrome,fb,gamma,i830,mach64,r128,r300,s3v,sis,trident}
+    rm -Rf "nx-X11/extras/Mesa/src/mesa/drivers/dri/common/"{depthtmp.h,drirenderbuffer.h,dri_util.h,memops.h,mmx.h,spantmp_common.h,stenciltmp.h,texmem.h,utils.h,vblank.h,xmlconfig.h,xmlpool.h,drirenderbuffer.c,dri_util.c,extension_helper.h,mmio.h,spantmp2.h,spantmp.h,texmem.c,utils.c,vblank.c,xmlconfig.c,xmlpool/}
+
+    find nx-X11/extras/Mesa/ -name Makefile | while read file; do rm "$file"; done
+    find nx-X11/extras/Mesa/ -name Makefile.* | while read file; do rm "$file"; done
+    find nx-X11/extras/Mesa/ -name descrip.mms | while read file; do rm "$file"; done
```
sunweaver, line 126: Just a comment: It would be good to test if the tarball stil builds after code got stripped by the above tarballing script enhancement. I can do that after the PR merged and improbe roll-tarball.sh if needed.

nx-libs.spec
```diff
@@ -20,6 +20,9 @@ BuildRequires:  libjpeg-devel
 BuildRequires:  pkgconfig
 BuildRequires:  imake
 
+# for patching Mesa with NXv3-specific patches
+BuildRequires: quilt
+
```
sunweaver: Drop this for now, as it needs deeper investigation. See #444.

sunweaver (changes requested):
Ok, some fine-tuning needed. Please work along all given requests. Thanks!

sunweaver:
@frantracer: also make sure you build everything on top of the latest 3.6.x branch. You local copy seems to be off by two PR  merges.

roll-tarballs.sh
frantracer: I launched the tarball script and when trying to compile the code it is not working. It seems that it removes some code from the Mesa/include/GL directory that is required for the compilation.

frantracer:
I have already fixed those minor tweaks.

UPDATE: Every comment is answered and the code is uploaded.

frantracer:
I have already fixed those minor tweaks.

Makefile
frantracer: > Avoid white space change here.

OK. It was removed in the latest commit, to be equal to master.

mesa-quilt
frantracer: > Don't change the mesa-quilt script at all. The version you had on my old branch is outdated, the one in 3.6.x is better.

OK. It was modified in the latest commit, to be equal to master.

nx-X11/extras/Mesa.patches/series
frantracer: > Drop the "Arctica" part from this patch name. Does not make sense, really. (I do stupid stuff sometimes, no need to copy me).

OK. It was modified in the latest commit.

nx-libs.spec
frantracer: > Drop this for now, as it needs deeper investigation. See #444.

OK. The commit where this only change was introduced is removed.

sunweaver:
@frantracer: this PR branch is still not optimal. You fixed every request, but via a new commit.

Ideally, you clean up the commits already there instead of commiting something with a flaw and then fixing the flaw with a later commit. We want a Git history that is as clean as possible.

Please optimize once more.

frantracer:
@sunweaver I have already squashed these commits with your changes.

sunweaver (approved):
Looks good to me now. Merging in an instant.

PR #470, opened 2017-06-27
mesa-quilt without quilt

sunweaver:
This PR works around various RPM based distro that lack the quilt utility in their build environments. In Debain this is not a problem, as quilt is one of the standard packaging tools.

Obviously, Fedora has quilt, too. For RHEL and SUSE, the mesa-quilt script now has a half-dumb fallback mode in case quilt is not available in $PATH.

### DIFF
--- a/mesa-quilt
+++ b/mesa-quilt
@@ -18,17 +18,84 @@
 # 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
 
 d=. ; while [ ! -d "$d/nx-X11/extras" ] && [ "$(readlink -e "$d")" != "/" ]; do d="$d/.."; done
-if [ -d "$d/nx-X11/extras" ] && [ -z "$QUILT_PATCHES" ]; then
-    # if in nx-libs tree with unset $QUILT_PATCHES
-    export QUILT_PATCHES="../Mesa.patches"
-    export QUILT_PATCH_OPTS="--reject-format=unified"
-    export QUILT_DIFF_ARGS="-p ab --no-timestamps --no-index --color=auto"
-    export QUILT_REFRESH_ARGS="-p ab --no-timestamps --no-index"
-    export QUILT_COLORS="diff_hdr=1;32:diff_add=1;34:diff_rem=1;31:diff_hunk=1;33:diff_ctx=35:diff_cctx=33"
-    if ! [ -d "$d/nx-X11/extras/Mesa.patches" ]; then mkdir "$d/nx-X11/extras/Mesa.patches"; fi
-    cd "$d/nx-X11/extras/Mesa/"
+
+# only continue, if in nx-libs tree
+if [ -d "$d/nx-X11/extras" ]; then
+
+	if which quilt 1>/dev/null; then
+
+		echo "$(basename $0): quilt utility found, so we use it..."
+		echo
+
+		export QUILT_PATCHES="../Mesa.patches"
+		export QUILT_PATCH_OPTS="--reject-format=unified"
+		export QUILT_DIFF_ARGS="-p ab --no-timestamps --no-index --color=auto"
+		export QUILT_REFRESH_ARGS="-p ab --no-timestamps --no-index"
+		export QUILT_COLORS="diff_hdr=1;32:diff_add=1;34:diff_rem=1;31:diff_hunk=1;33:diff_ctx=35:diff_cctx=33"
+		mkdir -p "$d/nx-X11/extras/Mesa.patches"
+
+		cd "$d/nx-X11/extras/Mesa/"
+
+		quilt "$@"
+
+		cd - 1> /dev/null
+
+
+	else
+
+		echo "$(basename $0): no quilt utility found, falling back to stupidly (un)applying our patches..."
+		echo
+
+		CMDLINE_ARGS="$@"
+
+		# we are really stupid here, we only understand 'push -a' and 'pop -a'
+		# this should be sufficient for the build process...
+
+		cd "$d/nx-X11/extras/Mesa/"
+
+		# make sure all need dirs and files exist
+		mkdir -p ./Mesa.patches/
+		touch "../Mesa.patches/series"
+
+		# this file will be needed temporarily
+		touch "../Mesa.patches/.applied-patches"
+
+		if [ "$CMDLINE_ARGS" = "push -a" ]; then
+
+			grep -E -v "^#.*" < "../Mesa.patches/series" | while read patchfile; do
+				if [ -n "$patchfile" ] && [ -r "../Mesa.patches/$patchfile" ] && ! grep -q "$patchfile" < "../Mesa.patches/.applied-patches"; then
+					echo "Applying patch ../Mesa.patches/$patchfile"
+					patch -p1 < "../Mesa.patches/$patchfile"
+					echo "$patchfile" >> "../Mesa.patches/.applied-patches"
+					echo
+				fi
+			done
+
+		elif [ "$CMDLINE_ARGS" = "pop -a" ]; then
+
+			tac "../Mesa.patches/.applied-patches" | while read patchfile; do
+				if [ -n "$patchfile" ] && [ -r "../Mesa.patches/$patchfile" ]; then
+					echo "Removing patch ../Mesa.patches/$patchfile"
+					patch -R -p1 < "../Mesa.patches/$patchfile"
+					sed -e "/$patchfile/d" "../Mesa.patches/.applied-patches" > "../Mesa.patches/.applied-patches.tmp"
+					mv "../Mesa.patches/.applied-patches.tmp" "../Mesa.patches/.applied-patches"
+					echo
+				fi
+			done
+
+		else
+
+			echo "$(basename $0): Doing nothing! Cmdline options not understood: $CMDLINE_ARGS"
+
+		fi
+
+		# remove the .applied-patches file, if empty again
+		if [ ! -s "../Mesa.patches/.applied-patches" ]; then
+			rm -f "../Mesa.patches/.applied-patches"
+		fi
+
+		cd - 1> /dev/null
+
+	fi
+
 fi
-
-quilt "$@"
-
-cd - 1> /dev/null
--- a/nx-libs.spec
+++ b/nx-libs.spec
@@ -20,6 +20,19 @@ BuildRequires:  libjpeg-devel
 BuildRequires:  pkgconfig
 BuildRequires:  imake
 
+# ideally we build with quilt (for mesa-quilt patch appliance script),
+# it seems Fedora has it...
+%if 0%{?fedora}
+BuildRequires:  quilt
+%endif
+
+# other distros sometimes do have quilt, sometimes don't, let's
+# not differentiate here when it is available and when not. Rather
+# rely on stupid patch application fallback mode in mesa-quilt...
+%if 0%{?rhel} || 0%{?suse_version}
+BuildRequires:  patch
+%endif
+
 # suse_version 1315 is SLE-12
 %if 0%{?suse_version} != 1315 && 0%{?suse_version} >= 1230
 BuildRequires:  gpg-offline
--- a/nx-X11/extras/README.Mesa.patches
+++ b/nx-X11/extras/README.Mesa.patches
@@ -1,0 +1,31 @@
+Patch naming scheme for Mesa.patches:
+-------------------------------------
+
+For patch file names, use characters from this ASCII subset:
+[a-zA-Z0-9_-.].
+
+This precaution is esp. important, if you build nx-libs without having
+the quilt utility available in the build environment. This is mostly true
+for non-Debian-based systems.
+
+Patch numbering scheme:
+-----------------------
+
+The patch naming scheme is about ordering patches by names and
+categories. All patches shall be prefixed with a number:
+
+0xxx: Mesa upstream backports
+1xxx: Patches that might be interesting for Mesa upstream
+2xxx: Distro specific patches (not applicable here, probably)
+3xxx: Security / CVE fixes for Mesa (probably obtained from upstream)
+4xxx: Patches that are needed to make Mesa work/build with/against nx-libs
+5xxx: Patches that silence compiler warnings
+
+The "series" file within the Mesa.patches directory defines the order of
+patch application.
+
+The patch names in the "series" file do not necessarily need to be in
+alpha-numerical order. Patch application order is about code and
+functionality, That is, patches in one category are allowed to depend
+upon patches in another category, even if the subsequent digits are not
+strictly ascending.


sunweaver:
@Ionic: please review this PR as paid work.

@uli42: @mikedep333: please try test builds in your various build envs. Thanks!

sunweaver:
@uli42: this is not working yet... As you know I use a tiling window manager (i3). I can easily escape the input lock if auto-grab is off, for example.

Also, I wonder, if we shouldn't rather introduce "flags" as opposed to the "autograb on|off" and "input lock on|off" strings in the title. E.g. "nxagent [GIX]" (grab, input lock, xinerama on) vs. "nxagent [giX]" (grab and input off, Xinerama on). That makes the agent not seem soooo English.

Furthermore, the autograb mode standalone seems to work. So I wonder, if we should separate those two features for now and get the autograb code in as is.

The input lock, I can e.g. escape from when I disable the autograb mode and then manipulate the window (e.g. minize or some fancy tiling command in i3). The window moves by keyboard event to "somewhere" else, the mouse is left outside nxagent and nxagent still shows input lock. Not true.

So for input lock we maybe need to capture mouse-in and mouse-out events (seen from nxagent's perspective). If the user manages to escape from the nxagent windows, then a new mouse-in event should trigger the input lock once more and capture the pointer.

I can take a closer look tomorrow, but maybe the above gives you a clue and things are fixed by tomorrow morning. Hehehe...

uli42:
The branch name is wrong ;-)

sunweaver:
On  Mi 28 Jun 2017 10:27:08 CEST, Ulrich Sibiller wrote:

> The branch name is wrong ;-)

Yeah, I noticed too late. Let's ignore that...

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

mesa-quilt
```diff
@@ -18,17 +18,73 @@
...
 
-quilt "$@"
+if [ -d "$d/nx-X11/extras" ]; then
+
+	if which quilt 1>/dev/null; then
+
+		echo "$(basename $0): quilt utility found, so we use it..."
+		echo
+
+		# if in nx-libs tree with unset $QUILT_PATCHES
```
Ionic: I figure you want to do this (i.e., exporting the variables and moving into the other directory) only after checking for `[ -z "$QUILT_PATCHES" ]`?

Ionic: Although, strictly speaking, the old implementation was buggy in the sense that it only moved into the directory of `$QUILT_PATCHES` was unset, but always unconditionally moved to the previous directory in its stack, so we shouldn't blindly put that back.

mesa-quilt
```diff
@@ -18,17 +18,73 @@
...
+
+
+	else
+
+		echo "$(basename $0): no quilt utility found, falling back to stupid patch appliance..."
+		echo
+
+		CMDLINE_ARGS="$@"
+
+		# we are really stupid here, we only under stand 'push -a' and 'pop -a'
```
Ionic: `under stand`?

mesa-quilt
```diff
@@ -18,17 +18,73 @@
...
+		# this should be sufficient for the build process...
 
-cd - 1> /dev/null
+		cd "$d/nx-X11/extras/Mesa/"
+
+		touch "../Mesa.patches/.applied-patches"
+
+		if [ "$CMDLINE_ARGS" = "push -a" ]; then
+
+			cat "../Mesa.patches/series" | grep -E -v "^#.*" | while read patchfile; do
```
Ionic: `grep -E -v "^#.*" < "../Mesa.patches/series" ...`

mesa-quilt
```diff
@@ -18,17 +18,73 @@
...
+				if [ -n "$patchfile" ] && [ -r "../Mesa.patches/$patchfile" ] && ! cat ../Mesa.patches/.applied-patches | grep -q $patchfile; then
```
Ionic: `... ! grep -q "$patchfile" < "../Mesa.patches/.applied-patches"`

mesa-quilt
```diff
@@ -18,17 +18,73 @@
...
+					echo "Applying patch ../Mesa.patches/$patchfile"
+					patch -p1 < "../Mesa.patches/$patchfile"
+					echo $patchfile >> ../Mesa.patches/.applied-patches
```
Ionic: `echo "$patchfile" >> "../Mesa.patches/.applied-patches"`

mesa-quilt
```diff
@@ -18,17 +18,73 @@
...
+				fi
+			done
+
+		elif [ "$CMDLINE_ARGS" = "pop -a" ]; then
+
+			tac "../Mesa.patches/.applied-patches" | while read patchfile; do
+				if [ -n "$patchfile" ] && [ -r "../Mesa.patches/$patchfile" ]; then
+					echo "Removing patch ../Mesa.patches/$patchfile"
+					patch -R -p1 < "../Mesa.patches/$patchfile"
+					sed -i "/$patchfile/d" "../Mesa.patches/.applied-patches"
```
Ionic: `sed -i` is not portable. :(

```bash
sed -e "/$patchfile/d" "../Mesa.patches/.applied-patches" > "../Mesa.patches/.applied-patches.tmp"
mv "../Mesa.patches/.applied-patches.tmp" "../Mesa.patches/.applied-patches"
```

nx-libs.spec
```diff
@@ -20,6 +20,19 @@ BuildRequires:  libjpeg-devel
...
 
+# ideally we build with quilt (for mesa-quilt patch appliance script),
+# it seems Fedora has it...
+%if 0%{?fedora_version}
+BuildRequires:  quilt
+%endif
+
+# other distros sometimes do have quilt, sometimes don't, let's
+# not differentiate here when it is available and when not. Rather
+# rely on stupid patch appliance fallback mode in mesa-quilt...
```
Ionic: `application`?

Ionic (changes requested):

mesa-quilt
uli42: Why quote the path here? It's a static string

mesa-quilt
```diff
@@ -18,17 +18,73 @@
 # 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
 
 d=. ; while [ ! -d "$d/nx-X11/extras" ] && [ "$(readlink -e "$d")" != "/" ]; do d="$d/.."; done
```
uli42, line 20: Can you explain why we need this? We have two cases: 
1. we are in the base directory. Then the path to mesa is clear
2. we are not in the base directory. Then we can use $(dirname $0) to get to defined place

I don't see why climbing up until / gets hit makes any sense...

mesa-quilt
```diff
@@ -18,17 +18,73 @@
...
-    export QUILT_REFRESH_ARGS="-p ab --no-timestamps --no-index"
-    export QUILT_COLORS="diff_hdr=1;32:diff_add=1;34:diff_rem=1;31:diff_hunk=1;33:diff_ctx=35:diff_cctx=33"
-    if ! [ -d "$d/nx-X11/extras/Mesa.patches" ]; then mkdir "$d/nx-X11/extras/Mesa.patches"; fi
-    cd "$d/nx-X11/extras/Mesa/"
-fi
 
-quilt "$@"
+if [ -d "$d/nx-X11/extras" ]; then
+
+	if which quilt 1>/dev/null; then
```
uli42: I've never seen anyone using 1> ...

mesa-quilt
```diff
@@ -18,17 +18,73 @@
...
+		echo "$(basename $0): quilt utility found, so we use it..."
+		echo
+
+		# if in nx-libs tree with unset $QUILT_PATCHES
+		export QUILT_PATCHES="../Mesa.patches"
+		export QUILT_PATCH_OPTS="--reject-format=unified"
+		export QUILT_DIFF_ARGS="-p ab --no-timestamps --no-index --color=auto"
+		export QUILT_REFRESH_ARGS="-p ab --no-timestamps --no-index"
+		export QUILT_COLORS="diff_hdr=1;32:diff_add=1;34:diff_rem=1;31:diff_hunk=1;33:diff_ctx=35:diff_cctx=33"
+		if ! [ -d "$d/nx-X11/extras/Mesa.patches" ]; then mkdir "$d/nx-X11/extras/Mesa.patches"; fi
```
uli42: use mkdir -p without the if

mesa-quilt
uli42: +1

mesa-quilt
uli42: @Ionic should be sed without `-i` here

uli42: One more thing: I prefer compact shell scripts, this one has far too many empty lines for me. But YMMV

uli42 (changes requested):

mesa-quilt
Ionic: Still, it's never a bad idea to quote as much as possible.

Ionic (changes requested):

mesa-quilt
sunweaver: > I figure you want to do this (i.e., exporting the variables and moving into the other directory) only after checking for [ -z "$QUILT_PATCHES" ]?

The comment is actually a left over, I moved it further up and dropped the "unset QUILT_PATCHES" part.

sunweaver: > Although, strictly speaking, the old implementation was buggy in the sense that it only moved into the directory of $QUILT_PATCHES was unset, but always unconditionally moved to the previous directory in its stack, so we shouldn't blindly put that back.

Indeed.

mesa-quilt
sunweaver: under_stood!

mesa-quilt
sunweaver: > `grep -E -v "^#.*" < "../Mesa.patches/series" ...`

Yes. Nice.

sunweaver: > Why quote the path here? It's a static string

@uli42: Quoting is _always_ good. It's become a habit...

mesa-quilt
sunweaver: > `! grep -q "$patchfile" < "../Mesa.patches/.applied-patches"`

... Yes...

mesa-quilt
sunweaver: > `echo "$patchfile" >> "../Mesa.patches/.applied-patches"`

Indeed.

mesa-quilt
sunweaver: > One more thing: I prefer compact shell scripts, this one has far too many empty lines for me. But YMMV

@uli42: /me prefers readable scripts to compact scripts. I consider bash more as a progamming language than an admin tool, actually.

nx-libs.spec
sunweaver: > `application`

Yes.

Using `... falling back to stupidly (un)applying our patches...` now...

mesa-quilt
sunweaver: > I don't see why climbing up until / gets hit makes any sense...

@uli42: I have a copy of the mesa-quilt script in ~/bin/. So I can use it from anywhere in the nx-libs tree and don't need to give the full path to the script either. Very handy.

mesa-quilt
sunweaver: > I've never seen anyone using 1> ...

Oh... common practise here. Suppress standard out, but see errors if they occur. Also...

```
<command> > /dev/null
```

is a bashism afaik.

In dash, you have to do:

```
<command> 1>/dev/null 2>/dev/null
```

... at least at the time that dash was introduced.

mesa-quilt
sunweaver: > use mkdir -p without the if

yes. Of course!

sunweaver:
@Ionic: @uli42: I have considered all of your suggestions. I push more commits to this PR branch, with atomic changes addressing each comment of yours individually.

Before merging, the history of this PR branch should be flattend... Don't merge right away, please.

mesa-quilt
Ionic: Oh, on the contrary, that's perfectly POSIX-compliant behavior:

>[2.7.2 Redirecting Output](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18)
>
>The two general formats for redirecting output are:
>
>[n]>word
>[n]>|word
>
>where the optional n represents the file descriptor number. If the number is omitted, the redirection shall refer to standard output (file descriptor 1).

If dash failed to implement that correctly (and yet claimed compliance), then it was broken.

nx-libs.spec
Ionic: Still here.

mesa-quilt
```diff
@@ -88,6 +88,11 @@ if [ -d "$d/nx-X11/extras" ]; then
 
 		fi
 
+		# remove the .applied-patches file, if empty again
+		if [ ! -s "../Mesa.patches/.applied-patches" ]; then
+			rm "../Mesa.patches/.applied-patches"
```
Ionic: Can fail if the file does not exist, use `rm -f` instead to ignore a nonexistent file.

Ionic (changes requested):
Also still using `sed -i`, which is non-portable.

mesa-quilt
sunweaver: > Can fail if the file does not exist, use rm -f instead to ignore a nonexistent file.

Ack. Done.

sunweaver: > Still here. (appliance -> application)

Done

mesa-quilt
sunweaver: > sed -i is not portable. :(

Got that now.

sunweaver:
@Ionic: next iteration for #470, please.

mesa-quilt
```diff
@@ -77,7 +77,8 @@ if [ -d "$d/nx-X11/extras" ]; then
 				if [ -n "$patchfile" ] && [ -r "../Mesa.patches/$patchfile" ]; then
 					echo "Removing patch ../Mesa.patches/$patchfile"
 					patch -R -p1 < "../Mesa.patches/$patchfile"
-					sed -i "/$patchfile/d" "../Mesa.patches/.applied-patches"
+					sed -e "/$patchfile/d" "../Mesa.patches/.applied-patches" > "../Mesa.patches/.applied-patches.tmp"
```
Ionic: Okay, much better.

There's still a big caveat, though: sed interprets `${patchfile}` as a regexp and there's no way to turn that off. We should be mostly fine if the filenames are sane, but once they contain chars like `[`, `]` and others, this will fail badly. Even without "evil" characters, each full stop character in filenames will be interpreted as "any character", which might lead to problems.

We could work around that, but it's too complicated for that PR. Will clean that up some time later.

Ionic (approved):
All done, squash and rebase against 3.6.x, please.

nx-libs.spec
```diff
@@ -20,6 +20,19 @@ BuildRequires:  libjpeg-devel
 BuildRequires:  pkgconfig
 BuildRequires:  imake
 
+# ideally we build with quilt (for mesa-quilt patch appliance script),
+# it seems Fedora has it...
+%if 0%{?fedora_version}
```
Ionic: It's `fedora`, not `fedora_version`.

Ionic (changes requested):

Ionic (approved):

PR #475, opened 2017-07-03
Pr/maximum size

uli42:
Set the limit for maximum screen size to 32767x32767

Please test if this works with your favourite window manager.


nx-X11/programs/Xserver/randr/rrinfo.c
```diff
@@ -81,7 +81,6 @@ RRScanOldConfig(ScreenPtr pScreen, Rotation rotations)
     RRModePtr mode, newMode = NULL;
     int i;
     CARD16 minWidth = MAXSHORT, minHeight = MAXSHORT;
-    CARD16 maxWidth = 0, maxHeight = 0;
     CARD16 width, height;
```
sunweaver: @uli42: I don't really like it that we have to mess around with code in Xserver/randr/. Can't this be addressed insider the nxagent code itself?

I wonder, if this could be approached differently by looking more deeply at nxagentRandRInitSizes() (in hw/nxagent/Extensions.c) so that we allow for more screen sizes initially. (Just a guess). Also, I wonder why RRGetInfo actually gets called as we have an override function named nxagentRandRGetInfo() (also in Extensions.c).

nx-X11/programs/Xserver/hw/nxagent/Events.c
```diff
@@ -639,9 +639,6 @@ void nxagentShadowSwitchResizeMode(ScreenPtr pScreen)
     nxagentShadowCreateMainWindow(screenInfo.screens[DefaultScreen(nxagentDisplay)], screenInfo.screens[0]->root,
                                       screenInfo.screens[0]->root -> drawable.width, screenInfo.screens[0]->root -> drawable.height);
 
-    sizeHints.max_width = nxagentOption(RootWidth);
-    sizeHints.max_height = nxagentOption(RootHeight);
-
```
sunweaver: Same here... The WM_NORMAL_HINTS property should always match the current situation of the screen layout, IMHO.

nx-X11/programs/Xserver/hw/nxagent/Screen.c
```diff
@@ -4450,8 +4425,8 @@ void nxagentShadowAdaptToRatio(void)
 
   nxagentShadowCreateMainWindow(pScreen, screenInfo.screens[0]->root, nxagentShadowWidth, nxagentShadowHeight);
 
-  sizeHints.max_width = WidthOfScreen(DefaultScreenOfDisplay(nxagentDisplay));
-  sizeHints.max_height = HeightOfScreen(DefaultScreenOfDisplay(nxagentDisplay));
+  sizeHints.max_width = MAXSHORT;
+  sizeHints.max_height = MAXSHORT;
```
sunweaver: Have you actually testing the above code in nxagentShadowAdaptToRatio()?

sunweaver (dismissed):
uli42: I don't think that the fix you propose here is the right solution to the underlying problem. I am not even sure, if what you saw as a problem really is a problem, or rather _the_ problem.

We need to discuss this further, maybe we need some screen casts, that demonstrate the problem.

nx-X11/programs/Xserver/randr/rrinfo.c
uli42: Good question, need to investigate.

nx-X11/programs/Xserver/hw/nxagent/Screen.c
uli42: honestly, no.

uli42:
For clarification about what this patch addresses: I normally connect from different clients to my long-running session. Some clients have two monitors, some have one, and also the resolutions differ. As long as the initial session start happened on the client with the biggest resolution I have no problems. But if I happen to start (not reconnect!) the session from a "smaller" client that client's resolution is the value that shows up in xrandr. Reconnecting from a "bigger" client leaves me with a session that is not extendable to the full resolution. As I am using fullscreen sessions this leaves me with some black bars of unusable screen estate with no way to do something against it.

uli42:
Please review once again

nx-X11/programs/Xserver/hw/nxagent/Extensions.c
```diff
@@ -284,9 +284,9 @@ static int nxagentRandRInitSizes(ScreenPtr pScreen)
                  1024,  768,  900,  900, 1200, 1050, 1080, 1200, 0, 0};
 */
   int w[] = {0, 320, 640, 640, 800, 800, 1024, 1024, 1152, 1280, 1280, 1280, 1360,
-                 1440, 1600, 1600, 1680, 1920, 1920, 0, 0};
+	     1440, 1600, 1600, 1680, 1920, 1920, 32767, 0, 0};
```
sunweaver: Why not MAXSHORT here instead of 32767?

nx-X11/programs/Xserver/hw/nxagent/Extensions.c
```diff
@@ -284,9 +284,9 @@ static int nxagentRandRInitSizes(ScreenPtr pScreen)
...
   int h[] = {0, 240, 360, 480, 480, 600,  600,  768,  864,  720,  800, 1024,  768,
-                  900,  900, 1200, 1050, 1080, 1200, 0, 0};
+	     900,  900, 1200, 1050, 1080, 1200, 32767, 0, 0};
```
sunweaver: Same here...

sunweaver (changes requested):
Two more comments...

Also, while at it, shouldn't we add typical resolutions up to 8k here to nxagentRandRInitSizes() right away?

nx-X11/programs/Xserver/hw/nxagent/Extensions.c
uli42: MAXSHORT ist better, agreed. We can add some more resolutions but with the new maximum size you can add any resolution on your own.

However: As long as rrxinerama is active your won't ever see those resolutions.

sunweaver:
On  Mi 12 Jul 2017 21:23:44 CEST, Ulrich Sibiller wrote:

>
>
>
>>    int h[] = {0, 240, 360, 480, 480, 600,  600,  768,  864,  720,   
>> 800, 1024,  768,
> -                  900,  900, 1200, 1050, 1080, 1200, 0, 0};
> +	     900,  900, 1200, 1050, 1080, 1200, 32767, 0, 0};
>
> MAXSHORT ist better, agreed. We can add some more resolutions but  
> with the new maximum size you can add any resolution on your own.

Yep.

> However: As long as rrxinerama is active your won't ever see those  
> resolutions.

Ok, then leave it as is (except using MAXSHORT) and update your PR branch.

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

sunweaver (approved):

sunweaver:
@uli42: please rebase against latest 3.6.x branch, so I can merge.

sunweaver:
@uli42: What is the status of #475? I already approved the merge, but you added more commits without comment. Was that by accident? It feels that some of the newer commits should be part of a second PR.

Please give feedback on your intention for this PR. Thanks!

uli42:
To answer your question: no, those commits where not by accident, I simply noticed that we need them in the `-rrxinerama` case.

This patch is working but kind of worthless. It revealed a general problem: with rrxinerama activated (which is the default) modifying screens via xrandr is not really working well due to the fact the the xinerama code will throw away what the user specified on the next move or resize of the nx window. I have not yet come up with a proper idea how to handle that. You can, however, use this code together `-rrxinerama`.

sunweaver:
@Ionic: can you take a second look at that (with my look being the first). I've stumbled over this quite often when switching sessions between machines. Esp. when switching to machines with 4k displays, the size limitation of the previous host is nasty. I'd love to get this changeset in.

uli42:
sunweaver: so you saw this limitation even with active  rrxinerama?

sunweaver:
@uli42: @Ionic: we should get this PR merged over the weekend, IMHO.

sunweaver:
@uli42: @Ionic: this PR #475 has become quite stale but is still an issue in real life cases. I am inclined to rebase and merge it. Thoughts? Vetos?

sunweaver:
What about this one? The maximum size issue is some sort of a problem, indeed. I was about to rebase it, but then spotted the top commit on this branch. Please cleanup (and rebase).

uli42:
must check - but not today

sunweaver:
> must check - but not today

Gentle reminder...

PR #478, opened 2017-07-04
Report Xlib-side window IDs to session.log in machine readable form. …

sunweaver:
…This feature can be enabled by the cmdline options -reportwids and -reportprivatewids.

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Display.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Display.c
@@ -198,6 +198,7 @@ unsigned long startTime;
  */
 
 extern void RejectWellKnownSockets(void);
+extern Bool nxagentReportWindowIds;
 
 int nxagentServerOrder()
 {
@@ -1334,8 +1335,12 @@ FIXME: Use of nxagentParentWindow is strongly deprecated.
                                        CopyFromParent,
                                        0L, NULL);
 
+  if (nxagentReportWindowIds) {
+    fprintf(stderr, "NXAGENT_WINDOW_ID: CONFINEMENT_WINDOW,WID:[0x%x]\n",
+              nxagentConfineWindow);
+  }
   #ifdef TEST
-  fprintf(stderr, "nxagentOpenDisplay: Created agent's confine window with id [%ld].\n",
+  fprintf(stderr, "nxagentOpenDisplay: Created agent's confine window with id [0x%x].\n",
               nxagentConfineWindow);
   #endif
 
@@ -2842,8 +2847,12 @@ Bool nxagentReconnectDisplay(void *p0)
                                      CopyFromParent,
                                      0L, NULL);
 
+  if (nxagentReportWindowIds) {
+    fprintf(stderr, "NXAGENT_WINDOW_ID: CONFINEMENT_WINDOW,WID:[0x%x]\n",
+              nxagentConfineWindow);
+  }
   #ifdef TEST
-  fprintf(stderr, "nxagentReconnectDisplay: Created agent's confine window with id [%ld].\n",
+  fprintf(stderr, "nxagentReconnectDisplay: Created agent's confine window with id [0x%x].\n",
               nxagentConfineWindow);
   #endif
 
--- a/nx-X11/programs/Xserver/hw/nxagent/Screen.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Screen.c
@@ -139,6 +139,8 @@ extern Pixmap nxagentIconPixmap;
 extern Pixmap nxagentIconShape;
 extern Bool useXpmIcon;
 
+extern Bool nxagentReportWindowIds;
+
 Window nxagentDefaultWindows[MAXSCREENS];
 Window nxagentInputWindows[MAXSCREENS];
 Window nxagentScreenSaverWindows[MAXSCREENS];
@@ -396,9 +398,13 @@ Window nxagentCreateIconWindow()
                                     DefaultVisual(nxagentDisplay, DefaultScreen(nxagentDisplay)),
                                         valuemask, &attributes);
 
+  if (nxagentReportWindowIds)
+  {
+    fprintf (stderr, "NXAGENT_WINDOW_ID: ICON_WINDOW,WID:[0x%x]\n", nxagentIconWindow);
+  }
   #ifdef TEST
-  fprintf(stderr, "nxagentCreateIconWindow: Created new icon window with id [%lu].\n",
-              (long unsigned int)nxagentIconWindow);
+  fprintf(stderr, "nxagentCreateIconWindow: Created new icon window with id [0x%x].\n",
+              nxagentIconWindow);
   #endif
 
   /*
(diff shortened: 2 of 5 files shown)


sunweaver:
@Ionic: please review this PR as paid work. It will be needed for properly managing nxagent instances via Telekinesis.

sunweaver:
@gznget: Let me know if the implementation provided by PR #478 works for Telekinesis / Mother Forker handling, of if anything needs to be adapted.

nx-X11/programs/Xserver/hw/nxagent/Display.c
```diff
@@ -2842,8 +2847,12 @@ Bool nxagentReconnectDisplay(void *p0)
                                      CopyFromParent,
                                      0L, NULL);
 
+  if (nxagentReportWindowIds) {
+    fprintf(stderr, "NXAGENT_WINDOW_ID: CONFINE,WID:[0x%x]\n",
```
uli42: Not sure if I get this right, but in line 1339 you use "CONFINEMENT_WINDOW" while here you are using "CONFINE".

nx-X11/programs/Xserver/hw/nxagent/Screen.c
```diff
@@ -1760,9 +1766,13 @@ N/A
                            valuemask , &attributes);
       }
 
+      if (nxagentReportWindowIds)
+      {
+        fprintf (stderr, "NXAGENT_WINDOW_ID: SCREEN_WINDOW:%d,WID:[0x%x]\n", pScreen -> myNum, nxagentInputWindows[pScreen->myNum]);
```
uli42: Hmm, why not enclose the screen number in brackets, too? That's usually how it is handled within all nxagent  debug/test outputs. Also normally we do not use ":". So I suggest:

```
fprintf (stderr, "NXAGENT_WINDOW_ID: SCREEN_WINDOW [%d][0x%x]\n", pScreen -> myNum, nxagentInputWindows[pScreen->myNum]);
```

nx-X11/programs/Xserver/hw/nxagent/Screen.c
```diff
@@ -1760,9 +1766,13 @@ N/A
...
+      }
       #ifdef TEST
-      fprintf(stderr, "nxagentOpenScreen: Created new default window with id [%ld].\n",
-                  (long int)nxagentDefaultWindows[pScreen->myNum]);
+      fprintf(stderr, "nxagentOpenScreen: Created new default window with id [0x%x].\n",
+              nxagentDefaultWindows[pScreen->myNum]);
```
uli42, line 1775: screen number is missing here

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -372,6 +372,12 @@ force use of NX protocol messages assuming communication through nxproxy
 .B \-nxrealwindowprop
 set property NX_REAL_WINDOW for each X11 client inside NX Agent, providing the window XID of the corresponding window object on the X server that NX Agent runs on
 .TP 8
+.B \-reportwids
+explicitly tell NX Agent to report its externally exposed X11 window IDs to the session log (in machine readable form), so that external parsers can obtain that information from there
+.TP 8
+.B \-reportprivatewids
+explicitly tell NX Agent to report X11 window IDs of internally created window objects to the session log (in machine readable form), so that external parsers can obtain that information from there; this creates a lot of output and may affect performance
+.TP 8
```
uli42, line 380: So is this really a debugging option? Or is is needed for Telekinesis to work?

Generally: why has this to be in the session log? Shouldn't we rather user a separate logfile with only machine readable stuff like this?

uli42 (changes requested):

sunweaver:
Hi Uli,

On  Mi 05 Jul 2017 09:13:05 CEST, Ulrich Sibiller wrote:

> uli42 requested changes on this pull request.
>
>
>
>> @@ -2842,8 +2847,12 @@ Bool nxagentReconnectDisplay(void *p0)
>                                       CopyFromParent,
>                                       0L, NULL);
>
> +  if (nxagentReportWindowIds) {
> +    fprintf(stderr, "NXAGENT_WINDOW_ID: CONFINE,WID:[0x%x]\n",
>
> Not sure if I get this right, but in line 1339 you use  
> "CONFINEMENT_WINDOW" while here you are using "CONFINE".

Good catch. Fixing it now.

>> @@ -1760,9 +1766,13 @@ N/A
>                             valuemask , &attributes);
>        }
>
> +      if (nxagentReportWindowIds)
> +      {
> +        fprintf (stderr, "NXAGENT_WINDOW_ID:  
> SCREEN_WINDOW:%d,WID:[0x%x]\n", pScreen -> myNum,  
> nxagentInputWindows[pScreen->myNum]);
>
> Hmm, why not enclose the screen number in brackets, too? That's  
> usually how it is handled within all nxagent  debug/test outputs.  
> Also normally we do not use ":". So I suggest:

The [...] brackets are ok for the screen num. Adding those.

Please note that this is not about debugging/testing. Obtaining the  
windows' XID directly from nxagent allows us to get rid of the guess  
work code in nxagent, which especially helpful, when having several  
nxagent sessions appearing on one local X display.

> ```
> fprintf (stderr, "NXAGENT_WINDOW_ID: SCREEN_WINDOW [%d][0x%x]\n",  
> pScreen -> myNum, nxagentInputWindows[pScreen->myNum]);
> ```
>
>>        #ifdef TEST
> -      fprintf(stderr, "nxagentOpenScreen: Created new default  
> window with id [%ld].\n",
> -                  (long int)nxagentDefaultWindows[pScreen->myNum]);
> +      fprintf(stderr, "nxagentOpenScreen: Created new default  
> window with id [0x%x].\n",
> +              nxagentDefaultWindows[pScreen->myNum]);
>
> screen number is missing here

The TEST part here has been +/- left untouched. Only change is the  
switch from %ld to %x. However, as this should be done to all TEST  
output passages whenever a window ID is written to the log, we can  
surely do that in a separate, more overall commit.

>> @@ -372,6 +372,12 @@ force use of NX protocol messages assuming  
>> communication through nxproxy
>  .B \-nxrealwindowprop
>  set property NX_REAL_WINDOW for each X11 client inside NX Agent,  
> providing the window XID of the corresponding window object on the X  
> server that NX Agent runs on
>  .TP 8
> +.B \-reportwids
> +explicitly tell NX Agent to report its externally exposed X11  
> window IDs to the session log (in machine readable form), so that  
> external parsers can obtain that information from there
> +.TP 8
> +.B \-reportprivatewids
> +explicitly tell NX Agent to report X11 window IDs of internally  
> created window objects to the session log (in machine readable  
> form), so that external parsers can obtain that information from  
> there; this creates a lot of output and may affect performance
> +.TP 8
>
> So is this really a debugging option? Or is is needed for  
> Telekinesis to work?

The -reportwids option is for Telekinesis to actually work. Easing  
detection of the nxagent's WID.

> Generally: why has this to be in the session log? Shouldn't we  
> rather user a separate logfile with only machine readable stuff like  
> this?

Actually, I'd love to see some bidirectional communication over some  
sort of socket between nxagent and some external control application.

For now, writing important information for Telekinesis to the  
session.log is the least invasive way to go.

Please look at the commit once more, with above changes worked in.
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

sunweaver:
@uli42: Could you take another look at this PR (#478)? Thanks.

uli42 (approved):

uli42:
Further considerations:
1. why does anyone need the private window ID? I understand the demand for the real window id, but what's the point about the private window?
2. _if_ the private window is really needed then all machine-readable outputs should contain both the real _and_ the private id. Having them in different lines does not make the relation obvious. if for any reason one line of output gets lost or garbled you cannot relate real and private ids in a safe fashion anymore.
3. If both are required then there's no need for two switches, one should be enough in that case.
4. the format of the machine readable output is not documented AFAIKS, and the print statements do not contain a hint that this output format must not be changed. Maybe introduce a print_machine_readable() wrapper function so it gets clearer.

Issue #191, opened 2016-08-02
nxagent -full


uli42:
`nxagent -h` shows

`-full                  utilize full regeneration`

That switch opens the session in fullscreen for me, so the description seems a bit odd at least.

Ionic:
Good catch, I think that should be changed to "fullscreen" then... but what exactly is the "full regeneration" and how does it toggled?!

uli42:
`-full` comes from Xnest and is described there as

> -full
> This option tells Xnest to utilize full regeneration of real server objects and reopen a new connection to the real server each time the nested server regenerates. The sample server implementation regenerates all objects in the server when the last client of this server terminates. When this happens, Xnest by default maintains the same top-level window and the same real server connection in each new generation. If the user selects full regeneration, even the top-level window and the connection to the real server will be regenerated for each server generation.

Using this option sets `nxagentFullGeneration` and `nxagentDoFullGeneration`; both are rarely used. I doubt that we need that feature, so it probably can be dropped completely. It is missing in the manpage, BTW.

sunweaver:
HI Uli,

On  Di 11 Jul 2017 22:02:03 CEST, Ulrich Sibiller wrote:

> `-full` comes from Xnest and is described there as
>
> ```
> -full
> This option tells Xnest to utilize full regeneration of real server  
> objects and reopen a new connection to the real server each time the  
> nested server regenerates. The sample server implementation  
> regenerates all objects in the server when the last client of this  
> server terminates. When this happens, Xnest by default maintains the  
> same top-level window and the same real server connection in each  
> new generation. If the user selects full regeneration, even the  
> top-level window and the connection to the real server will be  
> regenerated for each server generation.
> ```
>
> Using this option sets `nxagentFullGeneration` and  
> `nxagentDoFullGeneration`; both are rarely used. I doubt that we  
> need that feature, so I probably can be dropped completely.

Looks sensible. There is a lot of Xnest copy+paste code (still) in nxagent.

Will you propose a PR for this? Or shall I?

Mike
-- 

DAS-NETZWERKTEAM
mike gabriel, herweg 7, 24357 fleckeby
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

GnuPG Fingerprint: 9BFB AEE8 6C0A A5FF BF22  0782 9AF4 6B30 2577 1B31
mail: mike.gabriel@das-netzwerkteam.de, http://das-netzwerkteam.de

uli42:
[...]

PR #485, opened 2017-07-11
remove -full, making FullGeneration the default

uli42:
Addresses ArcticaProject/nx-libs#191

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -137,7 +137,6 @@ char nxagentDialogName[256];
 char nxagentSessionId[256] = {0};
 char *nxagentOptionFile;
 
-Bool nxagentFullGeneration = False;
 int nxagentDefaultClass = TrueColor;
 Bool nxagentUserDefaultClass = False;
 int nxagentDefaultDepth;
@@ -407,11 +406,6 @@ int ddxProcessArgument(int argc, char *argv[], int i)
 
   if (!strcmp(argv[i], "-nxrealwindowprop")) {
     nxagentRealWindowProp = True;
-    return 1;
-  }
-
-  if (!strcmp(argv[i], "-full")) {
-    nxagentFullGeneration = True;
     return 1;
   }
 
@@ -2081,7 +2075,6 @@ void ddxUseMsg()
 {
   ErrorF("-display string        display name of the real server\n");
   ErrorF("-sync                  synchronize with the real server\n");
-  ErrorF("-full                  utilize full regeneration\n");
   ErrorF("-class string          default visual class\n");
   ErrorF("-depth int             default depth\n");
   ErrorF("-geometry WxH+X+Y      window size and position\n");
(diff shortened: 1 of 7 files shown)


sunweaver:
> making FullGeneration the default

Are you sure, we want this? From the man page of Xnest as quoted in #191, I get the impression that making nxagentFullGeneration the default could create a lot of extra round trips (as the full generation happens betweem local Xserver and nxagent Xserver)... Have you checked this with xtrace?

uli42:
I have not checked this. AFAIU this is only needed when you close all clients. Normally Xnest will stay alive and not close down the Display connection. With FullGeneration it will close it and build a new one and then await new clients.

I am unsure now if that case ever happens in nxagent. We have -reset/-noreset, -once and -full. And we have nxagent quitting when the last client disconnects. Must think through this, help appreciated.

In any case we need a FullGeneration at the startup. So this seems like the natural default.

sunweaver:
Hi Uli,

On  Mi 12 Jul 2017 13:20:22 CEST, Ulrich Sibiller wrote:

> I have not checked this. AFAIU this is only needed when you close  
> all clients. Normally Xnest will stay alive and not close down the  
> Display connection. With FullGeneration it will close it and build a  
> new one and then await new clients.
>
> I am unsure now if that case ever happens in nxagent. We have  
> -reset/-noreset, -once and -full. And we have nxagent quitting when  
> the last client disconnects. Must think through this, help  
> appreciated.

hmmm... Basically, I'd suggest removing -full and leaving the default  
as if nxagentFullGeneration was False.

> In any case we need a FullGeneration at the startup. So this seems  
> like the natural default.

Note, that nxagentDoFullGeneration is always True at agent startup.  
(see Init.c).

nxagentFullGeneration is a static flag obtained from Args.c, where  
nxagentDoFullGeneration adapts to the situation at runtime.

Mike

-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

uli42:
I am aware of that code. But if I keep all that logic this patch does not make any sense, as I have to keep most of the code.

sunweaver:
@uli42: Where are we with this (#485)?

uli42:
Well, I must check if any benefit remains after doing it as proposed above. Also my initial bug report #191 does not seem appropriate anymore, I cannot trigger the described behaviour. Can you?

sunweaver:
@uli42: ping(?)

uli42:
Aehm, my last queytion was, if _you_ can trigger the fullscreen behaviour described in #191. So I'd better ping _you_ ;-)

sunweaver:
@uli42: No, I can't reproduce "-full" switching nxagent to full screen mode. Tested against i3wm and marco WM (from MATE).

sunweaver:
@uli42: You just updated the -full option in the man page. Is this PR obsolete? Shall -full stay?

uli42:
Well, I just added the text from xnest explaining what the switch should do. As long as we have it we should also include it in the manpage. I have not looked any further regarding this pr

sunweaver:
On  Mi 19 Dez 2018 13:21:56 CET, Ulrich Sibiller wrote:

> Well, I just added the text from xnest explaining what the switch  
> should do. As long as we have it we should also include it in the  
> manpage. I have not looked any further regarding this pr

Ok.

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

sunweaver:
sigh... I am sorry, but pr/drop-full-generation got forgotten meanwhile. It should still be merged, I guess. Please rebase. Thanks.

PR #488, opened 2017-07-17
nx-X11/programs/Xserver/Render.c: Re-introduce nxagentGlyphsCleanup()…

sunweaver:
… from libNX_Xrender's formerly removed XRenderGlyphsCleanup().


Ionic:
We can add this back and the commit looks good, but I wonder if it would create inconsistencies between remote and local glyphs (which might mean that hashes differ and glyphs that are supposed to be the same ones be regarded at different ones.)

I wonder why this was implemented in the libXrender copy instead of the nxagent DDX in the first place?

sunweaver:
Hi Ionic,

On  Do 24 Aug 2017 04:27:57 CEST, Mihai Moldovan wrote:

> We can add this back and the commit looks good, but I wonder if it  
> would create inconsistencies between remote and local glyphs (which  
> might mean that hashes differ and glyphs that are supposed to be the  
> same ones be regarded at different ones.)

Good question. If something like this happens, it is not visible.

> I wonder why this was implemented in the libXrender copy instead of  
> the nxagent DDX in the first place?

Bad implementation? No clue, really. The log entry for nxcomp/3.0.0-18  
hints to the implementation being present in nxcomp first and that got  
moved to libNX_Xrender. As the re-implementation in nxcomp will be  
more complex and the only use case would be nxproxy <-> nxproxy  
sessions, I'd say let's move it into nxagent for now. Also, there may  
have been good reasons for moving the code out of nxcomp.

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

Ionic (approved):
*shrug* I guess it won't hurt to include this. If necessary (missing glyphs or the like) we can simple disable this feature.

Waiting to merge PR#498 first, though.

sunweaver:
Hi Mihai,

On  Fr 25 Aug 2017 03:47:43 CEST, Mihai Moldovan wrote:

> Ionic approved this pull request.
>
> *shrug* I guess it won't hurt to include this. If necessary (missing  
> glyphs or the like) we can simple disable this feature.
>
> Waiting to merge PR#498 first, though.

branch has been rebased against 3.6.x and pushed. Ready for merging now.

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

PR #492, opened 2017-07-23
Pr/improve keyboard

uli42:
I have done lots of changes to the keyboard stuff:
1. fix keyboard file creation once again
2. add keyboard=clone (clone xkb configuration of remote side). Also accept x2goclient's null/null keyboard as a synonym for clone
3. extend manpage for keyboard stuff (options keyboard, noignore, noxkblock)
4. add new -xkblock option
5. switch from old 'xfree86' rules file to 'base'
6. code cleanups and simplifications

Many of my commits affect multiple Issues, simply because we have overlapping Issues for the keyboard stuff.

If "clone" proves stable I would like to make that the default. What so you think?

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
@@ -958,18 +958,39 @@ XkbError:
             goto XkbError;
           }
 
-          update_string(&model, nxagentKeyboard, i);
-          update_string(&layout, &nxagentKeyboard[i + 1], 0);
-
+          /*
+            The original nxagent only supports model/layout values
+            here. It uses these values together with the default rules
+            and empty variant and options. We use a more or less
+            compatible hack here: The special keyword rlmvo for model
+            means that the layout part of the string will contain a
+            full RMLVO config, separated by #, e.g.
+            rlmvo/base#pc105#de,us#nodeadkeys#lv3:rwin_switch
+          */
+          if (strncmp(nxagentKeyboard, "rlmvo/", 6))
+          {
+            const char * sep = "#";
+            char * rmlvo = strdup(&nxagentKeyboard[i+1]);
+            char * tmp = rmlvo;
+            /* strtok cannot handle empty fields, so use strsep */
+            update_string(&rules,    strsep(&tmp, sep), 0);
+            update_string(&model,    strsep(&tmp, sep), 0);
+            update_string(&layout,   strsep(&tmp, sep), 0);
+            update_string(&variants, strsep(&tmp, sep), 0);
+            update_string(&options,  strsep(&tmp, sep), 0);
+            free(rmlvo);
+          }
+          else
+          {
+            update_string(&model, nxagentKeyboard, i);
+            update_string(&layout, &nxagentKeyboard[i + 1], 0);
+            update_string(&variants, XKB_DFLT_KB_VARIANT, 0);
+            update_string(&options, XKB_DFLT_KB_OPTIONS, 0);
+          }
           /*
            * There is no description for pc105 on Solaris.
            * Need to revert to the closest approximation.
            */
-
-          #ifdef TEST
-          fprintf(stderr, "nxagentKeyboardProc: Using keyboard model [%s] with layout [%s].\n",
-                      model, layout);
-          #endif
 
           #ifdef __sun
 
@@ -990,6 +1011,8 @@ XkbError:
         {
           update_string(&layout, XKB_DFLT_KB_LAYOUT, 0);
           update_string(&model, XKB_DFLT_KB_MODEL, 0);
+          update_string(&variants, XKB_DFLT_KB_VARIANT, 0);
+          update_string(&options, XKB_DFLT_KB_OPTIONS, 0);
 
           #ifdef TEST
           fprintf(stderr, "nxagentKeyboardProc: Using default keyboard: model [%s] layout [%s].\n",
@@ -997,8 +1020,11 @@ XkbError:
           #endif
         }
 
-        update_string(&variants, XKB_DFLT_KB_VARIANT, 0);
-        update_string(&options, XKB_DFLT_KB_OPTIONS, 0);
+        #ifdef TEST
+            fprintf(stderr, "nxagentKeyboardProc: Using keyboard rules [%s] model [%s] layout [%s] variant [%s] options [%s].\n",
+                    rules, model, layout, variants, options);
+        #endif
+
 
         #ifdef TEST
         fprintf(stderr, "nxagentKeyboardProc: XkbInitialMap (option -xkbmap) is [%s]\n", XkbInitialMap ? XkbInitialMap : "NULL");
--- a/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
@@ -937,7 +937,9 @@ XkbError:
           layout.
         */
 
-        if (nxagentKeyboard && (strcmp(nxagentKeyboard, "query") != 0))
+        if (nxagentKeyboard &&
+            (strcmp(nxagentKeyboard, "query") != 0) &&
+            (strcmp(nxagentKeyboard, "clone") != 0))
         {
           for (i = 0; nxagentKeyboard[i] != '/' && nxagentKeyboard[i] != 0; i++);
 
@@ -1043,7 +1045,20 @@ XkbError:
           #endif
 
           nxagentWriteKeyboardFile(drulesLen, drules, dmodel, dlayout, dvariant, doptions);
-          nxagentKeycodeConversionSetup(drules, dmodel);
+          if (nxagentKeyboard && (strcmp(nxagentKeyboard, "clone") != 0))
+          {
+            /* Keycode conversion is not of use in clone mode */
+            nxagentKeycodeConversionSetup(drules, dmodel);
+          }
+
+          if (drules && nxagentKeyboard && (strcmp(nxagentKeyboard, "clone") == 0))
+          {
+            update_string(&rules, drules, 0);
+            update_string(&model, dmodel, 0);
+            update_string(&layout, dlayout, 0);
+            update_string(&variants, dvariant, 0);
+            update_string(&options, doptions, 0);
+          }
 
           if (drules)
           {
@@ -1784,7 +1799,9 @@ void nxagentTuneXkbWrapper(void)
     return;
   }
 
-  if (nxagentKeyboard && strcmp(nxagentKeyboard, "query") == 0)
+  if (nxagentKeyboard &&
+      (strcmp(nxagentKeyboard, "query") == 0 ||
+       strcmp(nxagentKeyboard, "clone") == 0))
   {
     nxagentDisableXkbExtension();
   }
--- a/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
@@ -923,6 +923,14 @@ XkbError:
         fprintf(stderr, "nxagentKeyboardProc: nxagentKeyboard is [%s].\n", nxagentKeyboard ? nxagentKeyboard : "NULL");
         #endif
 
+        if (nxagentKeyboard && (strcmp(nxagentKeyboard, "null/null") == 0))
+        {
+          #ifdef TEST
+          fprintf(stderr, "nxagentKeyboardProc: changing nxagentKeyboard from [null/null] to [clone].\n");
+          #endif
+          free(nxagentKeyboard);
+          nxagentKeyboard = strdup("clone");
+        }
 
         update_string(&rules, nxagentXkbGetRules(), 0);
 
--- a/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
@@ -159,7 +159,7 @@ extern        Status        XkbGetControls(
 #define XKB_CONFIG_FILE_X2GO "/etc/x2go/x2goagent.keyboard"
 #endif
 #ifndef XKB_DFLT_RULES_FILE
-#define XKB_DFLT_RULES_FILE  "xfree86"
+#define XKB_DFLT_RULES_FILE  "base"
 #endif
 #ifndef XKB_ALTS_RULES_FILE
 #define XKB_ALTS_RULES_FILE  "xorg"
--- a/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
@@ -94,6 +94,8 @@ void nxagentWriteKeyboardFile(unsigned int ruleslen, char *rules, char *model, c
 
 #endif /* XKB */
 
+static void update_string(char ** current, char *new, size_t maxlen);
+
 /*
  * Set here the required log level.
  */
@@ -699,6 +701,30 @@ void nxagentChangeKeyboardControl(DeviceIntPtr pDev, KeybdCtrl *ctrl)
   #endif
 }
 
+/*
+  update a string that has been allocated with len chars of new
+  string. If the new string is NULL nothing will be changed. If maxlen
+  is 0 the whole new string will be used.
+*/
+static void update_string(char ** current, char *new, size_t maxlen)
+{
+  if (!current || !new)
+    return;
+
+  if (*current)
+  {
+    free(*current);
+    *current = NULL;
+  }
+
+  if (maxlen > 0)
+    *current = strndup(new, maxlen);
+  else
+    *current = strdup(new);
+
+  return;
+}
+
 int nxagentKeyboardProc(DeviceIntPtr pDev, int onoff)
 {
   XModifierKeymap *modifier_keymap;
@@ -709,8 +735,7 @@ int nxagentKeyboardProc(DeviceIntPtr pDev, int onoff)
   CARD8 modmap[256];
   int i, j;
   XKeyboardState values;
-  char *model = NULL, *layout = NULL;
-  int free_model = 0, free_layout = 0;
+  char *rules = NULL, *model = NULL, *layout = NULL, *variants = NULL, *options = NULL;
   XkbDescPtr xkb = NULL;
 
   switch (onoff)
@@ -856,16 +881,13 @@ XkbError:
 
         XkbFreeKeyboard(xkb, XkbAllComponentsMask, True);
         xkb = NULL;
-        if (free_model) 
-        {
-          free_model = 0;
-          free(model);
-        }
-        if (free_layout)
-        {
-          free_layout = 0;
-          free(layout);
-        }
+        free(rules);
+        free(model);
+        free(layout);
+        free(variants);
+        free(options);
+
+        rules = model = layout = variants = options = NULL;
 #endif
         XGetKeyboardControl(nxagentDisplay, &values);
 
@@ -892,7 +914,6 @@ XkbError:
         char *nxagentXkbConfigFilePath;
 
         XkbComponentNamesRec names = {0};
-        char *rules, *variants, *options;
 
         #ifdef TEST
         fprintf(stderr, "nxagentKeyboardProc: Using XKB extension.\n");
@@ -903,7 +924,7 @@ XkbError:
         #endif
 
 
-        rules = nxagentXkbGetRules();
+        update_string(&rules, nxagentXkbGetRules(), 0);
 
         /*
           from nxagent changelog:
@@ -927,11 +948,8 @@ XkbError:
             goto XkbError;
           }
 
-          free_model = 1;
-	  model = strndup(nxagentKeyboard, i);
-
-          free_layout = 1;
-	  layout = strdup(&nxagentKeyboard[i + 1]);
+          update_string(&model, nxagentKeyboard, i);
+          update_string(&layout, &nxagentKeyboard[i + 1], 0);
 
           /*
            * There is no description for pc105 on Solaris.
@@ -960,8 +978,8 @@ XkbError:
         }
         else
         {
-          layout = XKB_DFLT_KB_LAYOUT;
-          model = XKB_DFLT_KB_MODEL;
+          update_string(&layout, XKB_DFLT_KB_LAYOUT, 0);
+          update_string(&model, XKB_DFLT_KB_MODEL, 0);
 
           #ifdef TEST
           fprintf(stderr, "nxagentKeyboardProc: Using default keyboard: model [%s] layout [%s].\n",
@@ -969,8 +987,8 @@ XkbError:
           #endif
         }
 
-        variants = XKB_DFLT_KB_VARIANT;
-        options = XKB_DFLT_KB_OPTIONS;
+        update_string(&variants, XKB_DFLT_KB_VARIANT, 0);
+        update_string(&options, XKB_DFLT_KB_OPTIONS, 0);
 
         #ifdef TEST
         fprintf(stderr, "nxagentKeyboardProc: XkbInitialMap (option -xkbmap) is [%s]\n", XkbInitialMap ? XkbInitialMap : "NULL");
@@ -1099,30 +1117,11 @@ XkbError:
             fclose(file);
             goto XkbError;
           }
-          if (config.rules_file)
-            rules = config.rules_file;
-          if (config.model)
-          {
-            if (free_model)
-            {
-              free_model = 0; 
-              free(model);
-            }
-            model = config.model;
-          }
-          if (config.layout)
-          {
-            if (free_layout)
-            {
-              free_layout = 0;
-              free(layout);
-            }
-            layout = config.layout;
-          }
-          if (config.variant)
-            variants = config.variant;
-          if (config.options)
-            options = config.options;
+          update_string(&rules, config.rules_file, 0);
+          update_string(&model, config.model, 0);
+          update_string(&layout, config.layout, 0);
+          update_string(&variants, config.variant, 0);
+          update_string(&options, config.options, 0);
 
           free(nxagentXkbConfigFilePath);
           nxagentXkbConfigFilePath = NULL;
@@ -1176,17 +1175,13 @@ XkbEnd:
           NXShadowInitKeymap(&(pDev->key->curKeySyms));
         }
 
-        if (free_model) 
-        {
-          free_model = 0;
-          free(model);
-        }
+        free(rules);
+        free(model);
+        free(layout);
+        free(variants);
+        free(options);
 
-        if (free_layout)
-        {
-          free_layout = 0;
-          free(layout);
-        }
+        rules = model = layout = variants = options = NULL;
 
         XkbFreeKeyboard(xkb, XkbAllComponentsMask, True);
         xkb = NULL;
--- a/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
@@ -1105,15 +1105,7 @@ XkbError:
         if (xkb == NULL || xkb->geom == NULL)
         {
           #ifdef TEST
-          fprintf(stderr, "nxagentKeyboardProc: No current keyboard.\n");
-          if (xkb == NULL)
-          {
-            fprintf(stderr, "nxagentKeyboardProc: xkb is null.\n");
-          }
-          else
-          {
-            fprintf(stderr, "nxagentKeyboardProc: xkb->geom is null.\n");
-          }
+          fprintf(stderr, "%s: No current keyboard: xkb%s is null.\n", __func__, xkb ? "->geom" : "");
           fprintf(stderr, "nxagentKeyboardProc: Going to set rules and init device.\n");
           #endif
           #ifdef DEBUG
--- a/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
@@ -88,7 +88,7 @@ is" without express or implied warranty.
 static int nxagentXkbGetNames(char **rules, char **model, char **layout,
                                   char **variant, char **options);
 
-static void nxagentKeycodeConversionSetup(void);
+static void nxagentKeycodeConversionSetup(char *rules, char *model);
 
 void nxagentWriteKeyboardFile(unsigned int ruleslen, char *rules, char *model, char *layout, char *variant, char *options);
 
@@ -999,9 +999,41 @@ XkbError:
           goto XkbError;
         }
 
-        xkb = XkbGetKeyboard(nxagentDisplay, XkbGBN_AllComponentsMask, XkbUseCoreKbd);
+        {
+          char *drules = NULL;
+          char *dmodel = NULL;
+          char *dlayout = NULL;
+          char *dvariant = NULL;
+          char *doptions = NULL;
 
-        nxagentKeycodeConversionSetup();
+          #ifdef DEBUG
+          unsigned int drulesLen =
+          #endif
+          nxagentXkbGetNames(&drules, &dmodel, &dlayout,
+                             &dvariant, &doptions);
+
+          #ifdef DEBUG
+          if (drulesLen && drules && dmodel)
+          {
+            fprintf(stderr, "%s: Remote: [rules='%s',model='%s',layout='%s',variant='%s',options='%s'].\n",
+                    __func__, drules, dmodel, dlayout, dvariant, doptions);
+          }
+          else
+          {
+            fprintf(stderr, "%s: Failed to retrieve remote rules.\n", __func__);
+          }
+          #endif
+
+          nxagentWriteKeyboardFile(drulesLen, drules, dmodel, dlayout, dvariant, doptions);
+          nxagentKeycodeConversionSetup(drules, dmodel);
+
+          if (drules)
+          {
+            XFree(drules);
+          }
+        }
+
+        xkb = XkbGetKeyboard(nxagentDisplay, XkbGBN_AllComponentsMask, XkbUseCoreKbd);
 
         if (xkb == NULL || xkb->geom == NULL)
         {
@@ -1901,15 +1933,8 @@ void nxagentWriteKeyboardFile(unsigned int ruleslen, char *rules, char *model, c
   }
 }
 
-void nxagentKeycodeConversionSetup(void)
+void nxagentKeycodeConversionSetup(char * rules, char * model)
 {
-  char *drules = NULL;
-  char *dmodel = NULL;
-  char *dlayout = NULL;
-  char *dvariant = NULL;
-  char *doptions = NULL;
-  unsigned int drulesLen;
-
   if (nxagentOption(KeycodeConversion) == KeycodeConversionOff)
   {
     fprintf(stderr, "Info: Keycode conversion is off\n");
@@ -1923,28 +1948,9 @@ void nxagentKeycodeConversionSetup(void)
     return;
   }
 
-  nxagentKeycodeConversion = False;
-
-  drulesLen = nxagentXkbGetNames(&drules, &dmodel, &dlayout,
-                                     &dvariant, &doptions);
-
-  #ifdef DEBUG
-  if (drulesLen != 0 && drules && dmodel)
-  {
-    fprintf(stderr, "%s: Remote: [rules='%s',model='%s',layout='%s',variant='%s',options='%s'].\n",
-	    __func__, drules, dmodel, dlayout, dvariant, doptions);
-  }
-  else
-  {
-    fprintf(stderr, "%s: Failed to retrieve remote rules.\n", __func__);
-  }
-  #endif
-
-  nxagentWriteKeyboardFile(drulesLen, drules, dmodel, dlayout, dvariant, doptions);
-
-  if (drules && dmodel &&
-      (strcmp(drules, "evdev") == 0 ||
-       strcmp(dmodel, "evdev") == 0))
+  if (rules && model &&
+      (strcmp(rules, "evdev") == 0 ||
+       strcmp(model, "evdev") == 0))
   {
     #ifdef DEBUG
     fprintf(stderr, "%s: Activating KeyCode conversion.\n", __func__);
@@ -1955,12 +1961,13 @@ void nxagentKeycodeConversionSetup(void)
   }
   else
   {
-    fprintf(stderr, "Info: Keycode conversion auto-determined as off\n");
-  }
+    #ifdef DEBUG
+    fprintf(stderr, "%s: Deactivating KeyCode conversion.\n", __func__);
+    #endif
 
-  if (drules)
-  {
-    XFree(drules);
+    fprintf(stderr, "Info: Keycode conversion auto-determined as off\n");
+
+    nxagentKeycodeConversion = False;
   }
 }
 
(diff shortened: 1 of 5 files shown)


nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -909,6 +909,17 @@ N/A
...
         rules = nxagentXkbGetRules();
 
+        /*
+          from nxagent changelog:
+          2.0.22:
+          - Implented handling of value "query" for nxagentKbtype. This value
+          is passed by the NX client for MacOSX. If value of nxagentKbtype is
+          "query" or NULL we init keyboard by core protocol functions reading
+          the keyvoard mapping of the X server. The property _XKB_RULES_NAMES
+          is always set on the root window with default values of model and
```
sunweaver: typo: s/keyvoard/keyboard/

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -370,13 +370,13 @@ The nx-X11 system adds the following command line arguments:
...
+set property NX_REAL_WINDOW for each X11 client inside \fBnxagent\fR, providing the window XID of the corresponding window object on the X server that \fBnxagent\fR runs on
 .TP 8
 .B \-reportwids
-explicitly tell NX Agent to report its externally exposed X11 window IDs to the session log (in machine readable form), so that external parsers can obtain that information from there
+explicitly tell \fBnxagent\fR to report its externally exposed X11 window IDs to the session log (in machine readable form), so that external parsers can obtain that information from there
 .TP 8
 .B \-reportprivatewids
-explicitly tell NX Agent to report X11 window IDs of internally created window objects to the session log (in machine readable form), so that external parsers can obtain that information from there; this creates a lot of output and may affect performance
+explicitly tell \fBnxagent\fR to report X11 window IDs of internally created window objects to the session log (in machine readable form), so that external parsers can obtain that information from there; this creates a lot of output and may affect performance
 .TP 8
```
sunweaver: It seems that this man page fix does not belong here. Please file as separate PR. Thanks!

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -406,13 +406,19 @@ disable use of shared pixmaps
 enable use of shared pixmaps
 .TP 8
 .B \-noignore
-don't ignore pointer and keyboard configuration changes mandated by clients
+don't ignore pointer and keyboard configuration changes mandated by
+clients. Using this commands like disabling the keyboard bell (xset -b)
```
sunweaver: missing ",": Using this, commands like...

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -497,6 +497,10 @@ XKEYBOARD will most likely not match the experienced settings. This is
 required on MacOS X to handle some keyboard problems that are special
 for this platform.
 .TP 8
+.I clone
+ask the real X server for the keyboard settings using XKEYBOARD protocol
+functions and clone them. This is the recommend setting.
```
sunweaver: typo: s/recommend/recommended/

sunweaver: also in the commit msg of the "add keyboard=clone configuration option" commit:

s/Should the default but is/Should be the default but is/

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -883,28 +883,28 @@ N/A
 
         XkbFreeKeyboard(xkb, XkbAllComponentsMask, True);
         xkb = NULL;
-	free(rules);
-	free(model);
-	free(layout);
-	free(variants);
-	free(options);
+        free(rules);
```
sunweaver: typo in commit message: 

"Keyboard.c: Fix indention" -> "Keyboard.c: Fix indentation"

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1899,10 +1899,32 @@ static int nxagentXkbGetNames(char **rules, char **model, char **layout,
...
+void writeKeyboardfileData(FILE *out, char *rules, char *model, char *layout, char *variant, char *options)
 {
-  if (ruleslen)
+  /*
+    How to set "empty" values with setxkbmap, result of trial and error:
+    - for model and layout emtpy strings are accepted by setxkbmap.
+    - for rules not, setxkmap will fail if rules are an empty string
+    (will stop above already in that case)
+    - to set emtpy variant values the variant line must be omitted completely.
+    - to set options prepend them with "," to override instead of extend extings ones.
```
sunweaver: typo: s/extings ones/existing ones/

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1899,10 +1899,32 @@ static int nxagentXkbGetNames(char **rules, char **model, char **layout,
...
 }
 
-void nxagentWriteKeyboardFile(unsigned int ruleslen, char *rules, char *model, char *layout, char *variant, char *options)
+void writeKeyboardfileData(FILE *out, char *rules, char *model, char *layout, char *variant, char *options)
 {
-  if (ruleslen)
+  /*
+    How to set "empty" values with setxkbmap, result of trial and error:
+    - for model and layout emtpy strings are accepted by setxkbmap.
+    - for rules not, setxkmap will fail if rules are an empty string
```
sunweaver: gramma: s/if rules are/if the rules parameter is/ (or such)

sunweaver (changes requested):
@uli42: Very COOL! I found some typos here and there. Only cosmetics. However, they should be amended.

Thanks!
Mike

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
sunweaver: about the above commit. Leave it in... Don't file an extra PR.

uli42:
Ok, will fix that. Can you please test if it works for your setup?

Vom Smartphone gesendet.

Von: "Mike Gabriel" <notifications@github.com>
Gesendet: ‎24.‎07.‎2017 12:24
Cc: "Ulrich Sibiller" <uli42@gmx.de>; "Mention" <mention@noreply.github.com>
Betreff: Re: [ArcticaProject/nx-libs] Pr/improve keyboard (#492)

@sunweaver requested changes on this pull request.
@uli42: Very COOL! I found some typos here and there. Only cosmetics. However, they should be amended.
Thanks!
Mike



In nx-X11/programs/Xserver/hw/nxagent/Keyboard.c:
> @@ -909,6 +909,17 @@ N/A
 
         rules = nxagentXkbGetRules();
 
+        /*
+          from nxagent changelog:
+          2.0.22:
+          - Implented handling of value "query" for nxagentKbtype. This value
+          is passed by the NX client for MacOSX. If value of nxagentKbtype is
+          "query" or NULL we init keyboard by core protocol functions reading
+          the keyvoard mapping of the X server. The property _XKB_RULES_NAMES
+          is always set on the root window with default values of model and
typo: s/keyvoard/keyboard/



In nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1:
>  .TP 8
 .B \-reportprivatewids
-explicitly tell NX Agent to report X11 window IDs of internally created window objects to the session log (in machine readable form), so that external parsers can obtain that information from there; this creates a lot of output and may affect performance
+explicitly tell \fBnxagent\fR to report X11 window IDs of internally created window objects to the session log (in machine readable form), so that external parsers can obtain that information from there; this creates a lot of output and may affect performance
 .TP 8
It seems that this man page fix does not belong here. Please file as separate PR. Thanks!



In nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1:
> @@ -406,13 +406,19 @@ disable use of shared pixmaps
 enable use of shared pixmaps
 .TP 8
 .B \-noignore
-don't ignore pointer and keyboard configuration changes mandated by clients
+don't ignore pointer and keyboard configuration changes mandated by
+clients. Using this commands like disabling the keyboard bell (xset -b)
missing ",": Using this, commands like...



In nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1:
> @@ -497,6 +497,10 @@ XKEYBOARD will most likely not match the experienced settings. This is
 required on MacOS X to handle some keyboard problems that are special
 for this platform.
 .TP 8
+.I clone
+ask the real X server for the keyboard settings using XKEYBOARD protocol
+functions and clone them. This is the recommend setting.
typo: s/recommend/recommended/



In nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1:
> @@ -497,6 +497,10 @@ XKEYBOARD will most likely not match the experienced settings. This is
 required on MacOS X to handle some keyboard problems that are special
 for this platform.
 .TP 8
+.I clone
+ask the real X server for the keyboard settings using XKEYBOARD protocol
+functions and clone them. This is the recommend setting.
also in the commit msg of the "add keyboard=clone configuration option" commit:
s/Should the default but is/Should be the default but is/



In nx-X11/programs/Xserver/hw/nxagent/Keyboard.c:
> @@ -883,28 +883,28 @@ N/A
 
         XkbFreeKeyboard(xkb, XkbAllComponentsMask, True);
         xkb = NULL;
-	free(rules);
-	free(model);
-	free(layout);
-	free(variants);
-	free(options);
+        free(rules);
typo in commit message:
"Keyboard.c: Fix indention" -> "Keyboard.c: Fix indentation"



In nx-X11/programs/Xserver/hw/nxagent/Keyboard.c:
>  {
-  if (ruleslen)
+  /*
+    How to set "empty" values with setxkbmap, result of trial and error:
+    - for model and layout emtpy strings are accepted by setxkbmap.
+    - for rules not, setxkmap will fail if rules are an empty string
+    (will stop above already in that case)
+    - to set emtpy variant values the variant line must be omitted completely.
+    - to set options prepend them with "," to override instead of extend extings ones.
typo: s/extings ones/existing ones/



In nx-X11/programs/Xserver/hw/nxagent/Keyboard.c:
>  {
-  if (ruleslen)
+  /*
+    How to set "empty" values with setxkbmap, result of trial and error:
+    - for model and layout emtpy strings are accepted by setxkbmap.
+    - for rules not, setxkmap will fail if rules are an empty string
gramma: s/if rules are/if the rules parameter is/ (or such)

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: it's an 1:1 paste from the changelog. So I'll leave it as is.

sunweaver:
Ähm... 

On Monday, July 24, 2017, Ulrich Sibiller wrote:
> @uli42 commented on this pull request. 
> 
> In nx-X11/programs/Xserver/hw/nxagent/Keyboard.c:
> > @@ -909,6 +909,17 @@ N/A
>  
>          rules = nxagentXkbGetRules();
>  
> +        /*
> +          from nxagent changelog:
> +          2.0.22:
> +          - Implented handling of value "query" for nxagentKbtype. This value
> +          is passed by the NX client for MacOSX. If value of nxagentKbtype is
> +          "query" or NULL we init keyboard by core protocol functions reading
> +          the keyvoard mapping of the X server. The property _XKB_RULES_NAMES
> +          is always set on the root window with default values of model and
> it's an 1:1 paste from the changelog. So I'll leave it as is. 

Please let's not duplicate such flaws. Please fix this.

Mike

-- 

uli42:
Ok, ready for another round.

Please also comment on the following:

I'd like to change some things which would make nx 3.6 behave slighty different from 3.5:
1. making clone the default if the remote side has XKEYBOARD
2. making -xkblock the default (thus dropping the explicit option again)
3. drop "clone" and integrate that functionality into "query". use previous code path if clientos is Darwin/MacOS/OSX/applenameoftheday. Someone with an Apple system should check if clone work on those systems. If so, wie could drop that special support completely.
4. drop keycode conversion (evdev->pc105). I suppose this was needed because some older architectures did not know about evdev. nx used pc105 instead and mapped the keycodes internally. It was only activated when the client was linux (this check is long removed) and rules and/or model was evdev.

What do you think about those changes?

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -406,13 +406,19 @@ disable use of shared pixmaps
...
 .TP 8
 .B \-nokbreset
 don't reset keyboard device if the session is resumed
 .TP 8
 .B \-noxkblock
-always allow applications to change layout through XKEYBOARD
+this is only relevant if you also specify \-keyboard=query. In that
+case \fBnxagent\fR will lock the keyboard settings and clients will
+get an error when trying to change keyboard settings via
+XKEYBOARD. With \-noxkblock the lock is not applied and clients can
```
Ionic: Hmm... that's not universally true.

There's actually a very funny way to get rid of the lock: suspend and resume the session. As far as I could see, this resets the internal `nxagentKeyboard` variable to `NULL`. The next time `nxagentTuneXkbWrapper` will be executed, this lock is removed.

This behavior sounds more like a bug than an intended feature, though...

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -686,7 +685,7 @@ void nxagentChangeKeyboardControl(DeviceIntPtr pDev, KeybdCtrl *ctrl)
 
     value_mask = KBLed | KBLedMode;
 
-    for (i = 1; i <= 32; i++)
+    for (int i = 1; i <= 32; i++)
```
Ionic: If you change this, also remove `int i` at the top of this block.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1338,17 +1337,15 @@ int nxagentResetKeyboard(void)
               savedBellPercent, savedBellPitch, savedBellDuration);
   #endif
 
-  devBackup = malloc(sizeof(DeviceIntRec));
-
-  if (devBackup == NULL)
+  if (devBackup = malloc(sizeof(DeviceIntRec)))
```
Ionic: We should use `calloc` instead of `malloc` + `memset` here (and in general, too.)

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1799,14 +1793,14 @@ static int nxagentXkbGetNames(char **rules, char **model, char **layout,
                                   atom, 0, 256, 0, XA_STRING, &type, &format,
                                       &n, &after, (unsigned char **)&data);
 
-  if (result !=Success || data == NULL)
+  if (result != Success || data == NULL)
```
Ionic: `!data`?

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1863,7 +1857,7 @@ void nxagentKeycodeConversionSetup(void)
                                      &dvariant, &doptions);
 
   #ifdef DEBUG
-  if (drulesLen != 0 && drules != NULL && dmodel != NULL)
+  if (drulesLen != 0 && drules && dmodel)
```
Ionic: Just `drulesLen`?

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -482,8 +482,24 @@ set resizing support (default: true)
...
-.B keyboard=<string>
-set remote keyboard layout
+.B keyboard=<string> or kbtype=<string>
+
+.BR query | <model>/<layout>
+
+.RS 8
+.TP 8
+.I query
+ask the real X server for the keyboard settings using core protocol
```
Ionic: I don't like that explanation. It's not your fault, since you just copied what NoMachine published in their ChangeLog, but it's just not what "query" does.

From what I could gather (and unless I'm wrong), "query" does not clone or synchronize (or actually query) anything. It just sets the values to their defaults and disables some xkb settings (ironically only until a reconnect.)

Did you actually take a look at that? Do you have a different opinion?

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -616,8 +616,8 @@ void nxagentChangeKeyboardControl(DeviceIntPtr pDev, KeybdCtrl *ctrl)
 
   if (!noXkbExtension)
   {
-    xkbi = pDev -> key -> xkbInfo;
```
Ionic: Meh... I mean, theoretically I'm not against this commit, but
  - it makes only visual changes (which means that it adds another level of indirection in "git blame")
  - makes the rest of the code look inconsistent with other NoMachine code
  - also touches other parts not mentioned like compacting !=/== NULL and adding whitespace between equals sign chars.

Not sure if that's a smart move.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1337,11 +1337,7 @@ int nxagentResetKeyboard(void)
               savedBellPercent, savedBellPitch, savedBellDuration);
   #endif
 
-  if (devBackup = malloc(sizeof(DeviceIntRec)))
```
Ionic: I'd squash that into the "Keyboard.c: Some code cleanups" commit.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1858,86 +1860,93 @@ static int nxagentXkbGetNames(char **rules, char **model, char **layout,
...
+
+      if ((keyboard_file = fopen(keyboard_file_path, "w")))
+      {
+        if (rules)
+          fprintf(keyboard_file, "rules=\"%s\"\n", rules[0] == '\0' ? "," : rules);
+        if (model)
+          fprintf(keyboard_file, "model=\"%s\"\n", model[0] == '\0' ? "," : model);
+        if (layout)
+          fprintf(keyboard_file, "layout=\"%s\"\n", layout[0] == '\0' ? "," : layout);
+        /* FIXME: this is not correct. We need to match the number of
```
Ionic: This is generally a bit wonky, yeah. I don't think rules and model can take multiple values and (for that matter), that a single comma in there resets anything, but layout, variant and options do can take multiple values. layout and variant should be "synced" in terms of matching entries, while the list of options is global to all layouts (and variants).

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -90,6 +90,8 @@ static int nxagentXkbGetNames(char **rules, char **model, char **layout,
 
 static void nxagentKeycodeConversionSetup(void);
 
+void nxagentWriteKeyboardFile(unsigned int ruleslen, char *rules, char *model, char *layout, char *variant, char *options);
```
Ionic: Make it static.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1004,9 +1004,39 @@ N/A
           goto XkbError;
         }
 
-        xkb = XkbGetKeyboard(nxagentDisplay, XkbGBN_AllComponentsMask, XkbUseCoreKbd);
+        {
+          char *drules = NULL;
```
Ionic, line 1057: Do you really think this makes sense? We're now duplicating code for the first part of the function in two locations.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1969,7 +1973,21 @@ void nxagentResetKeycodeConversion(void)
...
+    char *dmodel = NULL;
+    char *dlayout = NULL;
+    char *dvariant = NULL;
+    char *doptions = NULL;
+    unsigned int drulesLen;
+
+    drulesLen = nxagentXkbGetNames(&drules, &dmodel, &dlayout,
+				   &dvariant, &doptions);
+
+    if (drulesLen && drules && dmodel)
```
Ionic: Why are you guarding this here, but not above? See why that duplication might prove unfortunate?

The new `nxagentKeycodeConversionSetup()` function can handle NULL rules and model values just fine.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1969,7 +1973,21 @@ void nxagentResetKeycodeConversion(void)
...
+      nxagentKeycodeConversionSetup(drules, dmodel);
+
+    if (drules)
```
Ionic, line 2049: Just free, don't check for NULL. Calling `free()` on NULL pointers is fine (we just have to be careful not to double-free stuff, so a free should be followed by setting the pointer to NULL. We don't need that here, since we don't intend to use `drules` later on.)

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1004,9 +1004,39 @@ N/A
...
+          {
+            fprintf(stderr, "%s: Failed to retrieve remote rules.\n", __func__);
+          }
+          #endif
 
-        nxagentKeycodeConversionSetup();
+          nxagentWriteKeyboardFile(drulesLen, drules, dmodel, dlayout, dvariant, doptions);
+          nxagentKeycodeConversionSetup(drules, dmodel);
+
+          if (drules)
```
Ionic: Also here, just free it unconditionally.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1969,7 +1973,21 @@ void nxagentResetKeycodeConversion(void)
...
+    char *drules = NULL;
+    char *dmodel = NULL;
+    char *dlayout = NULL;
+    char *dvariant = NULL;
+    char *doptions = NULL;
+    unsigned int drulesLen;
+
+    drulesLen = nxagentXkbGetNames(&drules, &dmodel, &dlayout,
+				   &dvariant, &doptions);
+
```
Ionic: Not calling/updating the keyboard file via `nxagentWriteKeyboardFile()` - which was the case previously.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -703,6 +703,30 @@ void nxagentChangeKeyboardControl(DeviceIntPtr pDev, KeybdCtrl *ctrl)
   #endif
 }
 
+/*
+  update a string that has been allocated with a new string. If the
+  new string is NULL nothing will be changed
+*/
+
+char * update_string(char ** current, char *new, int offset)
```
Ionic: Missing static + missing definition at the start of this file.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -860,16 +883,11 @@ N/A
...
-        {
-          free_model = 0;
-          free(model);
-        }
-        if (free_layout)
-        {
-          free_layout = 0;
-          free(layout);
-        }
+	free(rules);
```
Ionic: Tabs instead of spaces here...

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
Ionic: "offset" is a very bad name. That's a length, not an offset.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -703,6 +703,30 @@ void nxagentChangeKeyboardControl(DeviceIntPtr pDev, KeybdCtrl *ctrl)
...
 
+/*
+  update a string that has been allocated with a new string. If the
+  new string is NULL nothing will be changed
+*/
+
+char * update_string(char ** current, char *new, int offset)
+{
+  if (!new)
+    return *current;
```
Ionic: And if `current` is NULL, stuff goes boom.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
Ionic: Also, why returning something if you're not using the return value anyway?

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -932,11 +949,8 @@ N/A
             goto XkbError;
           }
 
-          free_model = 1;
-	  model = strndup(nxagentKeyboard, i);
-
-          free_layout = 1;
-	  layout = strdup(&nxagentKeyboard[i + 1]);
+	  update_string(&model, nxagentKeyboard, i);
```
Ionic: Tabs instead of spaces.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1101,30 +1115,11 @@ N/A
...
-              free_layout = 0;
-              free(layout);
-            }
-            layout = config.layout;
-          }
-          if (config.variant)
-            variants = config.variant;
-          if (config.options)
-            options = config.options;
+	  update_string(&rules, config.rules_file, 0);
```
Ionic: More tabs.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1176,17 +1171,11 @@ N/A
...
-          free_model = 0;
-          free(model);
-        }
-
-        if (free_layout)
-        {
-          free_layout = 0;
-          free(layout);
-        }
+	free(rules);
```
Ionic: And more tabs.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
Ionic: Follow up these `free()` calls with setting the pointers to NULL, or we risk double-frees. Double-fries would be cool, double-frees not so much.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
Ionic: Follow up with setting pointers to NULL.

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
Ionic: Ping.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -923,6 +923,14 @@ N/A
         fprintf(stderr, "nxagentKeyboardProc: nxagentKeyboard is [%s].\n", nxagentKeyboard ? nxagentKeyboard : "NULL");
         #endif
 
+        if (nxagentKeyboard && (strcmp(nxagentKeyboard, "null/null") == 0))
+        {
+          #ifdef TEST
+          fprintf(stderr, "nxagentKeyboardProc: changing nxagentKeyboard from [null/null] to [clone].\n");
+          #endif
+          free(nxagentKeyboard);
+          nxagentKeyboard = strdup("clone");
```
Ionic, line 932: `update_string(&nxagentKeyboard, "clone", 0);` instead of `free` and `strdup`?

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -937,7 +945,9 @@ N/A
           layout.
         */
 
-        if (nxagentKeyboard && (strcmp(nxagentKeyboard, "query") != 0) && (strcmp(nxagentKeyboard, "clone") != 0))
+        if (nxagentKeyboard &&
```
Ionic: This change should have been squashed into a previous one.

nx-X11/programs/Xserver/hw/nxagent/Reconnect.c
```diff
@@ -456,6 +456,16 @@ Bool nxagentReconnectSession(void)
 
   nxagentProcessOptionsFile();
 
+  if (nxagentKeyboard && (strcmp(nxagentKeyboard, "null/null") == 0))
+  {
+    #ifdef TEST
+    fprintf(stderr, "nxagentReconnect: changing nxagentKeyboard from [null/null] to [clone].\n");
+    #endif
+
+    free(nxagentKeyboard);
```
Ionic, line 467: `update_string`

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -498,12 +498,16 @@ required on MacOS X to handle some keyboard problems that are special
 for this platform.
 .TP 8
 .I clone
-ask the real X server for the keyboard settings using XKEYBOARD protocol
-functions and clone them. This is the recommend setting.
+ask the real X server for the keyboard settings using XKEYBOARD
```
Ionic: This part should be squashed into a previous commit.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1678,7 +1678,7 @@ void nxagentInitXkbWrapper(void)
   fprintf(stderr, "nxagentInitXkbWrapper: Called.\n");
   #endif
 
-  if (!nxagentOption(InhibitXkb))
+  if (nxagentOption(InhibitXkb) == 0)
```
Ionic, line 1701: Keep it `!nxagentOption(InhibitXkb)`.

nx-X11/programs/Xserver/hw/nxagent/Options.h
```diff
@@ -384,6 +384,9 @@ typedef struct _AgentOptions
 
   /*
    * Inhibit some XKEYBOARD requests.
+   * 0 = allow all XKEYBOARD requests
```
Ionic: No, definitely not "all". Only "some".

nx-X11/programs/Xserver/hw/nxagent/Options.h
```diff
@@ -384,6 +384,9 @@ typedef struct _AgentOptions
...
+   * 1 = forbid all XKEYBOARD requests if keyboard=clone or keyboard=query
```
Ionic: Also wrong, this is actually the default - regardless of the `keyboard` setting.

Ionic: I wonder if `InhibitXkb` is actually doing what it's supposed to do? We should see denied change requests a lot more often if it were working correctly.

Ionic: Ah, and it makes sense: since the XKB extension is enabled when a client (re-)connects (and our first connection is also a reconnection for that matter), this will always be the case and disabling the XKB extension pretty much useless.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1919,25 +1941,15 @@ void nxagentWriteKeyboardFile(unsigned int ruleslen, char *rules, char *model, c
...
-        if (layout)
-          fprintf(keyboard_file, "layout=\"%s\"\n", layout[0] == '\0' ? "," : layout);
-        /* FIXME: this is not correct. We need to match the number of
-           comma separated values between variant and layout */
-        if (variant)
-          fprintf(keyboard_file, "variant=\"%s\"\n", variant[0] == '\0' ? "," : variant);
-        if (options)
-          fprintf(keyboard_file, "options=\"%s\"\n", options[0] == '\0' ? "," : options);
-        fclose(keyboard_file);
+	writeKeyboardfileData(keyboard_file, rules, model, layout, variant, options);
```
Ionic: Tabs.

Ionic (changes requested):
>1. making clone the default if the remote side has XKEYBOARD

Sounds sane, would currently always be overridden in X2Go though.
 
> 2. making -xkblock the default (thus dropping the explicit option again)

It is the default. It's just not working correctly.

> 3. drop "clone" and integrate that functionality into "query". use previous code path if clientos is Darwin/MacOS/OSX/applenameoftheday. Someone with an Apple system should check if clone work on those systems. If so, wie could drop that special support completely.

No, that won't work. OS X keyboard value is a generic one and does not really reflect the type and layout. We need special handling there.

> 4. drop keycode conversion (evdev->pc105). I suppose this was needed because some older architectures did not know about evdev. nx used pc105 instead and mapped the keycodes internally. It was only activated when the client was linux (this check is long removed) and rules and/or model was evdev.

Sounds sane, but let's not do that now. We should keep that in mind/as a separate bug report.

uli42:
>> making clone the default if the remote side has XKEYBOARD

> Sounds sane, would currently always be overridden in X2Go though.

Not competely: x2go passes "null/null" and nxagent treats that as "clone". However, later on a setxkbmap call is made, which will only work _if_ the xkb stuff is _not_ locked.

>> making -xkblock the default (thus dropping the explicit option again)

> It is the default. It's just not working correctly.

It is only working if keyboard="query" . 

I do not believe that this code is broken because I have been using nx for more than 10 years, beginngin with 1.5, and the keyboard has _never_ been locked. So that feature would have been broken from the moment it got implemented. Which is highly improbable IMHO.

>> drop "clone" and integrate that functionality into "query". use previous code path if clientos is
>> Darwin/MacOS/OSX/applenameoftheday. Someone with an Apple system should check if clone work on
>> those systems. If so, wie could drop that special support completely.

> No, that won't work. OS X keyboard value is a generic one and does not really reflect the
> type and layout. We need special handling there.

That's the reason I want to keep the current code for darwin clients (only). So nothing would break. Wat am I missing here?

>> drop keycode conversion (evdev->pc105). I suppose this was needed because
>> some older architectures did not know about evdev. nx used pc105 instead and
>> mapped the keycodes internally. It was only activated when the client was linux
>> (this check is long removed) and rules and/or model was evdev.

> Sounds sane, but let's not do that now. We should keep that in mind/as a separate bug report.

ok, I'll open one.

I'll have a look at your other comments now.

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
uli42: I did not succeed with query. And I did not fully get the intentions of that code. So I assumed "query" is a value only being used by MacOS clients and only working there (you have told me that MacOs in special in that respect). I do not have (or know) MacOs, so I could not test further. But my tests so far revealed that xkblock is active only when query is used.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: OK, I have reverted it (but I won't use ` -> ` for my own code)

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: fixed

uli42: right, fixed

uli42: well, I had other plans when I created that function. Removed.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: fixed

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: man XFree says "A NULL pointer cannot be passed to this function."

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: man XFree says "A NULL pointer cannot be passed to this function."

nx-X11/programs/Xserver/hw/nxagent/Reconnect.c
uli42: update_string is local to Keyboard.c.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: done

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: man Xfree says "A NULL pointer cannot be passed to this function."

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: done

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: Hmm, I can't see the `int i` in this function, so what are you reffering to?

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: done

nx-X11/programs/Xserver/hw/nxagent/Options.h
uli42: Again I am not sure what you mean. Have look at nxagentInitXkbWrapper().

If `InihibitXkb` is `0` no own filter wrapper is installed. Which effectively means that all requests are accepted.

If `InihibitXkb` is 1 (the default) all _writing_ XKEYBOARD requests are silently ignored, the others are accepted. 

If `InihibitXkb` is 2 the behaviour is like 1 but only if neither query nor clone are used. If so, the behaviour is like 0.

So my description is wrong, I have improved that.

However, I have just tested again and found that 1 is only working for clone and query (as I wrote in the code) despite being the default. This needs further investigation.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
Ionic: Uhm there definitely is one, or how else would `i = 1` work? I've seen it at the start of the scope.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: no, I don't want to use update_string() outside Keyboard.c. Reconnect.c has very similar code. Soi I prefer to let it this way. Besides, nxagentKeyboard is a global variable and update_string ist only using variables local to Keyboard.c

uli42:
Hmm, just came to my mind: if xkblock is active there's no reason to create teh keyboard file as one won't be able to use it...

sunweaver:
@uli42: where are we with #492?

uli42:
@sunweaver Well, Ionic had some concerns (regarding code duplication, xkblockingm keycode conversion) we have not discussed finally. Further I wanted to combine the ```keyboard=clone``` mode with automatically adapting keyboard settings when they change on the client side. But I have found that this is not easily doable because you cannot simply get an event if that happens. This needs more research.

Also missing is a way for x2go (or other nx clients) to set the keyboard with activated keyboard locking. I have something in mind like calling setxkbmap first with some magic value and then again with the desired values. But this is not implemented yet.

So I'd say most of the current changes could (and should) be integrated, but no defaults should be changed. The ```keyboard=null/null``` synonym for ```keyboard=clone``` should be dropped.

sunweaver:
Time to revisit this PR, I guess... xkb-1.3.0.0 branch has been merged.

sunweaver:
Obsoleted by #743, thus closing.

PR #499, opened 2017-08-03
nxcomp/configure.ac: Don't rely on pkg-config alone when it comes to …

sunweaver:
…testing for presence of libjpeg shared lib and header files.

 Especially systems still using the IJG's libjpeg implementation
 are likely to lack the libjpeg.pc file.
 .
 So we add some alternative way of detecting if all libjpeg related
 build-dependencies are in place.

This fixes nx-libs builds on Ubuntu trusty:
https://jenkins.arctica-project.org/job/nx-libs.deb+nightly/target=trusty/272/

### DIFF
--- a/nxcomp/configure.ac
+++ b/nxcomp/configure.ac
@@ -25,10 +25,6 @@ AC_SUBST([COMP_VERSION])
 LT_COMP_VERSION=[`echo $COMP_VERSION | sed -r -e 's/^([0-9]+\.[0-9]+\.[0-9]+).*$/\1/' -e 's/\./:/g'`] 
 AC_SUBST([LT_COMP_VERSION])
 
-PKG_CHECK_MODULES(JPEG, libjpeg)
-PKG_CHECK_MODULES(PNG, libpng)
-PKG_CHECK_MODULES(Z, zlib)
-
 # Upstream's pkg.m4 (since 0.27) offers this now, but define our own
 # compatible version in case the local version of pkgconfig isn't new enough.
 # https://bugs.freedesktop.org/show_bug.cgi?id=48743
@@ -38,6 +34,10 @@ m4_ifdef([PKG_INSTALLDIR], [PKG_INSTALLDIR],
                       [install directory for nxcompshad.pc pkg-config file])],
                        [],[with_pkgconfigdir='$(libdir)/pkgconfig'])
           AC_SUBST([pkgconfigdir], [${with_pkgconfigdir}])])
+
+PKG_CHECK_MODULES([JPEG], [libjpeg], [], [LIBJPEG_FALLBACK_CHECK])
+PKG_CHECK_MODULES([PNG], [libpng], [], [])
+PKG_CHECK_MODULES([Z], [zlib], [], [])
 
 AC_LANG([C++])
 NX_COMPILER_BRAND
--- a/m4/nx-macros.m4
+++ b/m4/nx-macros.m4
@@ -303,3 +303,21 @@ rm -f conftest*])
 FreeBSD=
 test "$nxconf_cv_freebsd" = yes && FreeBSD=yes
 ]) # NX_BUILD_ON_FreeBSD
+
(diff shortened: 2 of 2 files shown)


sunweaver:
@Ionic: we should get this one in first... (#499, FTBFS on Ubuntu trusty and other older Linux distros).

nxcomp/configure.ac
```diff
@@ -39,6 +35,43 @@ m4_ifdef([PKG_INSTALLDIR], [PKG_INSTALLDIR],
                        [],[with_pkgconfigdir='$(libdir)/pkgconfig'])
           AC_SUBST([pkgconfigdir], [${with_pkgconfigdir}])])
 
+AC_PATH_PROG(PKGCONF, pkg-config)
```
Ionic: I don't like that...

We should keep using `PKG_CHECK_MODULES(JPEG, libjpeg)` (and actually "convert" this and all the other calls to `PKG_CHECK_MODULES([...], [...])` instead).

`PKG_CHECK_MODULES()` has the following declaration:
> `PKG_CHECK_MODULES(VARIABLE-PREFIX, MODULES, [ACTION-IF-FOUND], [ACTION-IF-NOT-FOUND])`

The default `ACTION-IF-NOT-FOUND` action will break out with an error, but we can also provide a special function to handle that case (i.e., in our case, the code in the `else` branch).

```m4
AC_DEFUN([LIBJPEG_CHECK],[
AC_MSG_CHECKING([for libjpeg shared libary file and headers])
AC_CHECK_LIB([jpeg], [jpeg_destroy_compress],
	[have_jpeg_lib=yes], [have_jpeg_lib=no])
AC_CHECK_HEADERS([jpeglib.h],
	[have_jpeg_headers=yes], [have_jpeg_headers=no])

if test x"$have_jpeg_lib" = "xyes" && test x"$have_jpeg_headers" = "xyes"; then
	AC_MSG_RESULT([yes])
	JPEG_CFLAGS=""
	JPEG_LIBS="-ljpeg"
else
	AC_MSG_RESULT([no])
	AC_MSG_FAILURE([Could not find libjpeg on your system, make sure
the JPEG shared library and header files are installed.])
fi
])

PKG_CHECK_MODULES([JPEG], [libjpeg], [], [LIBJPEG_CHECK])
```

I'd put the `LIBJPEG_CHECK` function in `nx-macros.m4`.

Please test that, though. Both on a system that has pkgconfig files for libjpeg and on a system which doesn't have them. Pay close attention to the output. I don't know if and how nested `AC_MSG_CHECKING`/`AC_MSG_RESULT` work out.

Ionic (changes requested):

nxcomp/configure.ac
sunweaver: @Ionic: wow, this works like sharm. Tested with three occasions:

  * no libjpeg-dev installed at all
  * libjpeg-dev installed (including .pc file)
  * libjpeg-dev installed, but .pc file renamed

sunweaver:
@Ionic: your turn again...

nxcomp/configure.ac
```diff
@@ -39,6 +35,10 @@ m4_ifdef([PKG_INSTALLDIR], [PKG_INSTALLDIR],
                        [],[with_pkgconfigdir='$(libdir)/pkgconfig'])
           AC_SUBST([pkgconfigdir], [${with_pkgconfigdir}])])
 
+PKG_CHECK_MODULES([JPEG], [libjpeg], [], [LIBJPEG_FALLBACK_CHECK])
+PKG_CHECK_MODULES(PNG, libpng)
```
Ionic: If touching these lines anyway, make that `PKG_CHECK_MODULES([...], [...])` too, please.

sunweaver:
@Ionic: go ahead, please (#499)

Ionic (approved):

Ionic:
With cleanups in 8b88289eea2d9325465fd4517e08118d2affcb7d and 8235925a2c3599a8980ce789e0a79369118727fe.

Issue #500, opened 2017-08-04
add property with agent version


uli42:
This way you can identify the version from any client. The property can also be used to check if we are running in an nx session. It can easily be checked using xprop --root

PR #501, opened 2017-08-09
Set NX_AGENT_VERSION property for root window

uli42:
Fixes ArcticaProject/nx-libs#500


sunweaver (approved):
LGTM. Please rebase the PR update branch.

uli42:
rebased, but... Thinking about my change I am unsure now if this code is correct. The plan was to set that property only for the root window. And this window is identified in CreateWindow by checking if it has a parent.  My tests showed that the property is indeed only showing on the root window. But what about non-reparenting window managers? Does it work with them, too?

uli42:
ok, new version that does the same job, but much better: it is now a new function called from InitRootWindow(), not from CreateWindow().

PR #508, opened 2017-08-31
nxcomp/src/{Loop,Proxy}.cpp: On Debian/kFreeBSD the sun_path property…

sunweaver:
… is 104 chars long, not 108.

 Fixes ArcticaProject/nx-libs#507.

### DIFF
--- a/nxcomp/src/Loop.cpp
+++ b/nxcomp/src/Loop.cpp
@@ -3810,7 +3810,15 @@ int SetupAuthInstance()
 
         launchdAddrUnix.sun_family = AF_UNIX;
 
+        #ifdef __linux__
         const int launchdAddrNameLength = 108;
+        #else
+        /* POSIX/SUS does not specify a length.
+         * BSD derivatives generally support 104 bytes, other systems may be more constrained.
+         * If you happen to run into such systems, extend this section with the appropriate limit.
+         */
+        const int launchdAddrNameLength = 104;
+        #endif
 
         int success = -1;
 
--- a/nxcomp/src/Proxy.cpp
+++ b/nxcomp/src/Proxy.cpp
@@ -6294,7 +6294,15 @@ int Proxy::handleNewGenericConnectionFromProxyUnix(int channelId, T_channel_type
 
   serverAddrUnix.sun_family = AF_UNIX;
 
+  #ifdef __linux__
   const int serverAddrNameLength = 108;
+  #else
+  /* POSIX/SUS does not specify a length.
+   * BSD derivatives generally support 104 bytes, other systems may be more constrained.
+   * If you happen to run into such systems, extend this section with the appropriate limit.
+   */
+  const int serverAddrNameLength = 104;
+  #endif
 
   strncpy(serverAddrUnix.sun_path, path, serverAddrNameLength);
 


sunweaver:
@Ionic: patch is tested on a Debian/kFreeBSD (amd64) machine. Any comments on style and such?

sunweaver:
Compiler pre-defined builtins taken from https://sourceforge.net/p/predef/wiki/OperatingSystems/

nxcomp/src/Loop.cpp
```diff
@@ -3810,7 +3810,11 @@ int SetupAuthInstance()
 
         launchdAddrUnix.sun_family = AF_UNIX;
 
+#if defined(__FreeBSD_kernel__) && defined(__GLIBC__)
```
Ionic: Hardcode this and the other definition to `104`. All BSD variants I could find (FreeBSD, NetBSD, OpenBSD, Dragonfly) and even OS X use a size of `104`.

Debian/kFreeBSD is a bit different in that it also defines `sun_path` as `104` chars long, but appends another internal `4` chars wide field for padding/Linux compatibility. Despite that, the `sun_path` comment claims that the kernel can only handle `104` chars.

Ionic (changes requested):

nxcomp/src/Loop.cpp
Ionic: I guess NoMachine just looked up the structure on their Linux-based development machines when writing that code and never actually bothered to check the target OS headers...

sunweaver: @Ionic: the code passage in discussion comes from me, not from NoMachine.

I will look at the code and possibly to the hard-coding to 104 chars.

sunweaver: @Ionic: oh, actually only the code in Proxy.cpp comes from me, the code in Loop.cpp is NoMachine'ish, I guess...

sunweaver:
@Ionic: next iteration on #508

nxcomp/src/Loop.cpp
Ionic: No, not really. According to `git blame`, both hunks came from NoMachine.

nxcomp/src/Proxy.cpp
```diff
@@ -6294,7 +6294,11 @@ int Proxy::handleNewGenericConnectionFromProxyUnix(int channelId, T_channel_type
 
   serverAddrUnix.sun_family = AF_UNIX;
 
-  const int serverAddrNameLength = 108;
+  // Some OSses define sockaddr_un.sun_path with a string length of 108 chars
+  // some to string length of 104 chars.
+  //
+  // Let's make them all happy by using the shorter string size...
+  const int serverAddrNameLength = 104;
```
Ionic: Good that I take another look at this. I totally missed that this section is for *general* UNIX sockets, not (just) launchd ones.

Let's better do

```C++
#ifdef __linux__
 ... = 108;
#else
  /* POSIX/SUS does not specify a length.
   * BSD derivatives generally support 104 bytes, other systems may be more constrained.
   * If you happen to run into such systems, extend this section with the appropriate limit.
   */
 ... = 104;
#endif

**Review by Ionic — CHANGES_REQUESTED** [review]

**Comment by sunweaver** [comment]
@Ionic: ok. Next round... Valid feedback, implemented as requested.

File: nxcomp/src/Loop.cpp [hunk 3]
```diff
@@ -3810,7 +3810,15 @@ int SetupAuthInstance()
 
         launchdAddrUnix.sun_family = AF_UNIX;
 
+        #ifdef __linux__
```
Ionic, line 3813: We can hardcode this to 104 here without any additional checks or comments, since this code section will only be used on systems that have launchd (which really is just OS X and FreeBSD.)

I don't care though, can leave this as-is or amend.

sunweaver:
Leave as is...
Mike

On Wednesday, September 6, 2017, Mihai Moldovan wrote:
> @Ionic commented on this pull request. 
> 
> In nxcomp/src/Loop.cpp:
> > @@ -3810,7 +3810,15 @@ int SetupAuthInstance()
>  
>          launchdAddrUnix.sun_family = AF_UNIX;
>  
> +        #ifdef __linux__
> We can hardcode this to 104 here without any additional checks or comments, since this code section will only be used on systems that have launchd (which really is just OS X and FreeBSD.)
> I don't care though, can leave this as-is or amend. 
> —
>

-- 

Ionic (approved):

PR #516, opened 2017-09-23
spec file fixes

mikedep333:
Multiple fixes to make the RPM .spec file work with the version of nx-libs.

### DIFF
--- a/nx-libs.spec
+++ b/nx-libs.spec
@@ -444,16 +444,18 @@ rm -r %{buildroot}%{_includedir}/nx-X11/Xtrans
 %dir %{_libdir}/nx
 %dir %{_datadir}/nx
 %{_datadir}/nx/SecurityPolicy
-%{_datadir}/nx/XErrorDB
-%{_datadir}/nx/Xcms.txt
+%{_datadir}/nx/X11/XErrorDB
+%{_datadir}/nx/X11/Xcms.txt
 
 %files -n libNX_X11-6
 %defattr(-,root,root)
 %{_libdir}/libNX_X11.so.6*
+%{_libdir}/libNX_X11.la
 
 %files -n libNX_X11-devel
 %defattr(-,root,root)
 %{_libdir}/libNX_X11.so
+%{_libdir}/pkgconfig/nx-x11.pc
 %dir %{_includedir}/nx
 %dir %{_includedir}/nx-X11
 %{_includedir}/nx-X11/ImUtil.h
@@ -486,6 +488,7 @@ rm -r %{buildroot}%{_includedir}/nx-X11/Xtrans
 %doc doc/nxcomp/README.on-retroactive-DXPC-license
 %doc doc/nxcomp/nxcomp-3.6-drops-compat-code-3.4.x-testing.pdf
 %{_libdir}/libXcomp.so.3*
+%{_libdir}/libXcomp.la
 
 %files -n libXcompshad-devel
 %defattr(-,root,root)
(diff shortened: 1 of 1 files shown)


sunweaver (approved):
LGTM. Anything else needed for this PR or can I merge?

THANKS!
Mike

nx-libs.spec
```diff
@@ -444,16 +444,18 @@ rm -r %{buildroot}%{_includedir}/nx-X11/Xtrans
...
 %{_datadir}/nx/SecurityPolicy
-%{_datadir}/nx/XErrorDB
-%{_datadir}/nx/Xcms.txt
+%{_datadir}/nx/X11/XErrorDB
+%{_datadir}/nx/X11/Xcms.txt
 
 %files -n libNX_X11-6
 %defattr(-,root,root)
 %{_libdir}/libNX_X11.so.6*
+%{_libdir}/libNX_X11.la
```
Ionic, line 456: I'm not sure we want to actually ship `.la` files. Debian and other distributions I know just delete them. Is Fedora/RHEL handling that differently?

nx-libs.spec
```diff
@@ -486,6 +488,7 @@ rm -r %{buildroot}%{_includedir}/nx-X11/Xtrans
 %doc doc/nxcomp/README.on-retroactive-DXPC-license
 %doc doc/nxcomp/nxcomp-3.6-drops-compat-code-3.4.x-testing.pdf
 %{_libdir}/libXcomp.so.3*
+%{_libdir}/libXcomp.la
```
Ionic, line 494: `.la`

nx-libs.spec
```diff
@@ -497,6 +500,7 @@ rm -r %{buildroot}%{_includedir}/nx-X11/Xtrans
 %defattr(-,root,root)
 %doc COPYING
 %{_libdir}/libXcompshad.so.3*
+%{_libdir}/libXcompshad.la
```
Ionic, line 506: `.la`

nx-libs.spec
```diff
@@ -586,7 +590,7 @@ rm -r %{buildroot}%{_includedir}/nx-X11/Xtrans
 %{_bindir}/nxproxy
 %{_datadir}/man/man1/nxproxy.1*
 %dir %{_libdir}/nx/bin
```
Ionic, line 595: Better change this to `%{_libexecdir}` as well, I guess.

nx-libs.spec
```diff
@@ -1,4 +1,7 @@
 %global _hardened_build 1
+# Put the nxproxy binary in the same dir as nxagent.
+# Fixes the /usr/bin/nxproxy script not finding it.
+%define _libexecdir %{_libdir}/nx/bin
```
Ionic, line 4: Merge this commit into the previous one, otherwise we'll get needless inconsistencies between commits.

Ionic (changes requested):
Some requests for changes, some questions.

nx-libs.spec
Ionic: More information: according to the [Fedora Packaging Guidelines](https://fedoraproject.org/wiki/Packaging:Guidelines#Packaging_Static_Libraries), `.la` files should be deleted and not shipped - static libraries should be avoided at all costs.

The better handling would probably thus to be deleting these files in the `%install` section like we do in `debian/rules`.

sunweaver:
@mikedep333: ping on the requested changes for nx-libs PR #516. Thanks!

PR #521, opened 2017-10-09
nxcomp/src/Loop.cpp: Reduce control->TokenSize slightly again in orde…

sunweaver:
…r to fix fragmentation of packages when on VPN (or otherwise encapsulated) connections.

### DIFF
--- a/nxcomp/src/Loop.cpp
+++ b/nxcomp/src/Loop.cpp
@@ -12937,7 +12937,7 @@ int SetLinkAdsl()
 
   control -> LinkMode = LINK_TYPE_ADSL;
 
-  control -> TokenSize  = 1536;
+  control -> TokenSize  = 1408;
   control -> TokenLimit = 24;
 
   control -> SplitMode             = 1;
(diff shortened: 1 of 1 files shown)


sunweaver:
@Ionic: please take notice of this PR. It fixes slloooowwwness of nx sessions, when connecting through VPN tunnels. The will slightly reduce SKIA or Cairo based rendering performance again slightly, but the benefit when connecting via VPN is worth the shot.

uli42:
Looks to me that we need a dynamic mode here that determines the
optimal value somehow.

sunweaver:
Hi Uli,

On  Mo 09 Okt 2017 11:56:22 CEST, Ulrich Sibiller wrote:

> Looks to me that we need a dynamic mode here that determines the
> optimal value somehow.

We need the possibility to tweak those values at runtime and  
externally. The nxcomp code should not be overloaded with connection  
checks. This should be done via an external tool that then modifies  
the nxcomp settings on-the-fly.

So for now, this fix should go in IMHO (I am also suffering from this  
flaw when using nested VPN tunnels over ADSL).

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

Ionic (approved):

realsimix:
Thanks for the commit!

PR #522, opened 2017-10-09
nxagent: Auto-detect client-side DPI at session startup.

sunweaver:


sunweaver:
@uli42: Would you agree with accepting this PR to the 3.6.x branch?

uli42:
Well, in principle I'd agree. However: I am not really sure if that is working properly for people reconnecting from clients with different DPI settings. Plus: We should also ensure that changing the screen size via xrandr will set/keep the DPI correctly. I have seen cases where the 3.5.0 code calculated wrong DPI after resizing.

uli42:
Another question: why is the DPI code duplicated in two files? Can't we unify that somehow?

nx-X11/programs/Xserver/hw/nxagent/Screen.c
```diff
@@ -209,7 +209,7 @@ void nxagentShadowAdaptDepth(unsigned int, unsigned int, unsigned int, char **);
 
 RegionRec nxagentShadowUpdateRegion;
 
-#define NXAGENT_DEFAULT_DPI 75
+#define NXAGENT_DEFAULT_DPI -1
```
uli42: So we are losing the previous default of 75 which has been the default for years now. With that PR it will default to the value of the client's X server, so that will surely make a difference for many people. What about leaving the 75 as default and activate autodpi separately, e.g. via `-dpi -1` (0 is not working, since that is the internal default of `os/utils.c`)

uli42 (commented):
Looks (and works) well, AFAIKS. However, it will not adapt the DPI on reconnect. And I am unsure if we want that, because it will only affect newly started apps, e.g. konsole (but not xterm). So if the dpi changes on reconnect of a session new windows within the session will look different from the older ones. If we do not want that: Is cloning the initial dpi any benefit then?

sunweaver:
Hi,

On  Mo 09 Okt 2017 23:06:18 CEST, Ulrich Sibiller wrote:

>
> Looks (and works) well, AFAIKS. However, it will not adapt the DPI  
> on reconnect. And I am unsure if we want that, because it will only  
> affect newly started apps, e.g. konsole (but not xterm). So if the  
> dpi changes on reconnect of a session new windows within the session  
> will look different from the older ones. If we do not want that: Is  
> cloning the initial dpi any benefit then?

We don't want changing DPI values on reconnects, indeed. The idea is  
that nx session and local Xserver use the same DPI and this gets  
auto-detected on session startup.

>> @@ -209,7 +209,7 @@ void nxagentShadowAdaptDepth(unsigned int,  
>> unsigned int, unsigned int, char **);
>
>  RegionRec nxagentShadowUpdateRegion;
>
> -#define NXAGENT_DEFAULT_DPI 75
> +#define NXAGENT_DEFAULT_DPI -1
>
> So we are losing the previous default of 75 which has been the  
> default for years now. With that PR it will default to the value of  
> the client's X server, so that will surely make a difference for  
> many people. What about leaving the 75 as default and activate  
> autodpi separately, e.g. via `-dpi -1` (0 is not working, since that  
> is the internal default of `os/utils.c`)

I'd prefer a separate cmdline option: -autodpi.

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

uli42:
Turns out that x2goclient already passes the client side dpi to nxagent (if
the user does not force dpi in the settings GUI).  So this PR will not
change anything for x2goclient users. For pyhoca-gui it will change
behaviour but then be identical to x2goclient's behaviour.

realsimix:
I have to clarify one important thing: in our environment we don't run persistent sessions so we don't do reconnects.

I liked the "autodpi" feature and therefore got it back when NoMachine changed the code. On the other hand we could easily read the DPI settings of the local display and use -dpi <$X> instead.

A separate cmdline option -autodpi sounds also nice but then doesn't it somehow conflict with the default of 75 DPIs.

uli42:
Using `-autodpi` will get the dpi from your native X server. Using `-dpi X` will set the dpi to X. Using no `-dpi` option at all will give you 75. That's the plan.

If you are using x2goclient you get autodpi behaviour already by _not_ activating the "set display dpi" option.

What do you mean by "I liked the "autodpi" feature and therefore got it back when NoMachine changed the code."?

realsimix:
A couple of years back, the "autodpi" feature (getting DPI info from local display) was the default with nxagent, if you didn't specify `-dpi <$X>`. It was most likely before `NXAGENT_DEFAULT_DPI 75` was introduced.

Just checked, that was back in June 2006 after updating to the NX-2.0.0 code :)

uli42:
Hmm, the changelog for nxagent-2.0.0-79 contains "fix -dpi option". So maybe autodpi wasn't there by intention... ;-)

sunweaver:
@realsimix: Do you have time to work on the -autodpi switch yourself? Or does it need some other dev to take over? Thanks for feedback!

realsimix:
I'd welcome if someone else could look at this. Mainly because I'm not a developer and what should take 10 minutes will take 10 hours for me ;-)

sunweaver:
Hi Simon,

On  Di 21 Nov 2017 05:26:34 CET, Simon Matter wrote:

> I'd welcome if someone else could look at this. Mainly because I'm  
> not a developer and what should take 10 minutes will take 10 hours  
> for me ;-)

Ok. I will do that. Please grant me permission to push commits to your  
PR branch:
https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

I will then add the required commit(s) to the branch and finalize the PR.

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

realsimix:
Hi Mike, because I've sent you the patch by email you've opened the PR so you're already the owner, right?

sunweaver:
On  Di 21 Nov 2017 13:19:39 CET, Simon Matter wrote:

> Hi Mike, because I've sent you the patch by email you've opened the  
> PR so you're already the owner, right?

Right! Good point. Thanks!
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

sunweaver:
@Ionic: @uli42: @realsimix: I have added an -autodpi command line option now. As it does not make much sense switching DPI at session resumptions (only new X clients would use the new setting), I refrained from having autodpi as an nx/nx option.

Also, on MATE 1.20, the new HiDPI code ignores the Xserver's DPI. I will ping MATE upstream on that.

Guys, please review! Thanks!

realsimix:
I've tested both new options `-nomagicpixel` and `-autodpi` and they work as expected in my environment. Thanks for creating the patches!

Regards,
Simon

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -603,6 +606,11 @@ proxy <-> agent remote sessions)
 .B magicpixel=<bool>
 enable/disable magic pixel support in fullscreen mode (default: 1, enabled)
 .TP 8
+.B autodpi=<bool>
+enable/disable deriving session DPI automatically from real server
+(default: 0, disabled); only takes effect on session startups, gets
+ignore when reconnecting to a suspended session
```
Ionic: `ignored`

nx-X11/programs/Xserver/hw/nxagent/Screen.c
```diff
@@ -1327,9 +1330,12 @@ Bool nxagentOpenScreen(ScreenPtr pScreen,
                   rootDepth, (long unsigned int)defaultVisual);
     #endif
 
-    if (monitorResolution < 1)
+    if ((monitorResolution < 1) && (nxagentAutoDPI == False))
     {
       monitorResolution = NXAGENT_DEFAULT_DPI;
+    } else if ((monitorResolution < 1) && (nxagentAutoDPI == True))
```
Ionic:
```
}
else ...
{
```

nx-X11/programs/Xserver/hw/nxagent/Extensions.c
```diff
@@ -338,18 +338,29 @@ static int nxagentRandRInitSizes(ScreenPtr pScreen)
...
+
+      mmHeight = height * DisplayHeightMM(nxagentDisplay, DefaultScreen(nxagentDisplay)) /
+                     DisplayHeight(nxagentDisplay, DefaultScreen(nxagentDisplay));
     }
+    else
+    {
+      mmWidth  = (width * 254 + monitorResolution * 5) / (monitorResolution * 10);
 
-    mmHeight = (height * 254 + monitorResolution * 5) / (monitorResolution * 10);
+      if (mmWidth < 1)
```
Ionic: Move that block and the `if (mmHeight < 1)` block out of the else branch so that it's executed in all cases.

nx-X11/programs/Xserver/hw/nxagent/Screen.c
```diff
@@ -2294,21 +2311,37 @@ Bool nxagentResizeScreen(ScreenPtr pScreen, int width, int height,
...
+    {
+      mmWidth  = width * DisplayWidthMM(nxagentDisplay, DefaultScreen(nxagentDisplay)) /
+                     DisplayWidth(nxagentDisplay, DefaultScreen(nxagentDisplay));
+    }
+    else
     {
-      mmWidth = 1;
+      mmWidth  = (width * 254 + monitorResolution * 5) / (monitorResolution * 10);
+
+      if (mmWidth < 1)
```
Ionic: Move block out.

nx-X11/programs/Xserver/hw/nxagent/Screen.c
```diff
@@ -2294,21 +2311,37 @@ Bool nxagentResizeScreen(ScreenPtr pScreen, int width, int height,
...
+    {
+      mmHeight = height * DisplayHeightMM(nxagentDisplay, DefaultScreen(nxagentDisplay)) /
+                     DisplayHeight(nxagentDisplay, DefaultScreen(nxagentDisplay));
+    }
+    else
     {
-      mmHeight = 1;
+      mmHeight = (height * 254 + monitorResolution * 5) / (monitorResolution * 10);
+
+      if (mmHeight < 1)
```
Ionic: Move block out.

Ionic (changes requested):
Just a few tiny changes.

nx-X11/programs/Xserver/hw/nxagent/Screen.c
sunweaver: Nope. If monitorResolution (which gets set in os/utils.c through the -dpi option) has been set to something >=1 we don't want to modify it at all. So 'else if' correct here.

Ionic: I was merely talking about the formatting, not the content.

nx-X11/programs/Xserver/hw/nxagent/Extensions.c
sunweaver: > Move that block and the if (mmHeight < 1) block out of the else branch so that it's executed in all cases.

Good point. Done.

nx-X11/programs/Xserver/hw/nxagent/Screen.c
sunweaver: > Move block out.

Done

nx-X11/programs/Xserver/hw/nxagent/Screen.c
sunweaver: > Move block out.

Done.

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
sunweaver: > ignored

Got it.

sunweaver:
@Ionic: ready for merging...

Ionic (approved):
Okay, + tiny fixup I'll do before merging.

PR #527, opened 2017-10-11
upgrade to Mesa 6.5

uli42:
Please check, I am unsure how to handle the directory stuff (add a Mesa_6.5 dir and commit all files of it?)

Works for mesa tag mesa_6_5


uli42:
please see my mesa70 branch (https://github.com/uli42/nx-libs/tree/mesa70) for a newer version

sunweaver (commented):
This PR is incomplete regarding the full switch to building against Mesa 6.5. However, the integration part looks good (without having tested it).

Shall we pull in this change or directly jump over to Mesa 7.0 or even a more later version of Mesa?

sunweaver:
will be superceded by #534

PR #537, opened 2017-10-19
nxcomp, nxproxy: don't use symlinks

salva:
Compilation under Windows/cygwin was broken due to the incomplete
symlink support on that platform.

nxcomp/VERSION and nxproxy/VERSION were both symbolic links pointing
to VERSION on their parent directory, but cygwin git was just writting
two regular files instead of symlinks but with the symlink path as their
contents (instead of the target file contents!)

This patch just avoids the symlinks making the configure script look
for the version information on real VERSION file on the parent
directory.


sunweaver:
closed because of..

```
11:05 < salva0> Ionic: you are right, I was using msys cygwin for cloning the repository!
11:06 < salva0> s/msys cygwin/msys git/
11:06 <@sunweaver> salva0: so please check if we can avoid your PR...
11:06 < Ionic> see :)
11:06 <@sunweaver> you could also use cygwin git, call roll-tarball.sh and build from that tarball.
11:06 <@sunweaver> then we could at least leave symlinks in the tree and you have a source tree without symlinks.
11:07 < salva0> sunweaver: yes, it is not needed at all!
```

salva:
The issue was caused by cloning the git repository using MSYS `git` which doesn't support symlinks by default. Using cygwin `git` everything works as expected.

PR #543, opened 2017-11-03
nxcomp/src/Auth.cpp: handle launchd sockets in DISPLAY variable correctly when fetching X cookie.

Ionic:
Previously, fetching the local X cookie always failed and a fake cookie was generated and used instead.

### DIFF
--- a/nxcomp/src/Auth.cpp
+++ b/nxcomp/src/Auth.cpp
@@ -279,10 +279,32 @@ int Auth::getCookie()
   //
   
   char line[DEFAULT_STRING_LIMIT];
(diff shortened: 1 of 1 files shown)


sunweaver (approved):
LGTM. I suppose it has been well tested on Mac OS X.

Ionic:
Tested and works. The general situation is still a little bit awkward since we expect `nxauth` to be available on OS X (and Windows), but I'll leave that for another day.

In general, we should be able to use the system `xauth` binary just fine on OS X (and probably should be doing that in any case.)

PR #544, opened 2017-11-03
nxcomp/src/Loop.cpp: hardcode X11 and (legacy/xfs) font UNIX paths.

### LINKED ISSUE #536: nxproxy should _not_ honor $TEMP for the X11 socket
If `TEMP` is set to anything other then `/tmp` nxproxy will try to find the X servers's socket there. IMHO this is wrong:
```
$ TEMP=dummy strace nxproxy -C :40
...
bind(6, {sa_family=AF_LOCAL, sun_path="dummy/.X11-unix/X40"}, 110) = -1 ENOENT (No such file or directory)
...
```
Clients using the system's libX11/libxcb are depending on the socket being at a predictable location so it must not depend on any local variable.

It is probably this code in `nxcomp/src/Loop.cpp` that causing this. Right before these lines the path is hardcoded for linux, but overwritten afterwards:
```
   4227     struct stat statInfo;
   4228 
   4229     char unixSocketDir[DEFAULT_STRING_LENGTH];
   4230 
   4231     snprintf(unixSocketDir, DEFAULT_STRING_LENGTH - 1, "%s/.X11-unix",
   4232                  control -> TempPath);
   4233 
```

Maybe caused by this commit: aaf4a59e4c4ec54a26283bce6bf25281e12f0424

Ionic:
Fixes: ArcticaProject/nx-libs#536

Allowing overrides via NX_TEMP and TEMP breaks system using
pam_tmpdir.so and most server infrastructure expects the location to be
fixed anyway.

### DIFF
--- a/nxcomp/src/Loop.cpp
+++ b/nxcomp/src/Loop.cpp
@@ -3750,21 +3750,14 @@ int SetupUnixSocket()
   // Open UNIX domain socket for display.
   //
 
-  if (!control->TempPath) {
-    nxfatal << "Loop: PANIC! Temporal path is null.\n" << std::flush;
-
-    cerr << "Error" << ": Temporal path is null.\n";
-    HandleCleanup();
-  }
-
-  unsigned int required = snprintf(unixSocketName, DEFAULT_STRING_LENGTH, "%s/.X11-unix", control->TempPath);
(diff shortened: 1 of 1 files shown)


sunweaver (approved):
LGTM

Ionic:
Rebased.

Ionic:
Also changed a source code comment that still referenced `$TEMP`.

PR #551, opened 2017-11-16
Allow options to be enclosed in '' "" or ||

uli42:
Note: Nesting and escaping is not supported currently!

This is not perfect, e.g. the following string will be processed
wrong:

`key=<v,a,l>,u,e1>,key2=value2`

However, it will help if some value needs to contain a comma (which will be required by one of my keyboard patches to pass full RMLVO keyboard definition via the `keyboard` option.

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -1565,7 +1565,7 @@ static char *processoption(char *string, char *fullstring)
 
   if ((delimiter = strchr(string, '=')))
   {
-    *delimiter = 0;
+    *delimiter = '\0';
     value = delimiter + 1;
     for (int i = 0; i < sizeof(quotechars) - 1; i++)
     {
@@ -1621,7 +1621,7 @@ static void nxagentParseOptionString(char *string)
 
   if ((delimiter = strrchr(string, ':')))
   {
-    *delimiter = 0;
+    *delimiter = '\0';
   }
   else
   {
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -1538,42 +1538,144 @@ static void nxagentParseOptions(char *name, char *value)
   free(argv[0]);
 }
 
+/*
+ * Little helper function to handle quoted options. Quoting must
+ * enclose the whole value part of an option.
+ *
+ * string      the option string up to the first comma (will be modified!)
+ * fullstring  the full option string to the end (will not be modified)
+ *
+ * If the function finds it needs more characters it will take those
+ * from fullstring. It will return the position in fullstring after
+ * the comma following the current processed option
+ *
+ * Returns a pointer to the character in fullstring where the
+ * following option starts.
+ *
+ * If anything goes wrong it will print an error to stderr and return NULL.
+ */
+static char *processoption(char *string, char *fullstring)
+{
+  const char quotechars[] = "'\"|";
+
+  char *delimiter = NULL;
+  char *value = NULL;
+  char *realvalue = NULL;
+  char *next = fullstring + strlen(string);
+
+  if ((delimiter = strchr(string, '=')))
+  {
+    *delimiter = 0;
+    value = delimiter + 1;
+    for (int i = 0; i < sizeof(quotechars) - 1; i++)
+    {
+      if (*value == quotechars[i])
+      {
+        char *valstart = fullstring + (value - string) + 1;
+        if ((next = strchr(valstart, quotechars[i])))
+        {
+          next++; /* skip endchar */
+
+          /* only accept endchar if it is really at the end of the option */
+          if ((*next != ',') && (*next != '\0'))
+          {
+            fprintf(stderr, "unexpected character(s) after end of value for option '%s'\n",
+                    string);
+            next = NULL;
+          }
+          else
+          {
+            realvalue = strndup(valstart, next - 1 - valstart);
+          }
+        }
+        else
+        {
+          fprintf(stderr, "matching '%c' not found for option '%s'\n",
+                  quotechars[i], string);
+        }
+        break;
+      }
+    }
+  }
+
+  if (next)
+  {
+#ifdef DEBUG
+    fprintf(stderr, "%s: found option [%s] value [%s]\n",
+            __func__, string, realvalue ? realvalue : value);
+#endif
+    nxagentParseOptions(string, realvalue ? realvalue : value);
+  }
+  free(realvalue);
+  return next;
+}
+
+
 static void nxagentParseOptionString(char *string)
 {
-  char *value     = NULL;
-  char *option    = NULL;
   char *delimiter = NULL;
 
   /*
    * Remove the port specification.
    */
 
-  delimiter = rindex(string, ':');
-
-  if (delimiter)
+  if ((delimiter = strrchr(string, ':')))
   {
     *delimiter = 0;
   }
   else
   {
-    fprintf(stderr, "Warning: Option file doesn't contain a port specification.\n");
+    fprintf(stderr,
+            "Warning: Option file doesn't contain a port specification.\n");
   }
 
-  while ((option = strtok(option ? NULL : string, ",")))
-  {
-    delimiter = rindex(option, '=');
+  int len = strlen(string);
+  char *end = string + len;
+  char *current = string;
+  char *nextcomma = NULL;
 
-    if (delimiter)
+#ifdef TEST
+  printf("%s: option string: [%s]\n", __func__, string);
+#endif
+
+  while (current && current < end)
+  {
+    if ((nextcomma = strchr(current, ',')))
     {
-      *delimiter = 0;
-      value = delimiter + 1;
+      char *tmp = strndup(current, nextcomma - current);
+      current = processoption(tmp, current);
+      free(tmp);
+      if (current)
+      {
+        if (*current == '\0')
+        {
+          break;
+        }
+        else if (*current == ',')
+        {
+          /* skip the comma */
+          current++;
+        }
+        else
+        {
+          fprintf(stderr, "unexpected character '%c' at position %ld of options string\n",*current, (current - string));
+          current = NULL;
+        }
+      }
+      if (current == NULL)
+      {
+        FatalError("Error parsing options\n");
+      }
     }
     else
     {
-      value = NULL;
+      /* process the last option */
+      if (!(processoption(current, current)))
+      {
+        FatalError("Error parsing options\n");
+      }
+      break;
     }
-
-    nxagentParseOptions(option, value);
   }
 }
 
--- a/nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
+++ b/nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
@@ -447,9 +447,22 @@ enable proxy binding mode
 .B \-options \fIfilename\fP
 path to an options file containing nx/nx options (see below).
 .PP
-Other than the command line options, \fBnxagent\fR can be configured at
-session startup and at runtime (i.e. when resuming a suspended session)
-by so-called nx/nx options.
+Other than the command line options, \fBnxagent\fR can be configured
+at session startup and at runtime (i.e. when resuming a suspended
+session) by so-called nx/nx options. These options are always passed
+as a (sometimes looong) options string consisting of option=value
+pairs, separated by commas. \fBnxagent\fR will extract that options
+string from the DISPLAY variable (see below) or an options file
+(provided via the \-\fIoptions\fR command line option). To overcome
+problems in case a value contains the comma character a simple quoting
+scheme is supported. The quoting character must immediately follow the
+= character and directly precede the next comma. Allowed quoting
+characters are \fI"\fR, \fI'\fR and \fI|\fR. There's no support for
+nesting or escaping. Example:
+.PP
+.nf
+  someoption="some,value",option2=othervalue,option3=|1,2,3,4|
+.fi
 .PP
 As nx/nx options all options supported by nxcomp (see \fBnxproxy\fR man
 page) and all \fBnxagent\fR nx/nx options (see below) can be used.
@@ -478,7 +491,7 @@ by nxcomp already):
 .TP 8
 .B options=<string>
 read options from file, this text file can contain a single loooong
-line with comma-separated nx/nx options
+line with comma-separated nx/nx options.
 .TP 8
 .B rootless=<bool>
 start \fBnxagent\fR in rootless mode, matches \-R given on the command


sunweaver (approved):
It took me sometime to emulate the string parsing in my mind... Looks good to me.

@Ionic: do you want to take a second look?

Ionic:
Yes, I also think we should remove any strings that are nestable as parseable delimiters to avoid ambiguity. I'll look into this later.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -1545,45 +1545,138 @@ static void nxagentParseOptions(char *name, char *value)
...
+{
+  const char startchars[] = "'\"|{([<";
+  const char endchars[]   = "'\"|})]>";
+
+  char *delimiter = NULL;
+  char *value = NULL;
+  char *realvalue = NULL;
+  char *next = fullstring + strlen(string);
+
+  if ((delimiter = index(string, '=')))
```
Ionic: `strchr` instead of `index`.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -1545,45 +1545,138 @@ static void nxagentParseOptions(char *name, char *value)
...
+  {
+    *delimiter = 0;
+    value = delimiter + 1;
+    for (int i = 0; i < strlen(startchars); i++)
```
Ionic: Could use `sizeof(startchars) - 1` here to avoid the overhead of `strlen()`, but not really important.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -1545,45 +1545,138 @@ static void nxagentParseOptions(char *name, char *value)
...
+                  endchars[i], string);
+        }
+        break;
+      }
+    }
+  }
+  else
+  {
+    /* option without value */
+    value = NULL;
```
Ionic: Useless, value will already be `NULL` here.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -1545,45 +1545,138 @@ static void nxagentParseOptions(char *name, char *value)
...
+    *delimiter = 0;
+    value = delimiter + 1;
+    for (int i = 0; i < strlen(startchars); i++)
+    {
+      if (*value == startchars[i])
+      {
+        char *valstart = fullstring + (value - string) + 1;
+        if ((next = strchr(valstart, endchars[i])))
+        {
+          realvalue = strndup(valstart, next - valstart);
```
Ionic: Let's hope that `strndup(s, 0)` is legit... on the platforms I tested, it does what a human would expect, but no idea if that's universally true.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -1545,45 +1545,138 @@ static void nxagentParseOptions(char *name, char *value)
...
   char *delimiter = NULL;
 
   /*
    * Remove the port specification.
    */
 
-  delimiter = rindex(string, ':');
-
-  if (delimiter)
+  if ((delimiter = rindex(string, ':')))
```
Ionic: Change to `strrchr()` if touching this anyway.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -1545,45 +1545,138 @@ static void nxagentParseOptions(char *name, char *value)
...
+          /* skip the comma */
+          current++;
+        }
+        else
+        {
+          fprintf(stderr, "unexpected character '%c' at position %ld of options string\n",*current, (current - string));
+          current = NULL;
+        }
+      }
+      if (current == NULL)
```
Ionic, line 1665: `else`?

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -1545,45 +1545,138 @@ static void nxagentParseOptions(char *name, char *value)
   free(argv[0]);
 }
 
+/*
```
Ionic, line 1541:
```
/*
 * Please use
 * the standard
 * multi-line
 * comment format
 */
```

nx-X11/programs/Xserver/hw/nxagent/Args.c
Ionic: An option string such as `foo="b,ar"baz` will be parsed wrongly, with only `b,ar` passed as the value. It's especially problematic that the remainder `baz` will be handled as a key in the next iteration.

I guess that more work is necessary here.

And it can get even more complex, think of a string like `foo="b,ar"b"az"` as well. We actually need to scan for "quotes" multiple times until we either hit another comma or a `NULL` character.

Ionic (changes requested):
Having thought even more about this, I still maintain that we shouldn't allow delimiters for which the start and end characters are different (like {}, [] and <>), but actually decide on one fixed delimiter.

I don't like `|` because it's non-standard and weird in this context.

From the remaining characters `'` and `"`, `"` is probably most natural.

I'd go with supporting `"` only.

uli42:
So let us participate in your thoughts. Why so you want to limit that? Having more possibilities should not hurt anyone. The rule is: Do not use quotes that conflict with the content or you will get wrong results.

Vom Smartphone gesendet.

Von: "Mihai Moldovan" <notifications@github.com>
Gesendet: ‎18.‎11.‎2017 06:28
Betreff: Re: [ArcticaProject/nx-libs] Allow options to be enclosed in '' "" ||{} () [] or <> (#551)

@Ionic requested changes on this pull request.
Having thought even more about this, I still maintain that we shouldn't allow delimiters for which the start and end characters are different (like {}, [] and <>), but actually decide on one fixed delimiter.
I don't like | because it's non-standard and weird in this context.
From the remaining characters ' and ", " is probably most natural.
I'd go with supporting " only.



In nx-X11/programs/Xserver/hw/nxagent/Args.c:
> +  If the function finds it needs more characters it will take
+  those from fullstring. It will return the position in fullstring
+  after the comma following the current processed option
+
+*/
+static char *processoption(char *string, char *fullstring)
+{
+  const char startchars[] = "'\"|{([<";
+  const char endchars[]   = "'\"|})]>";
+
+  char *delimiter = NULL;
+  char *value = NULL;
+  char *realvalue = NULL;
+  char *next = fullstring + strlen(string);
+
+  if ((delimiter = index(string, '=')))
strchr instead of index.



In nx-X11/programs/Xserver/hw/nxagent/Args.c:
> +*/
+static char *processoption(char *string, char *fullstring)
+{
+  const char startchars[] = "'\"|{([<";
+  const char endchars[]   = "'\"|})]>";
+
+  char *delimiter = NULL;
+  char *value = NULL;
+  char *realvalue = NULL;
+  char *next = fullstring + strlen(string);
+
+  if ((delimiter = index(string, '=')))
+  {
+    *delimiter = 0;
+    value = delimiter + 1;
+    for (int i = 0; i < strlen(startchars); i++)
Could use sizeof(startchars) - 1 here to avoid the overhead of strlen(), but not really important.



In nx-X11/programs/Xserver/hw/nxagent/Args.c:
> +          realvalue = strndup(valstart, next - valstart);
+          next++; /* skip endchar */
+        }
+        else
+        {
+          fprintf(stderr, "matching [%c] not found in value for option [%s]\n",
+                  endchars[i], string);
+        }
+        break;
+      }
+    }
+  }
+  else
+  {
+    /* option without value */
+    value = NULL;
Useless, value will already be NULL here.



In nx-X11/programs/Xserver/hw/nxagent/Args.c:
> +  char *value = NULL;
+  char *realvalue = NULL;
+  char *next = fullstring + strlen(string);
+
+  if ((delimiter = index(string, '=')))
+  {
+    *delimiter = 0;
+    value = delimiter + 1;
+    for (int i = 0; i < strlen(startchars); i++)
+    {
+      if (*value == startchars[i])
+      {
+        char *valstart = fullstring + (value - string) + 1;
+        if ((next = strchr(valstart, endchars[i])))
+        {
+          realvalue = strndup(valstart, next - valstart);
Let's hope that strndup(s, 0) is legit... on the platforms I tested, it does what a human would expect, but no idea if that's universally true.



In nx-X11/programs/Xserver/hw/nxagent/Args.c:
>    char *delimiter = NULL;
 
   /*
    * Remove the port specification.
    */
 
-  delimiter = rindex(string, ':');
-
-  if (delimiter)
+  if ((delimiter = rindex(string, ':')))
Change to strrchr() if touching this anyway.



In nx-X11/programs/Xserver/hw/nxagent/Args.c:
> +        if (*current == '\0')
+        {
+          break;
+        }
+        else if (*current == ',')
+        {
+          /* skip the comma */
+          current++;
+        }
+        else
+        {
+          fprintf(stderr, "unexpected character '%c' at position %ld of options string\n",*current, (current - string));
+          current = NULL;
+        }
+      }
+      if (current == NULL)
else?



In nx-X11/programs/Xserver/hw/nxagent/Args.c:
> @@ -1545,45 +1545,138 @@ static void nxagentParseOptions(char *name, char *value)
   free(argv[0]);
 }
 
+/*
/*
 * Please use
 * the standard
 * multi-line
 * comment format
 */



In nx-X11/programs/Xserver/hw/nxagent/Args.c:
> +  char *value = NULL;
+  char *realvalue = NULL;
+  char *next = fullstring + strlen(string);
+
+  if ((delimiter = index(string, '=')))
+  {
+    *delimiter = 0;
+    value = delimiter + 1;
+    for (int i = 0; i < strlen(startchars); i++)
+    {
+      if (*value == startchars[i])
+      {
+        char *valstart = fullstring + (value - string) + 1;
+        if ((next = strchr(valstart, endchars[i])))
+        {
+          realvalue = strndup(valstart, next - valstart);
An option string such as foo="b,ar"baz will be parsed wrongly, with only b,ar passed as the value. It's especially problematic that the remainder baz will be handled as a key in the next iteration.
I guess that more work is necessary here.
And it can get even more complex, think of a string like foo="b,ar"b"az" as well. We actually need to scan for "quotes" multiple times until we either hit another comma or a NULL character.

Ionic:
I'd exclude delimiters for which start and end characters do not match (`{}`, `()`, `[]` and `<>`), because they are inherently nestable. Explaining to users that this is not supported would be possible, but not all users check the full documentation and it's counter-intuitive. At least in most languages I know such enclosing schemes are used if stuff *needs* to be nestable (think programming languages, JSON, regular expressions etc.) - which doesn't work for delimiters that have the same start and end characters.

This leaves `''`, `""` and `||`.

My point against `||` is that it's non-standard. Sometimes horizontal bars are used to delimit things, but that's mostly for separating columns in data, not for grouping parts.

Leaves `''` and `""`.

Of course we could allow both of them, but these are also slightly ambiguous. In most cases where quoting is allowed via quote characters, double and single quotes behave differently. In such cases, having two options available makes sense, since, for instance, one option (mostly single quotes) treats the quotes text literally, while the other option (mostly double quotes) allows substitutions within the literal text if specific markers are used.

In our case, we don't really have such a connotation in mind - or a need for it, or functionality available.

Using `""` as the only quoting character combination is common, makes our and user's lives easier and is absolutely sufficient.

uli42:
The whole purpose of all this is allowing value containing commas. Depending on the character set in the value you _must_ choose a quoting character that is not contained. So having more to choose makes it more probable that the user (t.i. the programmer of the client application that sets up the option file) finds a suitable one.

Sticking to " will make it impossible to quote anything that contains " _and_ , in the value.

Further, I don't see anyone using nested quotes, simply because there's no use case IMHO. Can you come up with a parameter example that would require nested quoting? As long as we do not offer escaping here I'd say there is none.

The character set I chose for quoting includes the ones I consider as being the best regarding readability. So I included the | character as you can clearly see where it begins and where it ends.

I came across all this when implementing rmlvo keyboard definition for the options string because in that case commas can occur. So there really is a need for that. Instead of enclosing the whole string in quotes or brackets I also thought about using an escaping character instead which would really only be required for escaping commas. However, I consider this being less readable and more complicated to handle from the user's standpoint (I have an nx client script written in bash; it is easier/clearer to add quoting than to escape stuff in bash strings).

One clean approach would be a new format of the options file: instead of having one long line with comma separated values the option file should consist of several lines, one for each option. Quoting/escaping would not be required as long as we restrict values to not contain any line break characters. But: I don't have an idea how to handle values with commas in the DISPLAY or in the case where you pass a string instead of a filename via -options. Maybe we should make a big step here and restrict those cases to the most common/required ones, the rest _has_ to be passed via options file.

Update: improved explanation

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: ok, can do that. The original used rindex (which is wrong IMHO). I can see posix has deprecated both of them.

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: "If s is longer than n, only n bytes are copied, and a terminating null byte ('\0') is added."

It's not clearly stated what happens for n=0 but it is also not stated that 0 is illegal.

uli42: No. Quoting characters are only allowed at the beginning and at the end of the string. Everything else is considered unsupported and will/can produce unexpected results.

However, what about extending that a bit by only excepting the end character it it is directly followed by ',' or '\0'?

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: ok, will fix that.

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: there are some more uses of (r)index() in Binder.c, should also be fixed, but separately

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: no, current can also become NULL inside the if clause. Using an else would skip the check in that case.

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: ok

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: ok

uli42:
pushed updated version: replaced index() calls, use sizeof(), abort if quoting characters appear in the middle of the value

Ionic:
> Further, I don't see anyone using nested quotes, simply because there's no use case IMHO. Can you come up with a parameter example that would require nested quoting? As long as we do not offer escaping here I'd say there is none.

Right, and that's why I recommend not supporting `{}`, `()`, `[]` or `<>`. They *look* nestable, but we don't want to support it and we don't have any use case either.

> Sticking to " will make it impossible to quote anything that contains " and , in the value.

Well... then let's keep `''`, `""` and `||`. Especially the last one should be uncommon enough to not accidentally be part of a quoted string.

nx-X11/programs/Xserver/hw/nxagent/Args.c
Ionic: Yep, right.

nx-X11/programs/Xserver/hw/nxagent/Args.c
Ionic: Yeah, that was my point, too.

Ionic: That at least prohibits unexpected/wrong further parsing and looks correct. I guess that's good enough under the constraint you mentioned.

nx-X11/programs/Xserver/hw/nxagent/Args.c
Ionic: Yes, there are a lot of other cases that use deprecated functionality, but we can update what we need to touch anyway "for free".

sunweaver:
On  Di 21 Nov 2017 03:25:34 CET, Mihai Moldovan wrote:

>> Further, I don't see anyone using nested quotes, simply because  
>> there's no use case IMHO. Can you come up with a parameter example  
>> that would require nested quoting? As long as we do not offer  
>> escaping here I'd say there is none.
>
> Right, and that's why I recommend not supporting `{}`, `()`, `[]` or  
> `<>`. They *look* nestable, but we don't want to support it and we  
> don't have any use case either.
>
>> Sticking to " will make it impossible to quote anything that  
>> contains " and , in the value.
>
> Well... then let's keep `''`, `""` and `||`. Especially the last one  
> should be uncommon enough to not accidentally be part of a quoted  
> string.

Reminds me a bit of ``sed``... Possibly, more chars make sense: ``@``,  
``/``, ``whatnot-supported-by-sed``...

Greets,
Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

uli42:
new version with only ", ' and | and manpage.

Ionic:
> Reminds me a bit of ``sed``... Possibly, more chars make sense: ``@``, ``/``, ``whatnot-supported-by-sed``...

Yes, `sed` is very similar in this respect. It also expects one specific character as the delimiter, but is agnostic as to *what* character is used (i.e., any character will work).

We cannot directly mimic sed here, since
  - we don't allow or support escapes in the options string
  - it's possible that the value starts and ends with the same character incidentally OR is followed by an option that ends in that same character, which opens a world of pain

Say, the string `foo=fo,of,` would be parsed as the kv pair `{foo,{o,o}}`. Not exactly what the user intended...

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -441,9 +441,22 @@ enable proxy binding mode
 .B \-options \fIfilename\fP
 path to an options file containing nx/nx options (see below).
 .PP
-Other than the command line options, \fBnxagent\fR can be configured at
-session startup and at runtime (i.e. when resuming a suspended session)
-by so-called nx/nx options.
+Other than the command line options, \fBnxagent\fR can be configured
+at session startup and at runtime (i.e. when resuming a suspended
```
Ionic, line 451: `i.e.,`

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -441,9 +441,22 @@ enable proxy binding mode
...
+session) by so-called nx/nx options. These options are always passed
+as a (sometime looong) options string consisting of option=value
```
Ionic: `sometimes`

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -441,9 +441,22 @@ enable proxy binding mode
...
+pairs, separated by commas. \fBnxagent\fR will extract that
+options string from the DISPLAY variable (see below) or an options
+file (provided via the \-\fIoptions\fR command line option). To
```
Ionic: Or an **options string** passed on the command line?

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -441,9 +441,22 @@ enable proxy binding mode
...
+overcome problems in case a value contains the comma character a
+simple quoting is supported. The quoting character must immediately
```
Ionic: `simple quoting scheme`?

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -441,9 +441,22 @@ enable proxy binding mode
...
+follow the = character and again directly precede the next
```
Ionic: Leave out `again`?

Ionic (changes requested):
Just a few doc fixups.

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
uli42: I'd rather leave it that way. We are not uniform in this respect, there are occurances of both forms in the manpage and in the sourcecode.

https://jakubmarian.com/comma-after-i-e-and-e-g/ (I prefer British here)

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
uli42: right

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
uli42: I have left that out on purpose:
1. the text is already confusing, I did not want to add more confusion.
2. the manpage does not include any hint that this is possible at all (which should probably be changed; I have opened an issue for that: #565)
3. an options string passed via cmdline is mainly usable for non-persistent sessions and testing. So it is nothing for the majority of users

Further, I'am wondering if we should drop even "options via DISPLAY" out of the equation, see #564.

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
uli42: I like that!

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
uli42: right. I have just pushed a new version.

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
Ionic: Okay, your choice (and likely makes sense since the statement is enclosed in parentheses.)

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
Ionic: Yes and no... the rules for this stuff are quite complicated and currently not documented - mainly the overrides.

I guess we can leave that out of this PR and only mention the most commonly used scheme, yes.

Ionic (approved):

Ionic:
Do you want to rebase the second commit into the first one prior to merging?

uli42:
Please do not merge now - need to adapt nxcomp's option handling in the same way first!

sunweaver:
On  Di 05 Dez 2017 08:55:39 CET, Ulrich Sibiller wrote:

> Please do not merge now - need to adapt nxcomp's option handling in  
> the same way first!

Noted... ACK.

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

uli42:
check: are the changes to nxcomp included already?

sunweaver:
On  Mi 22 Aug 2018 15:41:11 CEST, Ulrich Sibiller wrote:

> check: are the changes to nxcomp included already?

Don't think so. Actually, I think that's why you wanted me to stand by  
on this.

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

uli42:
Looking at the code in nxcomp/src/Loop.cpp I found that nxcomp supports typical URL encoding techniques in the option values (see function `URLDecodeInPlace`) . So instead of quoting we could replace ambiguous chars by their hex representation e.g. `%20` for space. That functionality seems to be missing in nxagent code, though...

uli42:
I have implemented URL decoding in #725. That solves the underlying problem in a very elegant way.

sunweaver:
Superceded by #725, thus closing...

PR #554, opened 2017-11-18
Pr/options on reconnect

uli42:
noticed that 3f7b3001988bf921e6cd860f03a6256b1451ee3d was incomplete. This is the fix.


sunweaver (approved):
LGTM

PR #555, opened 2017-11-18
Pr/simplify free

uli42:
free() can handle NULL so there's not need to check the input ourselves.

This PR also contains some code simplifications that became obvious after doing the free() stuff.

This requires #554.


sunweaver (approved):
LGTM. Please rebase against 3.6.x branch.

PR #556, opened 2017-11-19
Pr/nokeyconv

uli42:
add new option `keyconv`

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -1483,6 +1483,25 @@ static void nxagentParseOptions(char *name, char *value)
 
     return;
   }
+  else if (!strcmp(name, "keyconv"))
+  {
+    if (!strcmp(value, "off")) {
+      nxagentChangeOption(KeycodeConversion, KeycodeConversionOff);
+    }
+    else if (!strcmp(value, "on")) {
+      nxagentChangeOption(KeycodeConversion, KeycodeConversionOn);
+    }
+    else if (!strcmp(value, "auto")) {
+      nxagentChangeOption(KeycodeConversion, KeycodeConversionAuto);
+    }
+    else
+    {
+      fprintf(stderr, "Warning: Ignoring bad value '%s' for option 'keyconv'.\n",
+              validateString(value));
+    }
+
+    return;
+  }
   else
   {
     #ifdef DEBUG
--- a/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
@@ -475,11 +475,11 @@ static CARD8 nxagentConvertedKeycodes[] =
   /* 255 */  255
 };
 
-static int nxagentKeycodeConversion = 0;
+static int nxagentKeycodeConversion = False;
 
 CARD8 nxagentConvertKeycode(CARD8 k)
 {
- if (nxagentKeycodeConversion != 0)
+ if (nxagentKeycodeConversion)
  {
    #ifdef DEBUG
    if (k != nxagentConvertedKeycodes[k])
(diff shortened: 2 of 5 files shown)


sunweaver:
@uli42: This PR looks messed up (als Mesa patches are applied in the diff). Please fix.

uli42:
Oh, that's a consequence of that &%$&/ mesa-quilt stuff. Will fix.

uli42:
ok, mesa mess is fixed

sunweaver (approved):
LGTM, please rebase against 3.6.x branch (or grant me push access to the PR branch)

uli42:
it has been rebased already

uli42:
rebased

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -1483,6 +1483,25 @@ static void nxagentParseOptions(char *name, char *value)
 
     return;
   }
+  else if (!strcmp(name, "keyconv"))
```
Ionic, line 1486: So we don't want to support this as an command line argument as well?

Don't have a strong opinion on this, just asking.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1894,7 +1907,12 @@ void nxagentKeycodeConversionSetup(void)
                 "Activating KeyCode conversion.\n");
     #endif
 
-    nxagentKeycodeConversion = 1;
+    fprintf(stderr, "Info: Keycode conversion auto-determined as on\n");
+    nxagentKeycodeConversion = True;
```
Ionic, line 1911: Need to check for `nxagentOption(KeycodeConversion) != KeycodeConversionOff` here, otherwise we'll ignore this parameter!

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -475,11 +475,11 @@ static CARD8 nxagentConvertedKeycodes[] =
   /* 255 */  255
 };
 
-static int nxagentKeycodeConversion = 0;
+static int nxagentKeycodeConversion = False;
```
Ionic: Could make that `Bool`, since we only have two logical values.

Ionic (changes requested):

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: It partly depends on the client side if keycode conversion is useful or not. If we make this a command line option we cannot change the setting afterwards. And it is really annoying having a session that is reporting wrong keycodes and you can't fix that easily (I ended up in such a situation recently having to undo the keycode conversion via xmodmap...)

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: no, this has already been checked at the beginning of the function.

uli42:
have pushed an update

nx-X11/programs/Xserver/hw/nxagent/Args.c
Ionic: Good point. Then again, if the same option is also specified in an options file or string, then it is possible to override this at run time, since the options file or string is re-parsed and applied upon reconnects, overriding the value passed at start time.

I can follow our argumentation, though. It's not really necessary.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
Ionic: Indeed, I missed the `return` statements for some reason and assumed that these sections were fall-through...

Ionic (approved):

PR #557, opened 2017-11-19
Pr/improve option handling

uli42:
some improvements for `Args.c`

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -1368,7 +1368,7 @@ static void nxagentParseOptions(char *name, char *value)
 
     if ((errno) && (0 == sleep_parse))
     {
-      fprintf(stderr, "Warning: Unable to convert value [%s] of option [%s]. "
+      fprintf(stderr, "Warning: Unable to convert value '%s' of option '%s'. "
                       "Ignoring option.\n",
                       validateString(value), name);
 
@@ -1379,8 +1379,8 @@ static void nxagentParseOptions(char *name, char *value)
     {
       sleep_parse = UINT_MAX;
 
-      fprintf(stderr, "Warning: value [%s] of option [%s] "
-                      "out of range, clamped to [%lu].\n",
+      fprintf(stderr, "Warning: value '%s' of option '%s' "
+                      "out of range, clamped to '%lu'.\n",
                       validateString(value), name, sleep_parse);
     }
 
@@ -1388,8 +1388,8 @@ static void nxagentParseOptions(char *name, char *value)
     {
       sleep_parse = 0;
 
-      fprintf(stderr, "Warning: value [%s] of option [%s] "
-                      "out of range, clamped to [%lu].\n",
+      fprintf(stderr, "Warning: value '%s' of option '%s' "
+                      "out of range, clamped to '%lu'.\n",
                       validateString(value), name, sleep_parse);
     }
 
@@ -1427,7 +1427,7 @@ static void nxagentParseOptions(char *name, char *value)
 
       if ((errno) && (0 == tolerance_parse))
       {
-        fprintf(stderr, "Warning: Unable to convert value [%s] of option [%s]. "
+        fprintf(stderr, "Warning: Unable to convert value '%s' of option '%s'. "
                         "Ignoring option.\n",
                         validateString(value), name);
 
@@ -1438,8 +1438,8 @@ static void nxagentParseOptions(char *name, char *value)
       {
         tolerance_parse = UINT_MAX;
 
-        fprintf(stderr, "Warning: value [%s] of option [%s] "
-                        "out of range, clamped to [%lu].\n",
+        fprintf(stderr, "Warning: value '%s' of option '%s' "
+                        "out of range, clamped to '%lu'.\n",
                         validateString(value), name, tolerance_parse);
       }
 
@@ -1447,8 +1447,8 @@ static void nxagentParseOptions(char *name, char *value)
       {
         tolerance_parse = 0;
 
-        fprintf(stderr, "Warning: value [%s] of option [%s] "
-                        "out of range, clamped to [%lu].\n",
+        fprintf(stderr, "Warning: value '%s' of option '%s' "
+                        "out of range, clamped to '%lu'.\n",
                         validateString(value), name, tolerance_parse);
       }
 
@@ -2415,7 +2415,7 @@ void nxagentSetDeferLevel(void)
     }
     default:
     {
-      fprintf(stderr, "Warning: Unknown link type [%d] processing the defer option.\n",
+      fprintf(stderr, "Warning: Unknown link type '%d' processing the defer option.\n",
                   nxagentOption(LinkType));
 
       deferLevel = 0;
@@ -2569,7 +2569,7 @@ void nxagentSetBufferSize(void)
 
   if (NXSetDisplayBuffer(nxagentDisplay, nxagentBuffer) < 0)
   {
-    fprintf(stderr, "Warning: Can't set the display buffer size to [%d].\n",
+    fprintf(stderr, "Warning: Can't set the display buffer size to '%d'.\n",
                 nxagentBuffer);
   }
 }
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -1438,7 +1438,7 @@ static void nxagentParseOptionString(char *string)
 
   if (delimiter)
   {
-    *delimiter = 0;
+    *delimiter = '\0';
   }
   else
   {
@@ -1451,7 +1451,7 @@ static void nxagentParseOptionString(char *string)
 
     if (delimiter)
     {
-      *delimiter = 0;
+      *delimiter = '\0';
       value = delimiter + 1;
     }
     else
@@ -1594,7 +1594,7 @@ void nxagentProcessOptionsFile(char * filename)
 
   for (offset = 0; (offset < sizeOfFile) && (data[offset] != '\n'); offset++);
 
-  data[offset] = 0;
+  data[offset] = '\0';
 
   nxagentParseOptionString(data);
 
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -219,35 +219,25 @@ int ddxProcessArgument(int argc, char *argv[], int i)
     {
       if ((!strcmp(argv[j], "-display")) && (j + 1 < argc))
       {
-        envOptions = strdup(argv[j + 1]);
-
-        #ifdef WARNING
-        if (envOptions == NULL)
+        if ((envOptions = strdup(argv[j+1])) == NULL)
         {
-          fprintf(stderr, "ddxProcessArgument: WARNING! failed string allocation.\n");
+          FatalError("malloc for -display failed");
         }
-        #endif
-
         break;
       }
     }
 
     if ((envOptions == NULL) && (envDisplay != NULL))
     {
-      envOptions = strdup(envDisplay);
-
-      #ifdef WARNING
-      if (envOptions == NULL)
+      if ((envOptions = strdup(envDisplay)) == NULL)
       {
-        fprintf(stderr, "ddxProcessArgument: WARNING! failed string allocation.\n");
+        FatalError("malloc for envoptions failed");
       }
-      #endif
     }
 
     if (envOptions != NULL)
     {
       nxagentParseOptionString(envOptions);
-
       free(envOptions);
     }
 
@@ -255,26 +245,11 @@ int ddxProcessArgument(int argc, char *argv[], int i)
     {
       if ((!strcmp(argv[j], "-options") || !strcmp(argv[j], "-option")) && j + 1 < argc)
       {
-        if (nxagentOptionsFilenameOrString)
+        free(nxagentOptionsFilenameOrString);
+        if ((nxagentOptionsFilenameOrString = strdup(argv[j + 1])) == NULL)
         {
-          nxagentOptionsFilenameOrString = (char *) realloc(nxagentOptionsFilenameOrString, strlen(argv[j + 1]) + 1);
+          FatalError("malloc for -options failed");
         }
-        else
-        {
-          nxagentOptionsFilenameOrString = (char *) malloc(strlen(argv[j + 1]) +1);
-        }
-
-        if (nxagentOptionsFilenameOrString != NULL)
-        {
-          nxagentOptionsFilenameOrString = strcpy(nxagentOptionsFilenameOrString, argv[j + 1]);
-        }
-        #ifdef WARNING
-        else
-        {
-          fprintf(stderr, "ddxProcessArgument: WARNING! failed string allocation.\n");
-        }
-        #endif
-
         break;
       }
     }
@@ -359,9 +334,13 @@ int ddxProcessArgument(int argc, char *argv[], int i)
       free(nxagentOptionsFilenameOrString);
       nxagentOptionsFilenameOrString = NULL;
 
-      if (-1 == asprintf(&nxagentOptionsFilenameOrString, "%s", argv[i]))
+      /* FIXME: why limit 10 1024 bytes? */
+      if ((size = strlen(argv[i])) < 1024)
       {
-        FatalError("malloc failed");
+        if ((nxagentOptionsFilenameOrString = strndup(argv[i], size)) == NULL)
+        {
+          FatalError("malloc for -options failed");
+        }
       }
       return 2;
     }
@@ -705,10 +684,13 @@ int ddxProcessArgument(int argc, char *argv[], int i)
       free(nxagentKeyboard);
       nxagentKeyboard = NULL;
 
-      nxagentKeyboard = strdup(argv[i]);
-      if (nxagentKeyboard == NULL)
+      /* FIXME: why limit to 256? */
+      if ((size = strlen(argv[i])) < 256)
       {
-        FatalError("malloc failed");
+        if ((nxagentKeyboard = strndup(argv[i], size)) == NULL)
+        {
+          FatalError("malloc for -keyboard failed");
+        }
       }
 
       return 2;
@@ -1017,12 +999,11 @@ int ddxProcessArgument(int argc, char *argv[], int i)
   {
     if (i + 1 < argc)
     {
-      if (NULL != (nxagentKeystrokeFile = strdup(argv[i + 1])))
+      if ((nxagentKeystrokeFile = strdup(argv[i + 1])) == NULL)
       {
-        return 2;
-      } else {
-	FatalError("malloc failed");
+        FatalError("malloc for -keystrokefile failed");
       }
+      return 2;
     }
     return 0;
   }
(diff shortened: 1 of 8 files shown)


nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -215,70 +215,38 @@ int ddxProcessArgument(int argc, char *argv[], int i)
...
-        if (envOptions != NULL)
-        {
-          envOptions = strcpy(envOptions, argv[j + 1]);
-        }
-        #ifdef WARNING
-        else
+        if ((envOptions = strdup(argv[j+1])) == NULL);
         {
-          fprintf(stderr, "ddxProcessArgument: WARNING! failed string allocation.\n");
+          FatalError("malloc failed");
```
sunweaver: Wouldn't it be good to mention what we tried to malloc here?

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -215,70 +215,38 @@ int ddxProcessArgument(int argc, char *argv[], int i)
...
 
     if ((envOptions == NULL) && (envDisplay != NULL))
     {
-      envOptions = malloc(strlen(envDisplay) + 1);
-
-      if (envOptions != NULL)
+      if ((envOptions = strdup(envDisplay)) == NULL);
       {
-        envOptions = strcpy(envOptions, envDisplay);
+        FatalError("malloc failed");
```
sunweaver: dito

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -215,70 +215,38 @@ int ddxProcessArgument(int argc, char *argv[], int i)
...
-          nxagentOptionsFilename = strcpy(nxagentOptionsFilename, argv[j + 1]);
-        }
-        #ifdef WARNING
-        else
+        free(nxagentOptionsFilename);
+        nxagentOptionsFilename = strdup(argv[j + 1]);
+        if (nxagentOptionsFilename == NULL)
         {
-          fprintf(stderr, "ddxProcessArgument: WARNING! failed string allocation.\n");
+          FatalError("malloc failed");
```
sunweaver: dito

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -1085,12 +1047,11 @@ int ddxProcessArgument(int argc, char *argv[], int i)
   {
     if (i + 1 < argc)
     {
-      if (NULL != (nxagentKeystrokeFile = strdup(argv[i + 1])))
+      if ((nxagentKeystrokeFile = strdup(argv[i + 1])) == NULL)
       {
-        return 2;
-      } else {
-	FatalError("malloc failed");
+        FatalError("malloc failed");
```
sunweaver: dito

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -758,16 +723,13 @@ int ddxProcessArgument(int argc, char *argv[], int i)
         nxagentKeyboard = NULL;
       }
 
+      /* FIXME: why limit to 256? */
```
sunweaver: Probably some comand line length limitation adopted here?

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -1351,7 +1351,7 @@ static void nxagentParseOptions(char *name, char *value)
     errno = 0;
     sleep_parse = strtol(value, NULL, 10);
 
-    if ((errno) && (0 == sleep_parse))
+    if ((errno) && (sleep_parse == 0))
```
sunweaver, line 1283: @Ionic: I noticed these comparisons, too, when you worked on the sleep and tolerance feature. Any reason for having the order <val> == <var>?

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -355,7 +355,7 @@ int ddxProcessArgument(int argc, char *argv[], int i)
       /* FIXME: why limit 10 1024 bytes? */
       if ((size = strlen(argv[i])) < 1024)
       {
-        if ((nxagentOptionsFilename = strndup(argv[i], size)) == NULL;
+        if ((nxagentOptionsFilename = strndup(argv[i], size)) == NULL);
```
sunweaver: the semicolon at the line end feels wrong. Also: please rebase with the previous commit that you try to fix with this commit.

sunweaver (changes requested):
@uil42: I left some comments on PR #557. Please amend. Thanks!

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: Well, maybe. But if you happen to suffer from malloc failing you have other problems... ;-) Will extend these messages anyway.

uli42:
I also noticed I still have some fixup commits in there. New version to come soon ;-)

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: I think this is a coding habit that prevents accidently writing `=` instead of `==`. The compiler will complain if you try to assign a value to a constant.

uli42:
Ready for another round. I have added some more changes, so please start over again.

nx-X11/programs/Xserver/hw/nxagent/Args.c
Ionic: Yes, exactly for this reason. It's a pretty common pattern to avoid mistakes (and I'd like to keep that part.)

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -175,10 +175,14 @@ static void nxagentParseOptionString(char*);
...
 
-char nxagentVerbose = 0;
+Bool nxagentVerbose = False;
 
 char *nxagentKeystrokeFile = NULL;
 
+/*
+ * This is the entry point, called from os/utils.c
+ */
+
```
Ionic: I'd drop the newline here and in the next hunk.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -215,71 +225,46 @@ int ddxProcessArgument(int argc, char *argv[], int i)
...
-          nxagentOptionsFilenameOrString = (char *) realloc(nxagentOptionsFilenameOrString, strlen(argv[j + 1]) + 1);
-        }
-        else
-        {
-          nxagentOptionsFilenameOrString = (char *) malloc(strlen(argv[j + 1]) +1);
-        }
-
-        if (nxagentOptionsFilenameOrString != NULL)
+        free(nxagentOptionsFilenameOrString);
+        if ((nxagentOptionsFilenameOrString = strdup(argv[j + 1])) == NULL)
```
Ionic, line 264: That's more compact, but potentially also slower. We probably don't really care since this code will be only called "once". (Actually multiple times, but only once during the program's lifetime.) Just mentioning it.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -215,71 +225,46 @@ int ddxProcessArgument(int argc, char *argv[], int i)
...
+      /* envOptions will be modified! */
       nxagentParseOptionString(envOptions);
-
       free(envOptions);
     }
 
     for (j = 0; j < argc; j++)
     {
+      /*
+       * Read the _last_ options file only and only at startup.
```
Ionic, line 255: This breaks compatibility. Previously, only the first options file was used, now it's the last one. Is that a good idea? At least it would make nxagent more compatible with other tools, since most programs allow users to override values for previous parameters by specifying them again with only the last one taking effect.

To be precise: it may look like it's compatible to the previous behavior, but it isn't. Previously, the code scanned for `-option(s)` and took the first match, then parsed it as an options file. Afterwards, subsequent uses of `ddxProcessArgument()` overrode the internal `nxagentOptionFile` (or nowadays `nxagentOptionsFilenameOrString`) variable. That's an important difference because it allows for a very special combination:

  - start up nxagent with a special options file (first `-option(s)` argument)
  - override it again via subsequent `-option(s)` arguments, which will only change the internal data, but NOT lead to immediate parsing
  - on reconnects ONLY parse the newly registered options file

With your changes, that would not be possible.


I wonder if that behavior is a bug or a feature (i.e., whether this was intended or just an unintentional feature).

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -306,93 +290,48 @@ int ddxProcessArgument(int argc, char *argv[], int i)
...
 
   if (!strcmp(argv[i], "-display"))
   {
     if (++i < argc)
     {
-      strncpy(nxagentDisplayName, argv[i], 1023);
-
-      nxagentDisplayName[1023] = '\0';
-
+      strncpy(nxagentDisplayName, argv[i], sizeof(nxagentDisplayName) - 1);
```
Ionic: Will cause hard to debug problems if we ever decide to change `nxagentDisplayName` to a pure pointer instead of an array. The alternatives are either hardcoding a value (like previously), or using a length macro. Both aren't very nice, but I guess the latter is the safest and best-looking route.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -306,93 +290,48 @@ int ddxProcessArgument(int argc, char *argv[], int i)
...
 
   if (!strcmp(argv[i], "-id"))
   {
     if (++i < argc)
     {
-      strncpy(nxagentSessionId, argv[i], 255);
-
-      *(nxagentSessionId + 255) = '\0';
-
+      strncpy(nxagentSessionId, argv[i], sizeof(nxagentSessionId) - 1);
```
Ionic: Same here.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -215,71 +225,46 @@ int ddxProcessArgument(int argc, char *argv[], int i)
...
       {
-        envOptions = malloc(strlen(argv[j + 1]) + 1);
-
-        if (envOptions != NULL)
-        {
-          envOptions = strcpy(envOptions, argv[j + 1]);
-        }
-        #ifdef WARNING
-        else
+        if ((envOptions = strdup(argv[j+1])) == NULL)
```
Ionic: So it looks like `-display` parameter overrides the environment variable. We might want to edit the comment above to make that clear.

The comment above mentions the order of precedence, and that's true with the exception of the `-display` parameter, which overrides the environment setting. It still holds for `-option(s)`.

uli42:
rebased

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -662,21 +585,17 @@ int ddxProcessArgument(int argc, char *argv[], int i)
...
 
   if (!strcmp(argv[i], "-name"))
   {
     if (++i < argc)
     {
-      strncpy(nxagentWindowName, argv[i], 255);
-
-      *(nxagentWindowName + 255) = '\0';
-
+      strncpy(nxagentWindowName, argv[i], sizeof(nxagentWindowName) - 1);
```
Ionic: Same here.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -732,16 +651,13 @@ int ddxProcessArgument(int argc, char *argv[], int i)
       free(nxagentKeyboard);
       nxagentKeyboard = NULL;
 
+      /* FIXME: why limit to 256? */
```
Ionic: There's probably no good reason for that. A cursory look through the code doesn't reveal any obvious length limitation for the keyboard type/strings. Maybe drop that completely?

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -873,48 +775,42 @@ int ddxProcessArgument(int argc, char *argv[], int i)
...
     return 1;
   }
 
   if (!strcmp(argv[i], "-shadow"))
   {
     if (++i < argc)
     {
-      strncpy(nxagentShadowDisplayName, argv[i], 1023);
+      strncpy(nxagentShadowDisplayName, argv[i],
+              sizeof(nxagentShadowDisplayName) - 1);
```
Ionic: Same problem as a few hunks above.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -1050,21 +935,19 @@ int ddxProcessArgument(int argc, char *argv[], int i)
...
+    nxagentVerbose = True;
     return 1;
   }
 
   if (!strcmp(argv[i], "-keystrokefile"))
   {
     if (i + 1 < argc)
     {
-      if (NULL != (nxagentKeystrokeFile = strdup(argv[i + 1])))
+      if ((nxagentKeystrokeFile = strdup(argv[i + 1])) == NULL)
```
Ionic, line 939: I'd keep that as `NULL == ...`

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -1435,31 +1311,30 @@ static void nxagentParseOptions(char *name, char *value)
       errno = 0;
       tolerance_parse = strtol(value, NULL, 10);
 
-      if ((errno) && (0 == tolerance_parse))
+      if ((errno) && (tolerance_parse == 0))
```
Ionic, line 1339: Also here.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -1761,9 +1624,8 @@ N/A
 
     if (*nxagentWindowName == '\0')
     {
-      strncpy(nxagentWindowName, "NX", 255);
-
-      *(nxagentWindowName + 255) = '\0';
+      strncpy(nxagentWindowName, "NX", sizeof(nxagentWindowName) - 1);
```
Ionic: Same issue here as some hunks above.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -2144,29 +2002,29 @@ static int nxagentGetDialogName()
 {
   strcpy(nxagentDialogName, "NX");
 
-  *(nxagentDialogName + 255) = '\0';
+  nxagentDialogName[sizeof(nxagentDialogName) - 1] = '\0';
```
Ionic: Same problem as some hunks above.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -2144,29 +2002,29 @@ static int nxagentGetDialogName()
...
-    strcpy(nxagentDialogName, "NX - ");
+    strcpy(nxagentDialogName, prefix);
 
-    if (length > (MD5_LENGTH * 2 + 1) &&
-           *(nxagentSessionId + (length - (MD5_LENGTH * 2 + 1))) == '-')
+    if (length > 0 && nxagentSessionId[length] == '-')
     {
-      strncat(nxagentDialogName, nxagentSessionId, length - (MD5_LENGTH * 2 + 1));
+      strncat(nxagentDialogName, nxagentSessionId,
+              MIN(length, sizeof(nxagentDialogName) - 1 - strlen(prefix)));
```
Ionic: Same problem as above.

Bounding the size is actually a pretty good idea here, though.

Ionic (changes requested):

Ionic:
Took me a while go through all of this, sorry...

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: ok

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: right; I'll extend the comment

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: making it behave like most other tools was the intention of my change.

You are also right about changing the options file name for reconnect (only). But consider this an unintended behaviour. The whole options parsing code is very ugly anyway... Besides: that is not documented anywhere, AFAIKS.

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: Speed was not an issue here. I just wanted to simplify the code to make it more readable.

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: (changing those array to pointers is on my list already)

Well, _if_ we change nxagentDisplayName to be a pointer using a hardcoded value will also be wrong. So the situation does not get worse with my change. It only makes it easier to change that hardcoded value because it is now at only one location.

Besides, all this nullbyte stuffing will be gone once we are using pointers. We can even get rid of it now by using s(n)printf() instead of str(n)cpy().

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: done

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: I fully get you point here, but I want to keep the code uniform within a file. And the rest of the code has the constant at the right side. The Xorg code is also using this technique.

uli42:
pushed updated version

nx-X11/programs/Xserver/hw/nxagent/Args.c
Ionic: >  Well, *if* we change nxagentDisplayName to be a pointer using a hardcoded value will also be wrong. So the situation does not get worse with my change. It only makes it easier to change that hardcoded value because it is now at only one location.

Yes, that's why I recommend using a macro defining each length, e.g., `NXAGENTDISPLAYNAME_LENGTH = ...` in the header file and using that throughout the code instead of a hardcoded value.

Even if we change the variables to pure pointers, the situation wouldn't really change. We probably should bound the data length with pure pointers as well. Otherwise, `str(n)cpy` will crash on invalid input data. With bounding, we will in the worst case get `LEN-1` bytes of garbage data, but no actual crashes.

`sizeof(ptr)` doesn't work for pure pointers, so we'd have to change these sections again anyway.

I agree that dropping the hardcoded numbers is a good idea, but IMHO the best way (also taking into account further changes to pure pointers) is to use macros instead.

nx-X11/programs/Xserver/hw/nxagent/Args.c
Ionic: Welll... okay.

Personally, I'll be keeping this style for new submissions, particularly since it's just a cosmetic issue and provides a potential benefit.

I totally respect your decision, though. It's fine.

Ionic (changes requested):

sunweaver:
@Ionic: @uli42: ping on the status of this PR.

sunweaver:
@Ionic: @uli42: ping (2) on the status of this PR.

uli42:
rebased

sunweaver:
@uli42: Is this PR outdated now with #725 merged?

uli42:
No, this is a different topic

Mike Gabriel <notifications@github.com> schrieb am Fr., 24. Aug. 2018,
13:41:

PR #567, opened 2017-11-24
Pr/fix strings

uli42:
Fix two possible buffer overflows, one memory leak and some minor string stuff.

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -215,14 +215,10 @@ int ddxProcessArgument(int argc, char *argv[], int i)
     {
       if ((!strcmp(argv[j], "-display")) && (j + 1 < argc))
       {
-        envOptions = malloc(strlen(argv[j + 1]) + 1);
+        envOptions = strdup(argv[j + 1]);
 
-        if (envOptions != NULL)
-        {
-          envOptions = strcpy(envOptions, argv[j + 1]);
-        }
         #ifdef WARNING
-        else
+        if (envOptions == NULL)
         {
           fprintf(stderr, "ddxProcessArgument: WARNING! failed string allocation.\n");
         }
@@ -234,14 +230,10 @@ int ddxProcessArgument(int argc, char *argv[], int i)
 
     if ((envOptions == NULL) && (envDisplay != NULL))
     {
-      envOptions = malloc(strlen(envDisplay) + 1);
+      envOptions = strdup(envDisplay);
 
-      if (envOptions != NULL)
-      {
-        envOptions = strcpy(envOptions, envDisplay);
-      }
       #ifdef WARNING
-      else
+      if (envOptions == NULL)
       {
         fprintf(stderr, "ddxProcessArgument: WARNING! failed string allocation.\n");
       }
@@ -696,34 +688,14 @@ int ddxProcessArgument(int argc, char *argv[], int i)
   {
     if (++i < argc)
     {
-      int size;
-
       free(nxagentKeyboard);
       nxagentKeyboard = NULL;
 
-      if ((size = strlen(argv[i])) < 256)
+      nxagentKeyboard = strdup(argv[i]);
+      if (nxagentKeyboard == NULL)
       {
-        if ((nxagentKeyboard = malloc(size + 1)) == NULL)
-        {
-          FatalError("malloc failed");
-        }
-
-        strncpy(nxagentKeyboard, argv[i], size);
-
-        nxagentKeyboard[size] = '\0';
+        FatalError("malloc failed");
       }
-      #ifdef WARNING
-      else
-      {
-        /*
-         * it is useless to remember a kbtype
-         * option that has just been truncated.
-         */
-
-        fprintf(stderr, "ddxProcessArgument: WARNING! Option [%s] too long. "
-                    "It will be ignored.\n", argv[i]);
-      }
-      #endif
 
       return 2;
     }
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -2098,30 +2098,23 @@ void ddxUseMsg()
 
 static int nxagentGetDialogName()
 {
-  snprintf(nxagentDialogName, NXAGENTDIALOGNAMELENGTH, "NX");
-
   if (*nxagentSessionId != '\0')
   {
     int length = strlen(nxagentSessionId);
-
-    strcpy(nxagentDialogName, "NX - ");
 
     /* if the session id contains an MD5 hash in a well-known format cut it off */
     if (length > (MD5_LENGTH * 2 + 1) &&
            *(nxagentSessionId + (length - (MD5_LENGTH * 2 + 1))) == '-')
     {
-      strncat(nxagentDialogName, nxagentSessionId,
-              MIN(NXAGENTDIALOGNAMELENGTH - strlen(nxagentDialogName), length - (MD5_LENGTH * 2 + 1)) - 1);
-    }
-    else
-    {
-      strncat(nxagentDialogName, nxagentSessionId, NXAGENTDIALOGNAMELENGTH - strlen(nxagentDialogName) - 1);
+      length -= (MD5_LENGTH * 2 + 1);
     }
 
-    nxagentDialogName[NXAGENTDIALOGNAMELENGTH - 1] = '\0';
+    snprintf(nxagentDialogName, NXAGENTDIALOGNAMELENGTH, "NX - %.*s", length, nxagentSessionId);
 
     return 1;
   }
+
+  snprintf(nxagentDialogName, NXAGENTDIALOGNAMELENGTH, "NX");
 
   return 0;
 }
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -1254,7 +1254,7 @@ static void nxagentParseOptions(char *name, char *value)
   }
   else if (strcmp(name, "clients") == 0)
   {
-    strcpy(nxagentClientsLogName, value);
+    snprintf(nxagentClientsLogName, NXAGENTCLIENTSLOGNAMELENGTH, "%s", value);
 
     return;
   }
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -126,15 +126,15 @@ extern char dispatchExceptionAtReset;
 
 extern const char *__progname;
 
-char nxagentDisplayName[1024];
+char nxagentDisplayName[NXAGENTDISPLAYNAMELENGTH];
 Bool nxagentSynchronize = False;
 Bool nxagentRealWindowProp = False;
 
-char nxagentShadowDisplayName[1024] = {0};
+char nxagentShadowDisplayName[NXAGENTSHADOWDISPLAYNAMELENGTH] = {0};
 
-char nxagentWindowName[256];
-char nxagentDialogName[256];
-char nxagentSessionId[256] = {0};
+char nxagentWindowName[NXAGENTWINDOWNAMELENGTH];
+char nxagentDialogName[NXAGENTDIALOGNAMELENGTH];
+char nxagentSessionId[NXAGENTSESSIONIDLENGTH] = {0};
 char *nxagentOptionsFilenameOrString;
 
 Bool nxagentFullGeneration = False;
@@ -326,10 +326,7 @@ int ddxProcessArgument(int argc, char *argv[], int i)
   {
     if (++i < argc)
     {
-      strncpy(nxagentDisplayName, argv[i], 1023);
-
-      nxagentDisplayName[1023] = '\0';
-
+      snprintf(nxagentDisplayName, NXAGENTDISPLAYNAMELENGTH, "%s", argv[i]);
       return 2;
     }
 
@@ -340,10 +337,7 @@ int ddxProcessArgument(int argc, char *argv[], int i)
   {
     if (++i < argc)
     {
-      strncpy(nxagentSessionId, argv[i], 255);
-
-      *(nxagentSessionId + 255) = '\0';
-
+      snprintf(nxagentSessionId, NXAGENTSESSIONIDLENGTH, "%s", argv[i]);
       return 2;
     }
 
@@ -360,35 +354,13 @@ int ddxProcessArgument(int argc, char *argv[], int i)
   {
     if (++i < argc)
     {
-      int size;
-
       free(nxagentOptionsFilenameOrString);
       nxagentOptionsFilenameOrString = NULL;
 
-      if ((size = strlen(argv[i])) < 1024)
+      if (-1 == asprintf(&nxagentOptionsFilenameOrString, "%s", argv[i]))
       {
-        if ((nxagentOptionsFilenameOrString = malloc(size + 1)) == NULL)
-        {
-          FatalError("malloc failed");
-        }
-
-        strncpy(nxagentOptionsFilenameOrString, argv[i], size);
-
-        nxagentOptionsFilenameOrString[size] = '\0';
+        FatalError("malloc failed");
       }
-      else
-      {
-        /*
-         * It is useless to store the file name
-         * that has just been truncated.
-         */
-
-        #ifdef WARNING
-        fprintf(stderr, "ddxProcessArgument: WARNING! Option file name "
-                    "too long. It will be ignored.\n");
-        #endif
-      }
-
       return 2;
     }
 
@@ -670,10 +642,7 @@ int ddxProcessArgument(int argc, char *argv[], int i)
   {
     if (++i < argc)
     {
-      strncpy(nxagentWindowName, argv[i], 255);
-
-      *(nxagentWindowName + 255) = '\0';
-
+      snprintf(nxagentWindowName, NXAGENTWINDOWNAMELENGTH, "%s", argv[i]);
       return 2;
     }
 
@@ -903,14 +872,12 @@ int ddxProcessArgument(int argc, char *argv[], int i)
   {
     if (++i < argc)
     {
-      strncpy(nxagentShadowDisplayName, argv[i], 1023);
+      snprintf(nxagentShadowDisplayName, NXAGENTSHADOWDISPLAYNAMELENGTH, "%s", argv[i]);
 
       if (strcmp(nxagentShadowDisplayName, "") == 0)
       {
         FatalError("Invalid shadow display option");
       }
-
-      *(nxagentShadowDisplayName + 1023) = '\0';
 
       return 2;
     }
@@ -1780,9 +1747,7 @@ N/A
 
     if (*nxagentWindowName == '\0')
     {
-      strncpy(nxagentWindowName, "NX", 255);
-
-      *(nxagentWindowName + 255) = '\0';
+      snprintf(nxagentWindowName, NXAGENTWINDOWNAMELENGTH, "NX");
     }
 
     /*
@@ -2161,9 +2126,7 @@ void ddxUseMsg()
 
 static int nxagentGetDialogName()
 {
-  strcpy(nxagentDialogName, "NX");
-
-  *(nxagentDialogName + 255) = '\0';
+  snprintf(nxagentDialogName, NXAGENTDIALOGNAMELENGTH, "NX");
 
   if (*nxagentSessionId != '\0')
   {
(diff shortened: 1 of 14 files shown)


nx-X11/programs/Xserver/hw/nxagent/Display.c
```diff
@@ -1907,8 +1905,8 @@ Bool nxagentMakeIcon(Display *display, Pixmap *nxIcon, Pixmap *nxMask)
     agentIconData=nxagentIconData;
   }
 
-
-  snprintf(default_path, PATH_MAX-1, "/usr/NX/share/images/%s", agent_icon_name);
+  /* FIXME: use a compile tiem define here, /usr/NX is a nomachine path */
```
sunweaver: s/tiem/time/

sunweaver (commented):
I went over this quickly. Need to take a second glance once we have merged the other PRs.

uli42:
fixed typo and rebased

nx-X11/programs/Xserver/hw/nxagent/Args.h
```diff
@@ -53,7 +53,7 @@ struct UserGeometry{
 extern Bool nxagentUseNXTrans;
 
 extern char nxagentSessionId[];
-extern char nxagentDisplayName[];
+extern char nxagentDisplayName[1024];
```
Ionic: Don't do this. The size given here is merely a hint and will be ignored.

Even worse, it's allowed for the definition in a source file to deviate from that declaration.

We *really* should use SIZE macros for these arrays.

`Args.h`:
```
char nxagentDisplayName[];
#define NXAGENTDISPLAYNAMELENGTH 1024;
```

`Args.c`:
```
char nxagentDisplayName[NXAGENTDISPLAYNAMELENGTH];
```

An alternative way would be define size variables:

`Args.h`:
```
char nxagentDisplayName[];
size_t nxagentDisplayNameLength;
```

`Args.c`:
```
char nxagentDisplayName[1024];
size_t nxagentDisplayNameLength = (sizeof(nxagentDisplayName) / sizeof(*nxagentDisplayName));
```

The drawback is that this alternative will only work for fixed-size arrays, but not for pure pointers.

nx-X11/programs/Xserver/hw/nxagent/Display.c
```diff
@@ -1844,22 +1842,35 @@ static FILE *nxagentLookForIconFile(char *iconName, const char *permission,
...
     {
+      if (strlen(path) > sizeof(singlePath) - 1)
+      {
+        fprintf(stderr, "Error: Path too long.\n");
+        return NULL;
+      }
+
       strcpy(singlePath, path);
 
       breakLoop = 1;
```
Ionic: With these changes, it doesn't make sense to keep `breakLoop` around. Drop it.

nx-X11/programs/Xserver/hw/nxagent/Display.c
```diff
@@ -1907,8 +1918,8 @@ Bool nxagentMakeIcon(Display *display, Pixmap *nxIcon, Pixmap *nxMask)
     agentIconData=nxagentIconData;
   }
 
-
-  snprintf(default_path, PATH_MAX-1, "/usr/NX/share/images/%s", agent_icon_name);
+  /* FIXME: use a compile time define here, /usr/NX is a nomachine path */
+  snprintf(default_path, sizeof(default_path), "/usr/NX/share/images/%s", agent_icon_name);
```
Ionic, line 1917: I agree, but wonder why the original code used `PATH_MAX - 1`. Probably didn't know that `snprintf` includes the terminating `0` character in the length and only writes `n - 1` elements to the target buffer (+ terminating `0`). `strncpy` is different in this regard. They probably got confused by that...

nx-X11/programs/Xserver/hw/nxagent/Font.c
```diff
@@ -1406,9 +1406,10 @@ Bool nxagentDisconnectAllFonts()
 
 static Bool nxagentGetFontServerPath(char * fontServerPath)
 {
-  char path[256];
+  /* ensure path is not longer that fontServerPath) */
+  char path[sizeof(fontServerPath)];
```
Ionic: Can't do this. `fontServerPath` is a pure pointer. We cannot use `sizeof()` here.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1858,16 +1858,17 @@ void nxagentKeycodeConversionSetup(void)
     char *sessionpath = nxagentGetSessionPath();
     if (sessionpath != NULL)
     {
-      int keyboard_file_path_size = strlen(sessionpath) + strlen("/keyboard");
-      char *keyboard_file_path = malloc((keyboard_file_path_size + 1) * sizeof(char));
+      char *keyboard_file_path = NULL;
       FILE *keyboard_file;
-      if (keyboard_file_path == NULL)
+
+      if ((asprintf(&keyboard_file_path, "%s/keyboard", sessionpath) == -1))
```
Ionic: Nope, we can't use `asprintf` yet. It's not your fault, but really mine since I didn't notice this error when reviewing some of @sunweaver's commits that used `asprintf`.

It doesn't mean that you have to change the code - but this PR will be blocked until I've backported `os/xprintf.c` so that `asprintf` is available across platforms.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1858,16 +1858,17 @@ void nxagentKeycodeConversionSetup(void)
...
       {
-        FatalError("nxagentKeycodeConversionSetup: malloc failed.");
+        free(sessionpath);
```
Ionic: Okay, but `FatalErrror` is `noreturn` anyway. At this point a memory leak won't hurt. Still good to free the memory.

nx-X11/programs/Xserver/hw/nxagent/Dialog.c
```diff
@@ -308,7 +305,7 @@ void nxagentLaunchDialog(DialogType dialogType)
               DECODE_DIALOG_TYPE(dialogType), *pid, dialogDisplay);
   #endif
 
-  *dialogDisplay = '\0';
+  dialogDisplay[0] = '\0';
```
Ionic, line 308: Functionally equivalent. Just a comment.

nx-X11/programs/Xserver/hw/nxagent/NXdixfonts.c
```diff
@@ -112,7 +112,7 @@ static const char *_NXGetFontPath(const char *path)
      * Check the environment only once.
      */
 
-    if (*_NXFontPath != '\0')
+    if (_NXFontPath[0] != '\0')
```
Ionic, line 116: Functionally equivalent. Just a comment.

Ionic (changes requested):

nx-X11/programs/Xserver/hw/nxagent/Args.h
uli42: Hmm, I had to add the size here because otherwise the sizeof would fail if used in sources _other than_ Args.c. So  I am wondering if you are really right about the "only a hint" stuff.

Generally I want to get rid of all those limits as they are only artificial. So I will change the code one day to not use fixed array.

nx-X11/programs/Xserver/hw/nxagent/Dialog.c
uli42: yeah, but better to read IMHO.

nx-X11/programs/Xserver/hw/nxagent/Display.c
uli42: that's what I thought, too.

nx-X11/programs/Xserver/hw/nxagent/Display.c
uli42: must check

nx-X11/programs/Xserver/hw/nxagent/Font.c
uli42: you are right, will fix that.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
uli42: what platforms are lacking it?

"     These functions are GNU extensions, not in C or POSIX.  They are also
       available under *BSD.  The FreeBSD implementation sets strp to NULL
       on error."

uli42: By the way: backporting asprintf doesn't sound like fun: http://insanecoding.blogspot.de/2014/06/memory-management-in-c-and-auto.html

nx-X11/programs/Xserver/hw/nxagent/Args.h
Ionic: Looks like I've been wrong - `T foo[];` declares an incomplete type that must be properly defined at a later point (unless the code in question does not need storage information).

Multiple declarations for he same object with different sizes lead a compile error in recent versions of both clang and gcc.

There is a special case that I confused this with: within parameters declared like this, the size is merely a hint:

```
b.c: In function ‘get_size_2’:
b.c:9:18: warning: ‘sizeof’ on array function parameter ‘arg’ will return size of ‘char *’ [-Wsizeof-array-argument]
   return (sizeof (arg));
                  ^
b.c:8:25: note: declared here
 size_t get_size_2 (char arg[6]) {
                         ^~~

(the array I pass in actually is `char foo[5]` - and that works)
```

For now, I'd really use a macro.

Bounding even pure pointers would also be a good idea, even if that adds an artificial limit. We can certainly raise this limit, but not doing it at all is dangerous in case the input data is invalid.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
Ionic: I don't know what platforms are not providing it exactly, but there certainly seem to be X.Org-supported ones that don't.

Backporting `asprintf` is really easy! But only because X.Org already did it. I'll take a stab at it tomorrow (or at a later time) after I went through the other open PRs.

nx-X11/programs/Xserver/hw/nxagent/Font.c
```diff
@@ -1406,15 +1406,17 @@ Bool nxagentDisconnectAllFonts()
...
   {
-    if (*path != '\0')
-    {
-      strncpy(fontServerPath, path + 1, *path);
+    unsigned int len = *path;
 
-      *(fontServerPath + *path) = '\0';
+    if (len)
+    {
+      strncpy(fontServerPath, path + 1, len);
```
Ionic: Might as well use `snprintf()` here.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
Ionic: `asprintf` comment obsolete.

nx-X11/programs/Xserver/hw/nxagent/Display.c
```diff
@@ -1844,22 +1842,35 @@ static FILE *nxagentLookForIconFile(char *iconName, const char *permission,
...
     if (end != NULL)
     {
+      if ((end - path) > sizeof(singlePath) - 1)
+      {
+        fprintf(stderr, "Warning: Path too long - ignored.\n");
+        path = end + 1;
+        continue;
+      }
+
       strncpy(singlePath, path, (unsigned long)(end - path));
```
Ionic: `snprintf` to drop the `NULL` setting.

nx-X11/programs/Xserver/hw/nxagent/Display.c
Ionic: Ping.

Ionic (changes requested):

nx-X11/programs/Xserver/hw/nxagent/Display.c
uli42: I don't think so. `breakLoop` indicates that we have processed the last path from the env variable. If we have checked that one (and found no file in that path we still need to terminate the loop).

I have improved the code:
- use True/False
- move variables to inner scope where possible
- add comments
- strip _all_ trailing slashes, not only the last one
- pass down the size of the return_path array

nx-X11/programs/Xserver/hw/nxagent/Args.h
uli42: ok, I have added macros (for now). And fixed two bugs in nxagentGetDialogName: the wrong string was terminated and there where no length checks.

nx-X11/programs/Xserver/hw/nxagent/Display.c
uli42: done

nx-X11/programs/Xserver/hw/nxagent/Font.c
uli42: <s>will leave that for now</s> have changed it to that we now the size here; also added a MIN() guard

uli42:
@Ionic can you please re-check?

nx-X11/programs/Xserver/hw/nxagent/Display.c
```diff
@@ -1836,56 +1834,56 @@ static FILE *nxagentLookForIconFile(char *iconName, const char *permission,
...
     {
-      singlePath[strlen(singlePath)- 1] = '\0';
+      singlePath[strlen(singlePath) - 1] = '\0';
     }
 
+    /* append slash and icon name */
     if (strlen(singlePath) + strlen(iconName) + 1 < PATH_MAX)
     {
       strncat(singlePath, slash, 1);
       strcat(singlePath, iconName);
```
Ionic: Hello, I'm Shlemiel the Painter!

How about...
```C
  snprintf(singlePath + strlen(singlePath), sizeof(singlePath), "%.1s%s", slash, iconName);
```
?

nx-X11/programs/Xserver/hw/nxagent/Display.c
```diff
@@ -1836,56 +1834,56 @@ static FILE *nxagentLookForIconFile(char *iconName, const char *permission,
...
-    if (singlePath[strlen(singlePath)- 1] == slash[0])
+    /* cut off trailing slashes, if any */
+    while (singlePath[strlen(singlePath) - 1] == slash[0])
     {
-      singlePath[strlen(singlePath)- 1] = '\0';
+      singlePath[strlen(singlePath) - 1] = '\0';
     }
 
+    /* append slash and icon name */
     if (strlen(singlePath) + strlen(iconName) + 1 < PATH_MAX)
```
Ionic: Replace `PATH_MAX` here with `sizeof(singlePath)`?

nx-X11/programs/Xserver/hw/nxagent/Error.c
```diff
@@ -118,6 +118,8 @@ int nxagentErrorHandler(Display *dpy, XErrorEvent *event)
```
Ionic, line 118: "commt" typo in commit message.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -696,34 +688,14 @@ int ddxProcessArgument(int argc, char *argv[], int i)
   {
     if (++i < argc)
     {
-      int size;
-
       free(nxagentKeyboard);
       nxagentKeyboard = NULL;
 
-      if ((size = strlen(argv[i])) < 256)
```
Ionic, line 735: Probably good to just drop this 255 character limitation here, I haven't found a reason why this would be necessary.

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -2098,31 +2098,24 @@ void ddxUseMsg()
...
-              MIN(NXAGENTDIALOGNAMELENGTH - strlen(nxagentDialogName), length - (MD5_LENGTH * 2 + 1)) - 1);
-    }
-    else
-    {
-      strncat(nxagentDialogName, nxagentSessionId, NXAGENTDIALOGNAMELENGTH - strlen(nxagentDialogName) - 1);
+      length -= (MD5_LENGTH * 2 + 1);
     }
 
-    nxagentDialogName[NXAGENTDIALOGNAMELENGTH - 1] = '\0';
+    snprintf(nxagentDialogName, NXAGENTDIALOGNAMELENGTH, "NX - %.*s", length, nxagentSessionId);
```
Ionic: I think that should be `length - 1` (or `1` subtracted above) because it looks like we don't want to copy the leading `-` char either. At least the original code seems to have skipped it.

nx-X11/programs/Xserver/hw/nxagent/Display.c
```diff
@@ -1838,43 +1834,56 @@ static FILE *nxagentLookForIconFile(char *iconName, const char *permission,
     return NULL;
   }
 
-  for(breakLoop = 0; breakLoop == 0 && fptr == NULL; )
+  for (int breakLoop = False; breakLoop == False && fptr == NULL; )
```
Ionic: Can drop `breakLoop` by using:

```C
  for (char *end = path; end == NULL && fptr == NULL; )
```

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
```diff
@@ -1904,12 +1904,14 @@ void nxagentKeycodeConversionSetup(void)
...
+        FatalError("malloc for keyboard file path failed.");
       }
       strcpy(keyboard_file_path, sessionpath);
       strcat(keyboard_file_path, "/keyboard");
+      free(sessionpath);
       if ((keyboard_file = fopen(keyboard_file_path, "w"))) {
-        if (drules)
+        if (drules != NULL)
```
Ionic: Changing this makes it weirdly inconsistent with the later code?

Ionic (changes requested):

nx-X11/programs/Xserver/hw/nxagent/Display.c
uli42: Can do that, but weren't you the one that didn't like that? ;-)

nx-X11/programs/Xserver/hw/nxagent/Display.c
uli42: Ok, accepted, but the .1 is not needed because slash is only one char.

nx-X11/programs/Xserver/hw/nxagent/Error.c
uli42: fixed

nx-X11/programs/Xserver/hw/nxagent/Display.c
uli42: done

nx-X11/programs/Xserver/hw/nxagent/Display.c
uli42: No, I'd have to check for `end != NULL && fprt == NULL`. Done.

uli42:
ok, I fixed the four things you mentioned

nx-X11/programs/Xserver/hw/nxagent/Display.c
Ionic: Yes, sorry.

nx-X11/programs/Xserver/hw/nxagent/Args.c
Ionic: Ping on that?

nx-X11/programs/Xserver/hw/nxagent/Display.c
Ionic: For global, fixed-size arrays yes. For local ones using a special macro doesn't make a lot of sense, though. `PATH_MAX` is a too generic one.

nx-X11/programs/Xserver/hw/nxagent/Display.c
Ionic: Okay, I thought thought it might be a good idea to limit reading data out of `slash` to one character, just in case.

nx-X11/programs/Xserver/hw/nxagent/Keyboard.c
Ionic: Ping on that?

uli42: ok, fixed

uli42: fixed

nx-X11/programs/Xserver/hw/nxagent/Args.c
uli42: hmm, but we are subtracting 2+MD5+1 from the length already!?

Ionic: `2*MD5 + 1` actually... and yes, that's fine. I got confused by the different behaviors regarding the total string length (since `snprintf` expects the full length, but the `str*` family normally the length excluding a terminating `NULL` character). That's not relevant here, so the code is fine. My bad.

Ionic (approved):

Issue #573, opened 2017-11-27
Buffer overflow in nxcomp


uli42:
1. run `nxproxy -C loopback=1:11`
2. run `nxproxy -S connect=localhost:11`

Proxy #1 crashes shortly after the connection is established:
```
NXPROXY - Version 3.5.99.10

Copyright (c) 2001, 2011 NoMachine (http://www.nomachine.com)
Copyright (c) 2008-2014 Oleksandr Shneyder <o.shneyder@phoca-gmbh.de>
Copyright (c) 2014-2016 Ulrich Sibiller <uli42@gmx.de>
Copyright (c) 2014-2016 Mihai Moldovan <ionic@ionic.de>
Copyright (c) 2011-2016 Mike Gabriel <mike.gabriel@das-netzwerkteam.de>
Copyright (c) 2015-2016 Qindel Group (http://www.qindel.com)

NXCOMP, NX protocol compression and NX extensions to this software
are copyright of the aforementioned persons and companies.

Redistribution and use of the present software is allowed according
to terms specified in the file LICENSE.nxcomp which comes in the
source distribution.

All rights reserved.

NOTE: This software has received contributions from various other
contributors, only the core maintainers and supporters are listed as
copyright holders. Please contact us, if you feel you should be listed
as copyright holder, as well.

NX protocol compression is derived from DXPC project.

Copyright (c) 1995,1996 Brian Pane
Copyright (c) 1996,1997 Zachary Vonler and Brian Pane
Copyright (c) 1999 Kevin Vigor and Brian Pane
Copyright (c) 2000,2003 Gian Filippo Pinzari and Brian Pane

All rights reserved.

See https://github.com/ArcticaProject/nx-libs for more information.

Info: Proxy running in client mode with pid '13979'.
Session: Starting session at 'Tue Nov 28 00:18:16 2017'.
Info: Using errors file '/home/uli/.nx/C-11/errors'.
Info: Using stats file '/home/uli/.nx/C-11/stats'.
Info: Waiting for connection from any host on socket 'tcp:localhost:4011'.
Info: Accepted connection from '127.0.0.1'.
Info: Connection with remote proxy completed.
Info: Using lan link parameters 1536/24/1/0.
Info: Using agent parameters 5000/0/50/0/0.
Info: Using pack method 'adaptive-9' with session ''.
Info: Not using NX delta compression.
Info: Not using ZLIB data compression.
Info: Not using ZLIB stream compression.
[...]

PR #575, opened 2017-11-28
Pr/fix memleaks

uli42:
add some missing free calls

Fixes parts of #573, #596 and some more unreported memleaks

### DIFF
--- a/nxcomp/src/ChannelEndPoint.cpp
+++ b/nxcomp/src/ChannelEndPoint.cpp
@@ -58,7 +58,7 @@ ChannelEndPoint::~ChannelEndPoint()
 
 void
 ChannelEndPoint::setSpec(const char *spec) {
-  if (spec_) free(spec_);
+  free(spec_);
 
   if (spec && strlen(spec))
   {
@@ -90,7 +90,7 @@ void
 ChannelEndPoint::setSpec(const char *hostName, long port) {
   int length;
 
-  if (spec_) free(spec_);
+  free(spec_);
   spec_ = NULL;
 
   isUnix_ = false;
@@ -162,7 +162,7 @@ ChannelEndPoint::setDefaultTCPInterface(int publicInterface) {
 
 void
 ChannelEndPoint::setDefaultUnixPath(char *path) {
-  if (defaultUnixPath_) free(defaultUnixPath_);
+  free(defaultUnixPath_);
 
   if (path && strlen(path))
     defaultUnixPath_ = strdup(path);
--- a/nxcomp/src/ChannelEndPoint.cpp
+++ b/nxcomp/src/ChannelEndPoint.cpp
@@ -91,6 +91,8 @@ ChannelEndPoint::setSpec(const char *hostName, long port) {
   int length;
 
   if (spec_) free(spec_);
+  spec_ = NULL;
+
   isUnix_ = false;
   isTCP_ = false;
 
--- a/nxcomp/src/ChannelEndPoint.cpp
+++ b/nxcomp/src/ChannelEndPoint.cpp
@@ -54,6 +54,12 @@ ChannelEndPoint::~ChannelEndPoint()
     if(S_ISSOCK(st.st_mode))
       unlink(unixPath);
   }
+  free(unixPath);
+  unixPath = NULL;
+  free(defaultUnixPath_);
+  defaultUnixPath_ = NULL;
+  free(spec_);
+  spec_ = NULL;
 }
 
 void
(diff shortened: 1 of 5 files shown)


sunweaver (approved):
LGTM. Thanks for these fixes! Please rebase.

uli42:
rebased

nxcomp/src/ChannelEndPoint.cpp
```diff
@@ -54,11 +54,14 @@ ChannelEndPoint::~ChannelEndPoint()
     if(S_ISSOCK(st.st_mode))
       unlink(unixPath);
   }
+  free(unixPath);
```
Ionic: Also set `unixPath` to `NULL` to avoid double-free issues. Yes, this a destructor and this is a local variable, so it shouldn't matter, but it's still good style to always do so.

nxcomp/src/ChannelEndPoint.cpp
```diff
@@ -54,11 +54,14 @@ ChannelEndPoint::~ChannelEndPoint()
...
+  free(defaultUnixPath_);
```
Ionic: Also for these two.

nxcomp/src/ChannelEndPoint.cpp
```diff
@@ -90,13 +93,13 @@ void
 ChannelEndPoint::setSpec(const char *hostName, long port) {
   int length;
 
-  if (spec_) free(spec_);
   isUnix_ = false;
```
Ionic, line 98: I'd keep `free(spec_)` here, it fits well into the scope logic.

nxcomp/src/Loop.cpp
```diff
@@ -4278,15 +4278,18 @@ int ListenConnectionTCP(const char *host, long port, const char *label)
...
   if (endpoint.getUnixPath(&unixPath)) {
-    return ListenConnectionUnix(unixPath, label);
+    result = ListenConnectionUnix(unixPath, label);
   }
   else if (endpoint.getTCPHostAndPort(&host, &port)) {
-    return ListenConnectionTCP(host, port, label);
+    result = ListenConnectionTCP(host, port, label);
   }
-  return -1;
+  free(unixPath);
```
Ionic: Set them to `NULL` for good measure.

nxcomp/src/Loop.cpp
```diff
@@ -6739,10 +6742,16 @@ int ConnectToRemote(ChannelEndPoint &socketAddress)
     }
   }
 
+  free(unixPath);
```
Ionic: `NULL`

nxcomp/src/Loop.cpp
```diff
@@ -6739,10 +6742,16 @@ int ConnectToRemote(ChannelEndPoint &socketAddress)
...
+  free(hostName);
+
   return pFD;
 
 ConnectToRemoteError:
 
+  free(unixPath);
```
Ionic: `NULL`

nxcomp/src/ChannelEndPoint.cpp
uli42: that line was the reason for doing all this because it caused the double-free. If we keep it we have to NULL _spec, too. Which makes the code ugly again.

nxcomp/src/Loop.cpp
uli42: they are local and it is at the end of the function, I really don't see a point in NULLing here.

nxcomp/src/Loop.cpp
uli42: same here, I only adds more code for no benefit.

nxcomp/src/Loop.cpp
uli42: here, too

nxcomp/src/ChannelEndPoint.cpp
Ionic: Ah, true, so move back and `NULL`. That wouldn't really make the code ugly IMHO.

nxcomp/src/Loop.cpp
Ionic: It's always a good (defensive) measure to set `free`d data to NULL. It's a very cheap operation and prevents errors like exactly these if code is ever added after the `free` call.

A lot of software defines a macro such as
```
#define SAFE_FREE(ptr) do { free(ptr); ptr = NULL; } while (0)
```

and uses them to free memory.

uli42: Then let's introduce this macro, too. But why the while "loop"?

Ionic: https://stackoverflow.com/questions/154136/why-use-apparently-meaningless-do-while-and-if-else-statements-in-macros

uli42:
rebased, added proposed changes and finally introduced the SAFE_FREE() macro

Ionic (approved):

Issue #558, opened 2017-11-20
backport 3 randr memory issues


uli42:
From xorg upstream:

commit fc7fb5bbe1c8f787e53500b9a2ca4af815f310d1
Author: Giuseppe Bilotta <giuseppe.bilotta@gmail.com>
Date:   Thu Nov 9 10:21:21 2017 +0100

    randr: free crtc->outputs on destroy
    
    Reviewed-by: Adam Jackson <ajax@redhat.com>
    Signed-off-by: Giuseppe Bilotta <giuseppe.bilotta@gmail.com>

commit 16381d186e7c791031392ed8afcfd33009854e9e
Author: Giuseppe Bilotta <giuseppe.bilotta@gmail.com>
Date:   Thu Nov 9 10:21:20 2017 +0100

    randr: always realloc crtcs and outputs
    
    When the last crtc (resp. output) is destroyed, the rrScrPriv crtcs
    (resp. outputs) fields do not get cleared, which can lead to a situation
    where the private's numCrtcs (resp. numOutputs) field is zero, but the
    associated memory is still allocated. Just checking if numCrtcs (resp.
    numOutputs) is zero is thus not a good criteria to determine whetehr to
    use a realloc or a malloc.
    
    Since crtcs (resp. outputs) are NULL-initialized anyway, relying on
    numCrtcs (resp. numOutputs) is actually unnecessary, because
    reallocation of a NULL ptr is equivalent to a malloc anyway.
    
    Therefore, just use realloc() unconditionally, and ensure that the
    fields are properly initialized.
    
    Reviewed-by: Adam Jackson <ajax@redhat.com>
    Signed-off-by: Giuseppe Bilotta <giuseppe.bilotta@gmail.com>

commit fb5ee77b91a93e27801006be8ee34d27984e7fa6
Author: Giuseppe Bilotta <giuseppe.bilotta@gmail.com>
Date:   Thu Nov 9 10:21:19 2017 +0100

    randr: rrGetScreenResources: initialize memory
    
    Similarly to bb766ef11227bd8c71ac65845d1930edd0eda40d, ensure that the
    extra padding is set to 0.
    
    Reviewed-by: Adam Jackson <ajax@redhat.com>
    Signed-off-by: Giuseppe Bilotta <giuseppe.bilotta@gmail.com>

uli42:
the third one in this list is already in our code (by 3e315cd17a519dbcafff18d4294a6f7bb8070767)

Issue #542, opened 2017-10-31
xfixes: backport off-by-one fix


uli42:
commit 152090ce442e94de1ae920208a92931af6493c8c
Author: Adam Jackson <ajax@nwnk.net>
Date:   Wed Mar 15 16:33:12 2006 +0000

    Coverity #487: Check version number correctly.

diff --git a/xfixes/xfixes.c b/xfixes/xfixes.c
index 9364309..a0e656d 100755
--- a/xfixes/xfixes.c
+++ b/xfixes/xfixes.c
@@ -143,7 +143,7 @@ ProcXFixesDispatch (ClientPtr client)
     REQUEST(xXFixesReq);
     XFixesClientPtr pXFixesClient = GetXFixesClient (client);
 
-    if (pXFixesClient->major_version > NUM_VERSION_REQUESTS)
+    if (pXFixesClient->major_version >= NUM_VERSION_REQUESTS)
        return BadRequest;
     if (stuff->xfixesReqType > version_requests[pXFixesClient->major_version])
        return BadRequest;

Issue #541, opened 2017-10-31
Fix memory leak


uli42:
```
commit 5e106a71b9f8077216d41619402952b0005dd8a4
Author: Adam Jackson <ajax@nwnk.net>
Date:   Wed Mar 15 16:49:04 2006 +0000

    Coverity #807: Fix a memory leak in XFixesExpandRegion.

diff --git a/xfixes/region.c b/xfixes/region.c
index d317f73..68c7015 100755
--- a/xfixes/region.c
+++ b/xfixes/region.c
@@ -839,6 +839,7 @@ ProcXFixesExpandRegion (ClientPtr client)
            REGION_INIT (pScreen, &r, &pTmp[i], 0);
            REGION_UNION (pScreen, pDestination, pDestination, &r);
        }
+       xfree(pTmp);
     }
     if (ret == Success) 
        ret = client->noClientException;
```

frantracer:
Hello,
I was running the latest version so far (in this moment is commit 7f777e01e3bb8780ba91e70171a0aae78a95692e) and I found a memory leak in the nx-agent during the execution of any kind of application that streams data, like spotify or firefox with youtube. After half an hour aproximately the kernel had to kill it because it was using almost the 100% of the memory.

Do you know whether the problem I described is related to this issue?

uli42:
Probably not. But I encourage you to test it!

Vom Smartphone gesendet.

----- Ursprüngliche Nachricht -----
Von: "frantracer" <notifications@github.com>
Gesendet: ‎29.‎11.‎2017 17:25
An: "ArcticaProject/nx-libs" <nx-libs@noreply.github.com>
Cc: "Ulrich Sibiller" <uli42@gmx.de>; "Author" <author@noreply.github.com>
Betreff: Re: [ArcticaProject/nx-libs] Fix memory leak (#541)

Hello,
I was running the latest version so far (in this moment is commit 7f777e0) and I found a memory leak in the nx-agent during the execution of any kind of application that streams data, like spotify or firefox with youtube. After half an hour aproximately the kernel had to kill it because it was using almost the 100% of the memory.
Do you know whether the problem I described is related to this issue?

uli42:
Thinking again: Some of the leaks are in the logging code, the longer it runs the more it logs. So it could help. Please test!

Vom Smartphone gesendet.

----- Ursprüngliche Nachricht -----
Von: "Ulrich Sibiller" <ulrich.sibiller@gmail.com>
[...]

PR #581, opened 2017-12-03
Pr/various backports

uli42:
Fixes for #541, #542, #558 and some more fixes without issue

### DIFF
--- a/nx-X11/programs/Xserver/damageext/damageext.c
+++ b/nx-X11/programs/Xserver/damageext/damageext.c
@@ -296,7 +296,7 @@ ProcDamageDispatch (ClientPtr client)
     REQUEST(xDamageReq);
     DamageClientPtr pDamageClient = GetDamageClient (client);
 
-    if (pDamageClient->major_version > NUM_VERSION_REQUESTS)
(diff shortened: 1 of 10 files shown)


sunweaver (changes requested):
Thanks for backporting those fixes.

However, I would like to see the following commit msg pattern:

```
<nx-libs first-line msg (equals first line xorg msg)>

 Backported from X.org:

 <X.org commit header>

     <commit msg from X.org, indented by 4 chars>

 Backported-to-NX-by: <you and your mailaddress>

 Fixes this-and-that/project#<nnn>
```

For an example see: 
 https://github.com/ArcticaProject/nx-libs/commit/565421ba18a09d20620e01fc877915a8b09a3735

sunweaver:
@uli42: When porting patches from X.org: most important is having the X.org commit hash in the nx-libs commit msg text for later reference. THANKS!

uli42:
reworked, added some more backports, rebased

Ionic (approved):

Ionic:
Going to fixup some commit messages before the actual merge.

Ionic:
We need spaces after hashes so that github doesn't automatically link our bug reports to the mentioned numbers.

Also fixed up some commit authors.

PR #582, opened 2017-12-04
Reintroduce -co as no-op

uli42:
nxagent will abort if it encounters an unknown command line option. As
we want 3.6 to be a drop-in replacement for 3.5 we cannot simply drop old
options but must accept them as no-op.

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
+++ b/nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
@@ -161,6 +161,12 @@ sets key-click volume (allowable range: 0-100).
 sets the visual class for the root window of color screens.
 The class numbers are as specified in the X protocol.
 Not obeyed by all servers.
+.TP 8
+.B \-co \fIfilename\fP
+This used to be the option for specifying the path to the RGB color
+database file. As the RGB color database is now embedded into the
+binary this option has no effect but is kept for
+compatibility. Deprecated.
 .ig
 .TP 8
 .B \-config \fIfilename\fP
--- a/nx-X11/programs/Xserver/os/utils.c
+++ b/nx-X11/programs/Xserver/os/utils.c
@@ -571,6 +571,9 @@ void UseMsg(void)
     ErrorF("-c                     turns off key-click\n");
     ErrorF("c #                    key-click volume (0-100)\n");
     ErrorF("-cc int                default color visual class\n");
+#ifdef NXAGENT_SERVER
+    ErrorF("-co file               deprecated, has no effect\n");
+#endif
 #ifdef COMMANDLINE_CHALLENGED_OPERATING_SYSTEMS
     ErrorF("-config file           read options from file\n");
 #endif
@@ -776,6 +779,14 @@ ProcessCommandLine(int argc, char *argv[])
 	    else
 		UseMsg();
 	}
+#ifdef NXAGENT_SERVER
+	else if ( strcmp( argv[i], "-co") == 0)
+	{
+	    fprintf(stderr, "Warning: Ignoring deprecated command line option '%s'.\n", argv[i]);
+	    if(++i >= argc)
+	        UseMsg();
+	}
+#endif
 	else if ( strcmp( argv[i], "-core") == 0)
 	    CoreDump = TRUE;
 	else if ( strcmp( argv[i], "-dpi") == 0)


nx-X11/programs/Xserver/os/utils.c
```diff
@@ -571,6 +571,7 @@ void UseMsg(void)
     ErrorF("-c                     turns off key-click\n");
     ErrorF("c #                    key-click volume (0-100)\n");
     ErrorF("-cc int                default color visual class\n");
+    ErrorF("-co file               (compatibility option, has no effect)\n");
```
sunweaver: we should put the work "deprecated" in here and move it to the very end under a title "deprecated options, kept for compatibility only"

nx-X11/programs/Xserver/os/utils.c
```diff
@@ -776,6 +777,11 @@ ProcessCommandLine(int argc, char *argv[])
 	    else
 		UseMsg();
 	}
+	else if ( strcmp( argv[i], "-co") == 0)
+	{
+	    if(++i >= argc)
+	        UseMsg();
```
sunweaver: print a deprecation warning to stderr, if we get here... IMHO

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -161,6 +161,11 @@ sets key-click volume (allowable range: 0-100).
 sets the visual class for the root window of color screens.
 The class numbers are as specified in the X protocol.
 Not obeyed by all servers.
+.TP 8
+.B \-co \fIfilename\fP
+This used to be the option for specifying the name of RGB color
+database. As the RGB color database is now integrated this option has
+no effect but is kept for compatibility.
```
sunweaver: there should be a big DEPRECATED: here... Maybe under a separate headline...

sunweaver (changes requested):
See my inline comments...

nx-X11/programs/Xserver/os/utils.c
uli42: there's not really a good place to put those options in the usage message. Putting it at the end will raise more attention than wanted because calling e.g. `nxagent -h` will print the usage and the last lines will stay right before the user's eyes. Putting deprecated options at the top seem wrong, too. Putting in the middle is also awkward because we'd need to put it before the valid options `ddxUseMsg()` prints.

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
uli42: We have the bc option above which has  "Deprecated." I have done the same now.

nx-X11/programs/Xserver/os/utils.c
uli42: no, stderr output may cause irritations while it should simply do nothing

uli42:
improved and rebased

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -161,6 +161,11 @@ sets key-click volume (allowable range: 0-100).
 sets the visual class for the root window of color screens.
 The class numbers are as specified in the X protocol.
 Not obeyed by all servers.
+.TP 8
+.B \-co \fIfilename\fP
+This used to be the option for specifying the name of RGB color
```
Ionic: `s/name of/path to/` and "**the** RGB color"?

nx-X11/programs/Xserver/hw/nxagent/man/nxagent.1
```diff
@@ -161,6 +161,11 @@ sets key-click volume (allowable range: 0-100).
...
+database. As the RGB color database is now integrated this option has
```
Ionic: "embedded into the binary" instead of "integrated"?

nx-X11/programs/Xserver/os/utils.c
Ionic: We're using calls such as  `fprintf(stderr, "Warning: Ignoring deprecated command line option '%s'.\n",` in nxagent's options parsing, so are you sure outputting to `stderr` would be bad idea here?

uli42:
proposed changes applied and rebased. Code uses `#ifdef NXAGENT_SERVER` now.

Ionic (approved):

PR #588, opened 2017-12-04
P/xquerytree leak

uli42:


nx-X11/programs/Xserver/hw/nxagent/Events.c
```diff
@@ -426,7 +426,7 @@ void nxagentRemoteWindowsTree(Window window, int level)
   int i, j;
   unsigned long rootWin, parentWin;
   unsigned int numChildren;
-  unsigned long *childList;
+  unsigned long *childList = NULL;
 
   if (!XQueryTree(nxagentDisplay, window, &rootWin, &parentWin, &childList,
                       &numChildren))
```
woglinde, line 434: Hi Uli,

I think there is still a memleak here, when XQueryTree fails and childList is set.

uli42:
Could be, but I was unsure if that can happen at all. But I have added it now.

nx-X11/programs/Xserver/hw/nxagent/Events.c
```diff
@@ -3570,6 +3571,12 @@ int nxagentHandleReparentNotify(XEvent* X)
         result = XQueryTree(nxagentDisplay, DefaultRootWindow(nxagentDisplay),
                                 &root_return, &parent_return, &children_return, &nchildren_return);
 
+        if (children_return)
+        {
+          XFree(children_return);
+          children_return = NULL;
+        }
+
```
woglinde: I think this is wrong too, children_return is free'd here but used later on.

woglinde: I think this is wrong too, children_return is free'd here but proably used in line 3582.

nx-X11/programs/Xserver/hw/nxagent/Events.c
```diff
@@ -3647,21 +3649,28 @@ int nxagentHandleReparentNotify(XEvent* X)
...
         w = parentReturn;
       }
       while (True);
 
+      if (childrenReturn)
+      {
+        XFree(childrenReturn);
+        childrenReturn = NULL;
+      }
+
```
woglinde: I think this check and Xfree is not needed.  childrenReturn is always free'd in the loop, there is no break before it. So it will be always NULL here.

nx-X11/programs/Xserver/hw/nxagent/Events.c
woglinde: Ah okay I was wrong here. In the error case the malloc of the children has failed and we should not free it, because bad things can happen, when it is not null.

nx-X11/programs/Xserver/hw/nxagent/Events.c
uli42: right

uli42:
Ok, I will rework that. Should have gone to bed instead of creating this patch...

sunweaver:
This PR is probably not ready for review, right? Waiting with this...

uli42:
reworked and rebased

woglinde (dismissed):
Hi Uli,

I think we should only free children list if XQueryTree was successful, otherwise the list was not memory allocated. According to the malloc man page in glibc in the error case NULL is returned. That woud save us the extra nullcheck at the end of some blocks.

nx-X11/programs/Xserver/hw/nxagent/Window.c
```diff
@@ -1476,16 +1473,16 @@ void nxagentConfigureWindow(WindowPtr pWin, unsigned int mask)
           }
         }
         fprintf(stderr, "\n");
-
-        if (children_return)
-        {
-          XFree(children_return);
-        }
       }
```
woglinde, line 1476: I think it is okay to let xfree here directly after the while loop.

woglinde (approved):
Okay I learned quite a few things. The commit looks now good to me. We could avoid some extra checks, but that would mean more nested if's. So its okay to merge.

uli42:
Thank you!

nx-X11/programs/Xserver/hw/nxagent/Events.c
```diff
@@ -3185,7 +3187,7 @@ int nxagentCheckWindowConfiguration(XConfigureEvent* X)
     #endif
   }
 
-  if (result && nchildren_return)
+  if (children_return)
```
Ionic, line 3190: Good! While it may look like `result && nchildren_return` are equivalent to checking for `children_return` to not be`NULL`, there is a corner case in which `nchildren_return` could be `0`, but `children_return` would still have been allocated a `sizeof(Window)` memory block.

nx-X11/programs/Xserver/hw/nxagent/Window.c
Ionic: Yes, but the proposed version is also correct and falls in line with the other changes.

Ionic (approved):

PR #592, opened 2017-12-07
Backport `asprintf` and `scnprintf`  implementations.

Ionic:
`asprintf` is not available everywhere. One of such platforms is Windows.

Luckily, X.Org has a wrapper for this. We currently use X.Org's implementation of `Xprintf` unconditionally anyway, since NoMachine implemented a workaround for `vsnprintf` returning `-1` on truncation on some platforms (older glibc versions).

This backport should have happened way earlier when we first started using `asprintf`, but I  didn't think of it back then.

Also adds `(v)scnprintf` implementations while we're at it.

### DIFF
--- a/nx-X11/programs/Xserver/include/Imakefile
+++ b/nx-X11/programs/Xserver/include/Imakefile
@@ -14,6 +14,7 @@ all::
 depend::
 
 InstallDriverSDKNonExecFile(XIstubs.h,$(DRIVERSDKINCLUDEDIR))
+InstallDriverSDKNonExecFile(Xprintf.h,$(DRIVERSDKINCLUDEDIR))
 InstallDriverSDKNonExecFile(bstore.h,$(DRIVERSDKINCLUDEDIR))
 InstallDriverSDKNonExecFile(bstorestr.h,$(DRIVERSDKINCLUDEDIR))
 InstallDriverSDKNonExecFile(client.h,$(DRIVERSDKINCLUDEDIR))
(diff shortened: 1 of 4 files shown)


sunweaver (changes requested):
```
11:26 <@sunweaver> Ionic: we have omitted the _X_EXPORT "decorator" (or how they are called) everywhere 
                   else, so far.
11:27 < Ionic> not really
11:27 < Ionic> I've git grepped for it and it turned up in a lot of locations, so why not add it
11:30 <@sunweaver> yep.
11:30 <@sunweaver> ok.
11:32 <@sunweaver> Ionic: would it make sense to do the same wrapping as in X.orgs Xprintf.h?`
11:32 <@sunweaver> http://paste.debian.net/999615/
11:32 <@sunweaver> compared with diff -w
11:32 <@sunweaver> so, as I see it, the diff could be empty.
11:32 <@sunweaver> why not go there?
11:34 < Ionic> hum
11:34  -!- narenas <~Adium@88.12.19.210> has quit (Quit: Leaving.)
11:36 <@sunweaver> same for xprintf.c in Xserver/os.
11:36 <@sunweaver> also, why did you skip +Xvscnprintf()?
11:36 <@sunweaver> that would make the porting complete.
11:36 < Ionic> did I?
11:36 <@sunweaver> http://paste.debian.net/999617/
```

Ionic:
Merged in changes from global re-indentation commit and added `(v)scnprintf` backport.

sunweaver (approved):

sunweaver:
closed via https://github.com/ArcticaProject/nx-libs/commit/e54e42af93017ba6ab5f18a83e705a8e24ed3906

PR #594, opened 2017-12-10
Pr/fix shadow warning

uli42:
two more fixes for open issues. The second one is really an automake bug but is not fixed upstream yet. So I took the fix posted here: https://github.com/kimwalisch/primesieve/issues/16


sunweaver (approved):
LGTM

PR #597, opened 2017-12-11
Xserver/os/xprintf.c: Drop NoMachine's own implementation of Xvasprin…

sunweaver:
…tf(). Use stock implementation instead.

 * Reason from NoMachine's NX CHANGELOG for having its own implementation
   of Xvasprintf() (from around nxagent 3.3.0):

    - Fixed TR06G02225. The implementation of Xvprintf() has been reviewed
      to work on more platforms. Previous implementation caused a failure
      in the build of keyboard map on some platform like Solaris 8 and 9.
    - Fixed TR03G02198. Reimplemented Xvprintf() in Xserver/os to handle
      the case in which vsnprintf returned -1.

 * Reason for removing it again:

    - Fixes segfaults in SetDefaultFontPath when launching nxagent.
    - All locations in Xserver/** using the code check for return value
      of -1.
    - Solaris 8 and 9 are beyond of our support scope.
    - Keep more in sync with X.org.
    - Good moment to drop more GPL-2 code from Xserver's code base.

### DIFF
--- a/nx-X11/programs/Xserver/os/xprintf.c
+++ b/nx-X11/programs/Xserver/os/xprintf.c
@@ -89,54 +89,7 @@
 #undef  TEST
 #undef  DEBUG
 
-#define START_SIZE  256
-#define END_SIZE   2048
-
-int
-Xvasprintf(char **ret, const char *_X_RESTRICT_KYWD format, va_list va)
-{
-    char *newret;
-    int size;
-    int r;
-
-    size = 0;
-
-    for (;;)
-    {
-      if (size == 0)
-      {
-        *ret = (char *)malloc(START_SIZE);
-        if (*ret == NULL)
-          return -1;
-        size = START_SIZE;
-      }
-      else if (size < END_SIZE &&
-                   (newret = (char *) realloc(*ret, 2 * size)) != NULL)
-      {
-        *ret = newret;
-        size = 2 * size;
-      }
-      else
-      {
-        free(*ret);
-        return -1;
-      }
-
-      r = vsnprintf(*ret, size, format, va);
-
-      if (r == -1 || r == size || r > size || r == size - 1)
-      {
-        continue;
-      }
-      else
-      {
-        (*ret)[r] = 0;
-        return r;
-      }
-    }
-}
-
-#else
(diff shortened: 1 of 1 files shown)


sunweaver:
```
09:34 <@sunweaver> https://github.com/ArcticaProject/nx-libs/pull/597
09:34  -!- uli421 <~sibiller@blackhole.science-computing.de> has joined #arctica
09:35 <@sunweaver> ok.
09:35 < Ionic> NAK
09:35 < Ionic> TR03G02198 is specific so some older glibc versions
09:35 < Ionic> not sure what exactly
09:35 < Ionic> according to the man page...
09:35 <@sunweaver> HOW OLD???
09:36 <@sunweaver> X.org does not have it...
09:36 <@sunweaver> so...
09:36 < Ionic> wait
09:36 <@sunweaver> it works here on a Debian stretch system...
09:36 <@sunweaver> segfault is gone with #597.
09:36 <@sunweaver> but anyway! Go and get your appointment in time.
09:37 < Ionic> according to the man page:
09:37 < Ionic> "The glibc implementation of the functions snprintf() and vsnprintf() conforms to the C99 standard, that is, behaves as described above, since glibc version 2.1. Until glibc 2.0.6, they would return -1 when the output was truncated."
09:37 < Ionic> I guess we should just drop it
09:37 < Ionic> glibc 2.0.6- sounds ancient
09:37 <@sunweaver> that's what #597 does.
09:37 * sunweaver nods heavily...
09:37 < Ionic> I actually wanted to look up when 2.1 was released
09:38 < Ionic> in 1999
09:38 < Ionic> well
09:38 * sunweaver is doing that...
09:38 < Ionic> new enough
09:38 <@sunweaver> ah...
09:38 <@sunweaver> well, indeed.
09:38 < Ionic> okay
09:38 < Ionic> in that case it's probably really best to drop it completely
09:38 < Ionic> see you later though
09:38 <@sunweaver> I'll do the merge now, so we have an nxagent that does not segfault anymore.
09:38 <@sunweaver> yeah, yee you.
```

sunweaver:
Approved by Ionic on IRC.

PR #602, opened 2017-12-13
Pr/pre rrxinerama fallback

uli42:
part of #475, was forgotten

The current version is not resizable via xrandr. This patch re-enables that feature if Xinerama is disabled (`-rrxinerama`)


sunweaver (approved):
Compared with 3.5.0
Checked integration into 3.6.x
LGTM.

PR #607, opened 2017-12-14
allow a custom version string

uli42:
run `make NX_VERSION_CUSTOM=bla` to get a custom version string into nxagent and nxproxy.

We can discuss how the output should look like, I am not really satiesfied with the current solution:

Currently you will get
```
    NXPROXY - Version 3.5.99.22
    NXAGENT - Version 3.5.99.22
```
when compiled _without_ a custom version and 
```
    NXPROXY - Version myvers (3.5.99.22)
    NXAGENT - Version myvers (3.5.99.22)
```
when compiled _with_ a custom version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,10 @@ ifneq ($(shell pkg-config --exists fontutil && echo yes), yes)
     $(warning fontutil devel package missing, using imake default values)
 endif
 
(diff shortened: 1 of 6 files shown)


sunweaver (approved):
LGTM

Issue #612, opened 2017-12-21
re-write abstract socket code


uli42:
- in line 3937 addr_length is 0, so the memset will not do anything. Should be fixed.
- comment line 3940: why +3? 
- before line 3950: delete xServerAddrUnix created in line  3907
- around 3956: delete xServerAddrABSTRACT created in line 3936
- around 3956: it is unclear to me if abstract sockets must be closed after use at all. However, the if branch of the code calls close (line 3948) while the else branch does not.
- line 3939: ensure that memcpy does not copy too many bytes. Could use snprintf or similar. The size of `xServerAddrABSTRACT -> sun_path` is surely limited somehow.
- line 3933: there's no need to use a directory path in abstract sockets.
- maybe more?

nxcomp/src/Loop.cpp:
```
   3907     sockaddr_un *xServerAddrUNIX = new sockaddr_un;
   3908 
   3909     addr_family = AF_UNIX;
   3910     xServerAddrUNIX -> sun_family = AF_UNIX;
  ...
   3929     #ifdef __linux__
   3930     int testSocketFD;
   3931     testSocketFD = socket(addr_family, SOCK_STREAM, PF_UNSPEC);
   3932 
   3933     int len = sprintf(unixSocketName + 1, "/tmp/.X11-unix/X%d", xPort);
   3934     unixSocketName[0] = '\0';
   3935 
   3936     sockaddr_un *xServerAddrABSTRACT = new sockaddr_un;
   3937     memset(xServerAddrABSTRACT, 0, addr_length);
   3938     xServerAddrABSTRACT -> sun_family = AF_UNIX;
   3939     memcpy(xServerAddrABSTRACT -> sun_path, unixSocketName, len+1);
   3940     addr_length = len +3;
   3941 
   3942     int ret = connect(testSocketFD, (struct sockaddr *) xServerAddrABSTRACT, addr_length);
   3943     if (ret == 0) {
   3944 
   3945         cerr << "Info" << ": Using abstract X11 socket in kernel namespace "
   3946              << "for accessing DISPLAY=:" << xPort << ".\n";
   3947 
   3948         close(testSocketFD);
   3949         addr = (sockaddr *) xServerAddrABSTRACT;
   3950         return;
   3951 
   3952     } else {
   3953 
   3954         cerr << "Info" << ": Falling back to file system X11 socket "
   3955              << "for accessing DISPLAY=:" << xPort << ".\n";
   3956 
   3957     #endif
   3958 
```

PR #615, opened 2017-12-27
Pr/fix abstract

uli42:
all fixes for #612 + some more

### DIFF
--- a/nxcomp/src/ChannelEndPoint.cpp
+++ b/nxcomp/src/ChannelEndPoint.cpp
@@ -273,8 +273,10 @@ ChannelEndPoint::getTCPHostAndPort(char **host, long *port) const {
   char *h = NULL;
   ssize_t h_len;
 
-  if (host) *host = NULL;
-  if (port) *port = 0;
+  if (host)
+    *host = NULL;
+  if (port)
+    *port = 0;
 
   if (getPort(&p)) {
     h_len = 0;
--- a/nxcomp/src/ChannelEndPoint.cpp
+++ b/nxcomp/src/ChannelEndPoint.cpp
@@ -113,13 +113,19 @@ ChannelEndPoint::setSpec(const char *hostName, long port) {
 bool
 ChannelEndPoint::getSpec(char **socketUri) const {
 
-  if (socketUri) *socketUri = NULL;
+  if (socketUri)
+  {
+    *socketUri = NULL;
+  }
+  else
+  {
+    return false;
+  }
 
   char *unixPath = NULL;
   char *hostName = NULL;
   long port = -1;
 
-  char *newSocketUri = NULL;
   int length = -1;
 
   if (getUnixPath(&unixPath))
@@ -133,17 +139,21 @@ ChannelEndPoint::getSpec(char **socketUri) const {
 
   if (length > 0)
   {
-    newSocketUri = static_cast<char *>(calloc(length + 1, sizeof(char)));
-    if (isUnixSocket())
-      snprintf(newSocketUri, length+1, "unix:%s", unixPath);
-    else
-      snprintf(newSocketUri, length+1, "tcp:%s:%ld", hostName, port);
+    char *newSocketUri = static_cast<char *>(calloc(length + 1, sizeof(char)));
 
-    if (socketUri)
+    if (newSocketUri)
+    {
+      if (isUnixSocket())
+        snprintf(newSocketUri, length+1, "unix:%s", unixPath);
+      else
+        snprintf(newSocketUri, length+1, "tcp:%s:%ld", hostName, port);
+
       *socketUri = strdup(newSocketUri);
+
+      SAFE_FREE(newSocketUri);
+    }
   }
 
-  SAFE_FREE(newSocketUri);
   SAFE_FREE(unixPath);
   SAFE_FREE(hostName);
 
--- a/nxcomp/src/ChannelEndPoint.cpp
+++ b/nxcomp/src/ChannelEndPoint.cpp
@@ -207,7 +207,10 @@ ChannelEndPoint::getPort(long *port) const {
 bool
 ChannelEndPoint::getUnixPath(char **unixPath) const {
 
-  if (unixPath) *unixPath = NULL;
+  if (unixPath)
+    *unixPath = NULL;
+  else
+    return false;
 
   long p;
   char *path = NULL;
@@ -227,8 +230,7 @@ ChannelEndPoint::getUnixPath(char **unixPath) const {
       return false;
   }
 
-  if (unixPath)
-    *unixPath = strdup(path);
+  *unixPath = strdup(path);
 
   return true;
 }
--- a/nxcomp/src/ChannelEndPoint.cpp
+++ b/nxcomp/src/ChannelEndPoint.cpp
@@ -180,8 +180,6 @@ ChannelEndPoint::setDefaultUnixPath(char *path) {
 
   if (path && strlen(path))
     defaultUnixPath_ = strdup(path);
-  else
-    defaultUnixPath_ = NULL;
 
   isUnix_ = getUnixPath();
 }
--- a/nxcomp/src/EncodeBuffer.cpp
+++ b/nxcomp/src/EncodeBuffer.cpp
@@ -70,6 +70,13 @@ EncodeBuffer::EncodeBuffer()
   initialSize_   = ENCODE_BUFFER_DEFAULT_SIZE;
   thresholdSize_ = ENCODE_BUFFER_DEFAULT_SIZE << 1;
   maximumSize_   = ENCODE_BUFFER_DEFAULT_SIZE << 4;
+
+  #ifdef VALGRIND
+
+  memset(buffer_, '\0', size_);
+
+  #endif
+
 }
 
 EncodeBuffer::~EncodeBuffer()
--- a/nxcomp/src/EncodeBuffer.cpp
+++ b/nxcomp/src/EncodeBuffer.cpp
@@ -108,7 +108,15 @@ void EncodeBuffer::fullReset()
     size_ = initialSize_;
 
     buffer_ = new unsigned char[size_ + ENCODE_BUFFER_PREFIX_SIZE +
-                                    ENCODE_BUFFER_POSTFIX_SIZE] + ENCODE_BUFFER_PREFIX_SIZE;
+                                    ENCODE_BUFFER_POSTFIX_SIZE];
+
+    #ifdef VALGRIND
+
+    memset(buffer_, '\0', size_ + ENCODE_BUFFER_PREFIX_SIZE + ENCODE_BUFFER_POSTFIX_SIZE);
+
+    #endif
+
+    buffer_ += ENCODE_BUFFER_PREFIX_SIZE;
   }
 
   end_ = buffer_ + size_;
--- a/nxcomp/src/Loop.cpp
+++ b/nxcomp/src/Loop.cpp
@@ -3861,7 +3861,8 @@ void SetupDisplaySocket(int &addr_family, sockaddr *&addr,
 
   #ifdef __APPLE__
 
-  if ((strncasecmp(display, "/tmp/launch", 11) == 0) || (strncasecmp(display, "/private/tmp/com.apple.launchd", 30) == 0))
+  if ((strncasecmp(display, "/tmp/launch", 11) == 0) ||
+      (strncasecmp(display, "/private/tmp/com.apple.launchd", 30) == 0))
   {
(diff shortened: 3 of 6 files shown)


nxcomp/src/Loop.cpp
```diff
@@ -3944,12 +3944,13 @@ void SetupDisplaySocket(int &addr_family, sockaddr *&addr,
     addr_length = len +3;
 
     int ret = connect(testSocketFD, (struct sockaddr *) xServerAddrABSTRACT, addr_length);
+    close(testSocketFD);
```
Ionic: I don't understand this commit message... the non-abstract part doesn't use a testing socket at all.

The commit is fine, though, since the socket should be closed in the error case as well, what we previously didn't.

nxcomp/src/ChannelEndPoint.cpp
```diff
@@ -140,7 +140,10 @@ ChannelEndPoint::getSpec(char **socketUri) const {
       snprintf(newSocketUri, length+1, "tcp:%s:%ld", hostName, port);
 
     if (socketUri)
+    {
+      SAFE_FREE(*socketUri);
```
Ionic: Uh, this might be dangerous.

I'd drop this commit, since it's better to let callers handle their memory and make sure that `socketUri` is free'd in the callers.

nxcomp/src/Loop.cpp
```diff
@@ -3786,6 +3786,7 @@ void SetupDisplaySocket(int &addr_family, sockaddr *&addr,
                            unsigned int &addr_length)
 {
   addr_family = AF_INET;
+  delete addr;
```
Ionic: Drop this commit. Same reasoning here.

nxcomp/src/Loop.cpp
```diff
@@ -3587,15 +3589,10 @@ int SetupAuthInstance()
...
-        /* POSIX/SUS does not specify a length.
-         * BSD derivatives generally support 104 bytes, other systems may be more constrained.
-         * If you happen to run into such systems, extend this section with the appropriate limit.
-         */
-        const int launchdAddrNameLength = 104;
-        #endif
+        // determine the maximum number of characters that fit into struct
+        // sockaddr_un's sun_path member
+        const int launchdAddrNameLength =
+          sizeof(struct sockaddr_un) - offsetof(sockaddr_un, sun_path);
```
Ionic: Use `offsetof(struct sockaddr_un, sun_path)`

nxcomp/src/Loop.cpp
```diff
@@ -3907,6 +3904,11 @@ void SetupDisplaySocket(int &addr_family, sockaddr *&addr,
     // UNIX domain port.
     //
 
+    // determine the maximum number of characters that fit into struct
+    // sockaddr_un's sun_path member
+    unsigned int maxlen_un =
```
Ionic: `std::size_t`

nxcomp/src/Loop.cpp
```diff
@@ -3941,10 +3943,28 @@ void SetupDisplaySocket(int &addr_family, sockaddr *&addr,
     sockaddr_un *xServerAddrABSTRACT = new sockaddr_un;
     memset(xServerAddrABSTRACT, 0, sizeof(struct sockaddr_un));
     xServerAddrABSTRACT -> sun_family = AF_UNIX;
-    // FIXME: ensure sun_path can hold len+1 bytes!
-    memcpy(xServerAddrABSTRACT -> sun_path, unixSocketName, len+1);
-    // FIXME: comment why + 3?
-    addr_length = len + 3;
+
+    if (maxlen_un < (unsigned int)len + 1)
```
Ionic: `std::size_t`

nxcomp/src/Loop.cpp
```diff
@@ -6535,12 +6564,18 @@ int PrepareProxyConnectionUnix(char** path, int* timeout, int* proxyFileDescript
 
   /* FIXME: Add socket file existence and permission checks */
 
+
   *proxyFileDescriptor = -1;
   *reason = -1;
 
+  // determine the maximum number of characters that fit into struct
+  // sockaddr_un's sun_path member
+  const int sockpathlen =
```
Ionic: `const std::size_t`

nxcomp/src/Loop.cpp
```diff
@@ -6535,12 +6564,18 @@ int PrepareProxyConnectionUnix(char** path, int* timeout, int* proxyFileDescript
...
+    sizeof(struct sockaddr_un) - offsetof(sockaddr_un, sun_path);
+
   sockaddr_un addr;
   addr.sun_family = AF_UNIX;
-  strncpy(addr.sun_path, *path, 108 - 1);
+  strncpy(addr.sun_path, *path, sockpathlen - 1);
```
Ionic: Should actually use `snprintf` here, since `strncpy` is not NULL-terminating the string in the worst case.

nxcomp/src/Loop.cpp
```diff
@@ -8844,6 +8844,8 @@ int ParseCommandLineOptions(int argc, const char **argv)
 
 int ParseBindOptions(char **host, int *port)
```
Ionic: This function is completely unused... I'm not sure if we should let the caller handle memory cleanup or do it in here. In general I'd say caller...

Ionic (changes requested):

nxcomp/src/Loop.cpp
uli42: The testsocket is created some lines above the connect statement by a socket() call, which returns a file descriptor. That code path is always executed for unix sockets, be it abstract or non-abstract.

nxcomp/src/Loop.cpp
uli42: Hmm, interesting. Should we drop it?

Regarding the memory cleanup: I free the memory just before overwriting the old value. In this special case there isn't an outside call where I could free it...

nxcomp/src/Loop.cpp
uli42: ok

nxcomp/src/Loop.cpp
uli42: ok

uli42:
I have reworked the pr, please re-check

nxcomp/src/Loop.cpp
```diff
@@ -3141,6 +3141,9 @@ int InitBeforeNegotiation()
     // Get ready to open the local display.
     //
 
+    delete xServerAddr;
+    xServerAddr = NULL;
```
Ionic, line 3147: `SAFE_FREE` instead?

nxcomp/src/Loop.cpp
Ionic: Ping.

Ionic: Also, this should probably be `len + 2` to account for the leading `NULL` character, right?

nxcomp/src/Loop.cpp
```diff
@@ -3943,10 +3945,28 @@ void SetupDisplaySocket(int &addr_family, sockaddr *&addr,
...
+              << "' is too long!" << std::flush;
+
+      delete [] display;
+      delete xServerAddrABSTRACT;
+
+      HandleCleanup();
+    }
+
+    // copy including the leading '\0'
+    memcpy(xServerAddrABSTRACT -> sun_path, unixSocketName, len + 1);
```
Ionic, line 3961: Uh and we've been probably lucky so far to not hit this bug, but this should also be `len + 2` to account for the trailing and leading `NULL` characters?!

nxcomp/src/Loop.cpp
```diff
@@ -3943,10 +3945,28 @@ void SetupDisplaySocket(int &addr_family, sockaddr *&addr,
...
+
+    // man 7 unix:
+    // "an abstract socket address is distinguished (from a
+    // pathname socket) by the fact that sun_path[0] is a null byte
+    // ('\0').  The socket's address in this namespace is given by the
+    // additional bytes in sun_path that are covered by the specified
+    // length of the address structure.
```
Ionic: Add terminating `"`.

nxcomp/src/Loop.cpp
```diff
@@ -3943,10 +3945,28 @@ void SetupDisplaySocket(int &addr_family, sockaddr *&addr,
...
+    addr_length = offsetof(struct sockaddr_un, sun_path) + len + 1;
```
Ionic: `+ 2`?

Ionic: Thinking further about it, I guess that should just plainly be `sizeof(struct sockaddr_un)` indefinitely, since that's what we pass to `connect()`.

nxcomp/src/EncodeBuffer.cpp
```diff
@@ -70,6 +70,13 @@ EncodeBuffer::EncodeBuffer()
   initialSize_   = ENCODE_BUFFER_DEFAULT_SIZE;
```
Ionic, line 70: Faulty "an" in commit message.

nxcomp/src/Loop.cpp
Ionic: > Hmm, interesting. Should we drop it?

Probably. It has been unused from the beginning, as far as I can tell. Same goes for `bindHost` and `bindPort`. Whatever the intention has been, it was either dropped partially at some point or never correctly implemented in the first place.

Not as part of this PR, though.

> Regarding the memory cleanup: I free the memory just before overwriting the old value. In this special case there isn't an outside call where I could free it...

Yes, so drop the commit completely.

The problem with freeing stuff from outside is that you cannot know what data is passed in. It might be a valid pointer to already allocated data, `NULL` (both of which would be fine) or just random garbage. Thus it shall be the caller's responsibility to handle this data.

Easy breaking example:
```C++
void some_function () {
  char *host;
  int port;

  ParseBindOptions (&host, &port);
}
```

If `ParseBindOptions` were to `free` `*host`, which at this point contains random data from the stack and thus can point to any location, things would fail badly.

nxcomp/src/Loop.cpp
```diff
@@ -13491,15 +13492,8 @@ int ParseArg(const char *type, const char *name, const char *value)
     return -1;
   }
 
-  char *string = new char[strlen(value)];
-
-  strncpy(string, value, strlen(value) - 1);
```
Ionic: No, dropping that is not equivalent. The idea is to drop the trailing character and only retain the number.

Probably replace it with

```C++
  char *string = new char[strlen(value)];
  snprintf(string, strlen(value), "%s", value);
```

and keep using `string`.

Ionic (changes requested):
Ping on some older entries + found some other stuff that doesn't look good yet.

nxcomp/src/Loop.cpp
uli42: No, we have created it using 'new' so we delete it via 'delete'. In this special case this might not be necessary but more correct. If one day we use a class instead of a struct for xServerAddr we want the destructor being called. I was thinking about SAFE_DELETE but we'd need also SAFE_DELETE_[] or something which  is ugly.

nxcomp/src/Loop.cpp
uli42: no, there's such thing as a trailing NULL for the abstract case. The string length is solely determined via the addr_length.

nxcomp/src/Loop.cpp
uli42: ok

nxcomp/src/Loop.cpp
uli42: no, abstract sockets are special, see above.

nxcomp/src/Loop.cpp
uli42: good spot! Fixed!

nxcomp/src/EncodeBuffer.cpp
uli42: fixed

nxcomp/src/Loop.cpp
uli42: commit dropped

nxcomp/src/Loop.cpp
uli42: ok

nxcomp/src/Loop.cpp
uli42: have replaced almost all strncpys in nxcomp now

nxcomp/src/Loop.cpp
Ionic: Oops, true, I didn't actually know this. Abstract sockets aren't `NULL` terminated...

Ionic (approved):

Issue #572, opened 2017-11-27
Lots of compiler warnings in nxcompshad


uli42:
On Ubuntu 16.04.2 LTS with gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4) the normal make in the main directory shows these warnings:

```
make[5]: Verzeichnis „/home/uli/work/nx/ArcticaProject/nx-libs/nxcompshad/src“ wird betreten
  CXX      Core.lo
In file included from Core.h:31:0,
                 from Core.cpp:38:
Logger.h: In function ‘void logUser(const char*, ...)’:
Logger.h:88:32: warning: format not a string literal, argument types not checked [-Wformat-nonliteral]
   logger.user(format, arguments);
                                ^
Logger.h: In function ‘void logWarning(const char*, const char*, ...)’:
Logger.h:112:41: warning: format not a string literal, argument types not checked [-Wformat-nonliteral]
   logger.warning(name, format, arguments);
                                         ^
Core.cpp: In member function ‘int CorePoller::isChanged(int (*)(void*), void*, int*)’:
Core.cpp:340:45: warning: narrowing conversion of ‘curLine’ from ‘int’ to ‘short int’ inside { } is ill-formed in C++11 [-Wnarrowing]
     XRectangle rect = {0, curLine, width_, 1};
                                             ^
Core.cpp:340:36: warning: narrowing conversion of ‘((CorePoller*)this)->CorePoller::width_’ from ‘unsigned int’ to ‘short unsigned int’ inside { } is ill-formed in C++11 [-Wnarrowing]
     XRectangle rect = {0, curLine, width_, 1};
                                    ^
Core.cpp:448:75: warning: narrowing conversion of ‘left’ from ‘int’ to ‘short int’ inside { } is ill-formed in C++11 [-Wnarrowing]
         XRectangle rect = {left, start, right - left + 1, last - start + 1};
                                                                           ^
Core.cpp:448:75: warning: narrowing conversion of ‘start’ from ‘int’ to ‘short int’ inside { } is ill-formed in C++11 [-Wnarrowing]
Core.cpp:448:54: warning: narrowing conversion of ‘((right - left) + 1)’ from ‘int’ to ‘short unsigned int’ inside { } is ill-formed in C++11 [-Wnarrowing]
         XRectangle rect = {left, start, right - left + 1, last - start + 1};
                                                      ^
[...]

PR #619, opened 2017-12-27
nxcompshad: silence GCC warnings

uli42:
Fixes ArcticaProject/nx-libs#572

### DIFF
--- a/nxcompshad/src/Logger.h
+++ b/nxcompshad/src/Logger.h
@@ -29,6 +29,9 @@
 #include <cerrno>
 #include <cstdarg>
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-nonliteral"
+
 //
 // Error handling macros.
 //
@@ -163,5 +166,7 @@ static inline void logDump(const char *name, const char *data, int size)
 
   #endif
 }
+
+#pragma GCC diagnostic pop
 
 #endif /* Logger_H */
--- a/nxcompshad/src/Core.cpp
+++ b/nxcompshad/src/Core.cpp
@@ -337,7 +337,10 @@ int CorePoller::isChanged(int (*checkIfInputCallback)(void *), void *arg, int *s
       }
     }
 
-    XRectangle rect = {0, curLine, width_, 1};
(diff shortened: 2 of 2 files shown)


nxcompshad/src/Logger.h
```diff
@@ -79,6 +82,7 @@ static inline void logDebug(const char *name, const char *format, ...) \
 static inline void logDump(const char *name, const char *data, int size) \
     __attribute__((__unused__));
 
+
```
Ionic: Stray newline?

nxcompshad/src/Logger.h
```diff
@@ -29,6 +29,9 @@
 #include <cerrno>
 #include <cstdarg>
 
+#pragma GCC diagnostic push
```
Ionic, line 32: This will only work for GCC... and maybe fail on older versions. I'm not sure since when `-Wformat-nonliteral` exists. Looks like it has been around since at least 3.4, so that's good enough.

We may want this for clang, too?

Let's go for

```C
#ifdef __clang__
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wformat-nonliteral"
#else
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wformat-nonliteral"
#endif
```
and

```C
#ifdef __clang__
#pragma clang diagnostic pop
#else
#pragma GCC diagnostic pop
#endif
```

uli42: according to the clang documentation clang also understands gcc pragmas(https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-via-pragmas):
 
"Clang can also control what diagnostics are enabled through the use of pragmas in the source code. This is useful for turning off specific warnings in a section of source code. Clang supports GCC’s pragma for compatibility with existing source code, as well as several extensions."

See also https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-via-pragmas

nxcompshad/src/Logger.h
uli42: fixed (and rebased)

Ionic (approved):

Issue #616, opened 2017-12-27
nxcomp: switch from ctime() to ctime_r()/ctime_s()


uli42:
man ctime:
```
The return value points to a statically allocated string which might be overwritten by subsequent calls to any of the date and time functions.  The function also  sets  the  external  variables
       tzname,  timezone,  and  daylight (see tzset(3)) with information about the current timezone.  The reentrant version ctime_r() does the same, but stores the string in a user-supplied buffer which should have room
       for at least 26 bytes.  It need not set tzname, timezone, and daylight.
```

The code in Timestamp.cpp overwrites the buffer returned by ctime itself, which IMHO violates the spec from the manpage. Also the ctime_new has a strange size, it is too large for the sprintf and too small to fully cover ctime_now.

```
//
// This is especially dirty. 
//

char *strMsTimestamp(const T_timestamp &ts)
{
  char *ctime_now = ctime((time_t *) &ts.tv_sec);

  char ctime_new[25];

  sprintf(ctime_new, "%.8s:%3.3f", ctime_now + 11,
              (float) ts.tv_usec / 1000);

  strncpy(ctime_now, ctime_new, 24);

  return ctime_now;
}
```

PR #623, opened 2017-12-30
Use ctime_s/ctime_r instead of ctime in nxcomp.

Ionic:
As mentioned in #616, the timestamping implementation in nxcomp is using ctime(), which isn't particularly modern or working well in multi-threaded code (should the need arise, anyway.)

Switch to ctime_r or ctime_s (with ctime_s being preferred if available), change to code to use on-stack `char*` buffers that do not need to be `free`'d and return std::string objects that will be copied around and leak no memory - without ugly code changes to "buffer" heap-allocated `char` arrays.

### DIFF
--- a/nxcomp/src/Timestamp.cpp
+++ b/nxcomp/src/Timestamp.cpp
@@ -44,34 +44,55 @@
 
 T_timestamp timestamp;
 
-//
-// The following functions all use the ctime
-// static buffer from the C library.
-//
-
-char *strTimestamp(const T_timestamp &ts)
+std::string strTimestamp(const T_timestamp &ts)
 {
-  char *ctime_now = ctime((time_t *) &ts.tv_sec);
+  std::string ret;
 
-  ctime_now[24] = '\0';
+  char ctime_now[26] = { };
+  bool err = true;
 
-  return ctime_now;
+#if HAVE_CTIME_S
+  errno_t retval = ::ctime_s(ctime_now, sizeof(ctime_now), static_cast<const time_t*>(&ts.tv_sec));
+
+  if (retval != 0)
+#else
+  char *retval = ::ctime_r(static_cast<const time_t*>(&ts.tv_sec), ctime_now);
+
+  if (!(retval))
+#endif
+  {
+    std::cerr << "WARNING: converting time to string failed." << std::endl;
+  }
+  else
+  {
+    /* Replace newline at position 25 with a NULL byte. */
+    ctime_now[24] = '\0';
+
+    ret = ctime_now;
+  }
+
+  return ret;
 }
 
 //
-// This is especially dirty. 
+// This is especially dirty.
 //
 
-char *strMsTimestamp(const T_timestamp &ts)
+std::string strMsTimestamp(const T_timestamp &ts)
 {
-  char *ctime_now = ctime((time_t *) &ts.tv_sec);
+  std::string ret;
 
-  char ctime_new[25];
+  std::string ctime_now = strTimestamp(ts);
 
-  sprintf(ctime_new, "%.8s:%3.3f", ctime_now + 11,
-              (float) ts.tv_usec / 1000);
+  if (!(ctime_now.empty()))
+  {
+    char ctime_new[26] = { };
 
-  strncpy(ctime_now, ctime_new, 24);
+    snprintf(ctime_new, sizeof(ctime_new), "%.8s:%3.3f",
+             ctime_now.c_str() + 11, static_cast<float>(ts.tv_usec) / 1000);
 
-  return ctime_now;
+    ret = ctime_new;
+  }
+
+  return ret;
 }
--- a/nxcomp/configure.ac
+++ b/nxcomp/configure.ac
@@ -94,6 +94,23 @@ std::tm tm = *std::localtime(&t);
                              [Use std::put_time to format times, must be made available by the compiler if turned on.])],
                   [AC_MSG_RESULT([no])])
 
+# Check if ::ctime_s is available.
+AC_MSG_CHECKING([if ::ctime_s is available])
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM(
+[[
+#define __STDC_WANT_LIB_EXT1__ 1
+#include <ctime>
+]],
+[[
+time_t res = time(NULL);
+char str[26] = { };
+::ctime_s(str, sizeof(str), &res);
+]])],
+                  [AC_MSG_RESULT([yes])
+                   AC_DEFINE(HAVE_CTIME_S, [1],
+                             [Use ::ctime_s to format times, must be made available by the compiler if turned on.])],
+                  [AC_MSG_RESULT([no])])
+
 AC_ARG_ENABLE([debug],
               [AS_HELP_STRING([--enable-debug],
                               [enable to get info session log output (disabled by default)])],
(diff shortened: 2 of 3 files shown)


nxcomp/src/Timestamp.cpp
```diff
@@ -44,34 +44,55 @@
...
-              (float) ts.tv_usec / 1000);
+    snprintf(ctime_new, sizeof(ctime_new), "%.8s:%3.3f",
+             ctime_now.c_str() + 11, static_cast<float>(ts.tv_usec) / 1000);
 
-  strncpy(ctime_now, ctime_new, 24);
+    ret = ctime_new;
+  }
 
-  return ctime_now;
+  return ret;
```
uli42, line 97: ret is uninitilaized here if ctime_now is empty!

nxcomp/src/Timestamp.cpp
```diff
@@ -44,34 +44,55 @@
...
+  if (err)
+  {
+    std::cerr << "WARNING: converting time to string failed." << std::endl;
+  }
+
+  /* Replace newline at position 25 with a NULL byte. */
   ctime_now[24] = '\0';
 
-  return ctime_now;
+  ret = ctime_now;
```
uli42: return ctime_now immediately, there's not need for ret AFAICS

nxcomp/src/Timestamp.cpp
```diff
@@ -44,34 +44,55 @@
...
+  errno_t retval = ::ctime_s(ctime_now, 26, static_cast<const time_t*>(&ts.tv_sec));
+
+  err = (retval != 0);
+#else
+  char *retval = ::ctime_r(static_cast<const time_t*>(&ts.tv_sec), ctime_now);
 
+  err = (!(retval));
+#endif
+
+  if (err)
```
uli42: drop err and move the if into the ifdef

uli42 (changes requested):
generally I agree with using std::string, but I fear this will be much slower than using the char pointer ctime reports. What do you think?

nxcomp/src/Timestamp.cpp
Ionic: Why? That would duplicate the code for no good reason, since it would have to be in both the `#if` and `#else` branches.

nxcomp/src/Timestamp.cpp
Ionic: Actually, there was a good reason why I intended to use `ret`, which I have done wrong the first time around. `ret` should only be set to `ctime_now` if the call to `ctime_{r,s}` actually succeeded, else left as-is.

nxcomp/src/Timestamp.cpp
Ionic: `std::string` is a C++ class and we can thus trust it to initialize its internal data in a sane way. This statement will use the default constructor, which creates/initializes an empty `std::string`.

Way different from C behavior.

Ionic:
The performance difference doesn't really matter. Plain `char` arrays might be about 30% faster than using `std::string`, but that is only relevant if doing *a lot* of string operations all the time. Writing to a terminal or a file (which is what usually happens with these strings) is *orders of magnitudes* slower than any `std::string` or `char*` operation, so any performance improvement `char*` operations might have will be completely irrelevant in our use case.

I've actually tried to keep the `char*` return value in my first implementation, but came to the conclusion that this is insane.

Every instance of `strTimestamp()` or `strMsTimestamp()` would have been to be rewritten into something like this:

```C++
  [char *tss = NULL; /* First time in block */]
  tss = str(Ms)Timestamp(...);
  [...] << tss << [...];
  SAFE_FREE(tss);
```

There are hundred of such locations and it blows the code up for no good reason.

nxcomp/src/Timestamp.cpp
uli42: the err is basically the result of the active `#ifdef` block. If there was only one of the blocks, would you still use the err variable? Probably not. My suggestion will save two or three lines of code. (only the `if` line should be moved to the `#ifdef`, not the whole `if` block

nxcomp/src/Timestamp.cpp
uli42: ok

nxcomp/src/Timestamp.cpp
```diff
@@ -44,34 +44,57 @@
...
-              (float) ts.tv_usec / 1000);
+    snprintf(ctime_new, sizeof(ctime_new), "%.8s:%3.3f",
+             ctime_now.c_str() + 11, static_cast<float>(ts.tv_usec) / 1000);
 
-  strncpy(ctime_now, ctime_new, 24);
+    ret = ctime_new;
+  }
 
-  return ctime_now;
+  return ret;
```
uli42, line 97: One thing: isn't there a timestamp class in C++ that will generate what we need? We could drop the C functions like snprintf altogether.

nxcomp/src/Timestamp.cpp
Ionic: Okay, I finally understood what you mean. Yeah, I've been using `err` only to map the error code correctly since the two functions return different values (esp. in the error case).

nxcomp/src/Timestamp.cpp
```diff
@@ -44,34 +44,57 @@
...
+std::string strTimestamp(const T_timestamp &ts)
 {
-  char *ctime_now = ctime((time_t *) &ts.tv_sec);
+  std::string ret;
+
+  char ctime_now[26] = { };
+  bool err = true;
+
+#if HAVE_CTIME_S
+  errno_t retval = ::ctime_s(ctime_now, 26, static_cast<const time_t*>(&ts.tv_sec));
```
uli42: sizeof(ctime_now) instead of hardcoded value?

nxcomp/src/Timestamp.cpp
Ionic: Done.

nxcomp/src/Timestamp.cpp
Ionic: Yep, done.

nxcomp/src/Timestamp.cpp
uli42: ping?

Ionic: *sigh* No, and I probably have to explain this.

The C++ standard doesn't include support for calendric services. Support is limited to timers at best, but nothing that would need a calender. Implementing stuff like this is a snake pit and the standards committee didn't want to delve into this only to in the end find out that it's unusable anyway since tiny details have been missed. So to save a great amount of time and spare themselves headaches, they decided to drop that matter completely.

And so they did till (almost) this day.

For such things, the only usable alternative are C functions.

The `ctime` function family is deprecated in favor of `strftime` (and especially the `ctime()` function not even thread-safe), but while implementing the new logging code that used `strftime` in the beginning I noticed that `strftime` is not implemented in a thread-safe fashion in glibc (although that's no standards problem, since it COULD be thread-safe - glibc just adds stuff that breaks that).

The only viable alternative would be `std::put_time`, which should be analogous to `strftime`, but maybe better implemented. This function is only available since C++11 though, and we cannot universally depend upon that (i.e., there are still platforms we want to build on that don't support this standard).

For now it's probably best to stay with `ctime_r`/`ctime_s`.

uli42: ok, thx.

uli42 (approved):

Issue #622, opened 2017-12-29
Add continuous integration


mjtrangoni:
Hi,

I think this could add some value to the project, building **automatically** against some different compilers and different compiler versions too!  See for example, [Travis C++](https://docs.travis-ci.com/user/languages/cpp/).

If you agree, I would like to collaborate with this implementation.

sunweaver:
Hi Mario,

On  Fr 29 Dez 2017 11:46:29 CET, Mario Trangoni wrote:

> Hi,
>
> I think this could add some value to the project, building  
> **automatically** against some different compilers and different  
> compiler versions too!  See for example, [Travis  
> C++](https://docs.travis-ci.com/user/languages/cpp/).
>
> If you agree, I would like to collaborate with this implementation.

we already do some package builds on .deb based systems under  
jenkins.arctica-project.org.

However, this is far from a versatile Travis CI setup. So, if you  
volunteer in settings something up around this, please go ahead and  
file PRs, so that we integrate the travis configuration you come up  
with.

Thanks for joining in!!!
Mike

PS: maybe you want to share in what context you use nx-libs / nxagent. Thanks!
-- 

DAS-NETZWERKTEAM
mike gabriel, herweg 7, 24357 fleckeby
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

GnuPG Fingerprint: 9BFB AEE8 6C0A A5FF BF22  0782 9AF4 6B30 2577 1B31
mail: mike.gabriel@das-netzwerkteam.de, http://das-netzwerkteam.de

Ionic:
> we already do some package builds on .deb based systems under jenkins.arctica-project.org.

And, more recently, via the X2Go build infrastructure, that also includes RPM-based platforms.

mjtrangoni:
Hi @sunweaver @Ionic,

Well I use the `nx-libs` that @uli42 provides, and I told him I can collaborate with the project too. 

About the many build infrastructures, what do you think about adding or documenting them in a GitHub wiki-page [here](https://github.com/ArcticaProject/nx-libs/wiki)? By the way, I think the `README.md` file should be a bit shorter too, linking the extended documentation to the wiki.

As well as the building topic, there is an option called [Projects](https://github.com/ArcticaProject/nx-libs/projects) where you can link all the open and closed issues related or expected for a release.

Thank you!

mjtrangoni:
Travis is working now.

PR #624, opened 2018-01-01
Add TravisCI

mjtrangoni:
Hi @sunweaver,

As you told me in #622, I made an initial `TravisCI` implementation, running only on `Linux` (Ubuntu Trusty container images), but I think it is also possible building the project on `OS X`.

It runs *four* different builds in parallel with,
 
1. GCC 4.9, as an old GCC release.
1. GCC 7.X, as the latest GCC release.
1. CLANG 3.9, as an old CLANG release.
1. CLANG 5.0, as the latest CLANG release.

taking only 15 minutes for building the complete project.

I had to fix only two errors reported by Clang when compiling, but the build are all working in my fork now. You have to configure/activate Travis [here](https://travis-ci.org/) with the **ArcticaProject** account, and it should works here too.

The next step would be doing some testing from those builds, and fixing some of the many warnings GCC 7.2 and CLANG 5.0 are reporting now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,56 @@
+# Use new trusty images, should yield newer compilers and packages
+sudo: true
+dist: trusty
+language: cpp
+
+matrix:
+  include:
+    - os: linux
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-4.9
+      env:
+         - MATRIX_EVAL="CC=gcc-4.9 && CXX=g++-4.9"
+
+    - os: linux
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-7
+      env:
+        - MATRIX_EVAL="CC=gcc-7 && CXX=g++-7"
+
+    - os: linux
+      addons:
+        apt:
+          sources:
+            - llvm-toolchain-trusty-3.9
+          packages:
+            - clang-3.9
+      env:
+        - MATRIX_EVAL="CC=clang-3.9 && CXX=clang++-3.9"
+
+    - os: linux
+      addons:
+        apt:
+          sources:
+            - llvm-toolchain-trusty-5.0
+          packages:
+            - clang-5.0
+      env:
+        - MATRIX_EVAL="CC=clang-5.0 && CXX=clang++-5.0"
+
+before_install:
+  - sudo apt-get update -qq
+  # imake
+  - sudo apt-get install -y xutils-dev
+  # X11 libaries
+  - sudo apt-get install -y libxcomposite-dev libxfont-dev libxinerama-dev libxrandr-dev libxtst-dev x11proto-fonts-dev
+  - eval "${MATRIX_EVAL}"
+script:
+  - make
--- a/nxcomp/src/Loop.cpp
+++ b/nxcomp/src/Loop.cpp
@@ -3573,7 +3573,7 @@ int SetupAuthInstance()
 
   if (control -> ProxyMode == proxy_server)
   {
-    if (authCookie != NULL && *authCookie != '\0')
+    if (*authCookie != '\0')
     {
       if (useLaunchdSocket == 1)
       {
--- a/nxcompshad/src/X11.cpp
+++ b/nxcompshad/src/X11.cpp
@@ -363,7 +363,7 @@ void Poller::shmInit(void)
 
     shminfo_ -> shmaddr = (char *)shmat(shminfo_ -> shmid, 0, 0);
 
-    if (shminfo_ -> shmaddr < 0)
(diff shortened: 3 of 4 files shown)


nxcomp/src/Loop.cpp
```diff
@@ -3573,7 +3573,7 @@ int SetupAuthInstance()
 
   if (control -> ProxyMode == proxy_server)
   {
-    if (authCookie != NULL && *authCookie != '\0')
+    if (*authCookie != '\0')
```
uli42, line 3576: Alright, it is always true because it is defined a fixed char array. I am astonished that this is the only place where a NULL check is done for those arrays...

.travis.yml
```diff
@@ -0,0 +1,56 @@
...
+            - llvm-toolchain-trusty-5.0
+          packages:
+            - clang-5.0
+      env:
+        - MATRIX_EVAL="CC=clang-5.0 && CXX=clang++-5.0"
+
+before_install:
+  - sudo apt-get update -qq
+  # imake
+  - sudo apt-get install -y xutils-dev
```
Ionic, line 51: Why do we need this? These are build-time dependencies, so they should be installed automatically by the environment (i.e., `pbuilder` or `sbuild` or whatever Travis uses), right?

Besides, I'm not exactly convinced that our build environment handles `CC` and `CXX` env flags correctly (mostly because this was never tested so far IIRC). That's not a problem with your PR, but something that should be kept in mind and investigated in a build log to see if they actually take correctly.

nxcompshad/src/X11.cpp
```diff
@@ -363,7 +363,7 @@ void Poller::shmInit(void)
 
     shminfo_ -> shmaddr = (char *)shmat(shminfo_ -> shmid, 0, 0);
 
-    if (shminfo_ -> shmaddr < 0)
+    if (shminfo_ -> shmaddr < (void *)0)
```
Ionic, line 366: Why not just checking for `< NULL`?

.travis.yml
mjtrangoni: I wanted to keep it simple, but we could make this as complex as you want to. 

About the `CC` and `CXX`, I agree with you that this could be not working 100% OK, and should be fixed (IMHO) in another PR. The same rules with the many warnings that the different compilers are reporting.

nxcompshad/src/X11.cpp
mjtrangoni: I found [this](http://www.stroustrup.com/bs_faq2.html#null)

Ionic: Yeah... In C, `NULL` is typically defined as `(void*)0`, but this doesn't seem to be legit in C++ for some reason.

Thus using `(void*)0` here is absolutely fine, you're right.

mjtrangoni:
@uli42 I tried some "strict" flags, and it seems so far ok, because the four cases are not breaking at all.
Maybe you would like to take look at the logs [here](https://travis-ci.org/mjtrangoni/nx-libs/builds). 
If @sunweaver approve this, we have to remove the `CPPFLAGS` variable or maybe check if this is ok at this way.

sunweaver (approved):
Thanks to uli42 and Ionic for doing the review already.

@mjtrangoni : three steps now to get this merged...

  * rebase your PR branch against latest 3.6.x branch
  * would you be willling to set up the official TravisCI integration for this repo. I sense that you will much quicker with this, as you have already done it in other contexts. Thanks in advance!!!
  * Please feel welcome to look into the CC and CXX flags issue addressed in one of the comments.

Very good work. More of that!

sunweaver:
@mjtrangoni: ping

mjtrangoni:
@sunweaver, sorry. I am on vacation now, and coming back next week. I think I can rebase and check on Sunday.

sunweaver:
Superb! Thanks,
Mike

On Thursday, February 22, 2018, Mario Trangoni wrote:
> @sunweaver, sorry. I am on vacation now, and coming back next week. I think I can rebase and check on Sunday. 
> —
> You are receiving this because you were mentioned.
>

-- 
Sent from my Fairphone 2 (running Sailfish OS)

mjtrangoni:
@sunweaver I just rebased the PR against the **3.6.x** branch, and everything is green [here](https://travis-ci.org/mjtrangoni/nx-libs).

I would need the github rights for setting up Travis. It is as easy as login at travis-ci.org with your (owner) github credentials and enabling this repository there. The entire configuration comes with the **.travis.yml** file.

sunweaver:
Now that this PR is merged, let me know what you need to set up the actual CI stuff on Travis. Thanks!!!

mjtrangoni:
@sunweaver well, It is working now!! Look at the green badge in **README.md**. It will/should be running every time someone makes a PR or merge something.

Issue #589, opened 2017-12-05
nxagent drops X client connections unexpectedly


mbecroft:
Version: release 3.5.99.10 built from release tarball

Using nx in 'virtual desktop' mode.

From time to time, nxagent will unexpectedly close the connection to an X client. The result is that client's windows disappear, and the client emits an error such as: `Fatal IO error 11 (Resource temporarily unavailable) on X server :124.`

nxagent itself, the connection from the NX client to nxagent (if connected at the time of the fault), and all other clients continue working normally.

The client program that failed can be restarted by the user immediately, and continues working normally...until the next time the problem reoccurs.

The problem seems to happen mostly with 'heavy' clients that own a large number of resources as reported by `xrestop`.

Specifically, I experience the fault frequently with Firefox 38 ESR and Palemoon web browsers, and occasionally with the fvwm2 window manager. It has never happened with: emacs, xterm, xclock.

In the case of the web browser (Firefox or Palemoon) the browser will die seemingly at random...sometimes while I am not interacting with the system at all, other times when I am in the middle of browsing to a web site. There is no obvious pattern except that sometimes the browser will die (and I will restart it) every 10 minutes or so; other times it can run for several days...but always will eventually die with this fault.

Often I have a large number of tabs in the browser with correspondingly high resource utilisation. But the problem can also happen when I have just a few windows open.

No relevant messages appear on the browser stderr (except the error noted above) nor in the nxagent stderr found in (for instance) `/tmp/.x2go-mb/C-mb-124-1508677854_stD.xsession-x2go_dp24/session.log`.

In all other respects there are no problems. I am able to suspend and resume NX sessions as normal. The NX session never dies or closes unexpectedly.

In this particular case, the X clients experiencing intermittent disconnection are running on the same host as nxagent and connecting to nxagent via UNIX sockets. I have not tested whether the problem also occurs with clients connecting via TCP sockets.

I am challenged by how to better characterize or analyze the issue.

This issue corresponds with a discussion in `#x2go` around 20171205T0000 UTC in which I (`mbecroft`) explained the issue and `uli42` made some comments and suggested raising this ticket.

mbecroft:
*Further observation* - the above described issue gradually gets worse and worse the longer nxagent has been running over a period of weeks (even if individual X clients have been closed and restarted many times).

*Further issue - possibly (probably) related.*

[...]

PR #628, opened 2018-01-08
Re-enable XC-MISC extension

uli42:
Got dropped by accident in 9bc6ff269aa8bf4c41696ebf4a686c93729ba151

Fixes ArcticaProject/nx-libs#589


Ionic (commented):
Looks good, but take a look at `xcmisc.c` in the mentioned commit: was the `nx-X11*` include stuff also reverted by accident? I'd imagine we need to revert the first hunk back?

uli42:
@Ionic: You are right, I have changed the includes, too.

Ionic (approved):

Issue #634, opened 2018-01-11
-rrxinerama is ignored on reconnect for 3.5.99.12


uli42:
I had a session that was started with commandline options =+xinerama -rrxinerama=. On reconnect these were ignored and rrxinerama got activated with a rather strange size (1440x1200). The original session was fullscreen 3840x1200 while the client used for reconnect has 1920x1200.

options file on both sides did not contain any xinerama related stuff.

Output of the nxagent during reconnect:
```
NXAGENT(anubis:8094): Info: Screen [0] resized to geometry [3840x1200] fullscreen [0].
NXAGENT(anubis:8094): X Error: BadRROutput (invalid Output parameter) 162
NXAGENT(anubis:8094):   Extension:    146 (RANDR)
NXAGENT(anubis:8094):   Minor opcode: 9 (RRGetOutputInfo)
NXAGENT(anubis:8094):   Resource id:  0x100
NXAGENT(anubis:8094): X Error: BadRROutput (invalid Output parameter) 162
NXAGENT(anubis:8094):   Extension:    146 (RANDR)
NXAGENT(anubis:8094):   Minor opcode: 9 (RRGetOutputInfo)
NXAGENT(anubis:8094):   Resource id:  0x100
NXAGENT(anubis:8094): Info: Screen [0] resized to geometry [1440x1200] fullscreen [0].
```

uli42:
I could reproduce this with a newly spawned session. At the beginning rrxinerama was deactivated (xrandr did _not_ show a NX1 output but a list of resolutions). I sigHUPed the nxagent and reconnected. Bam - NX1 in xrandr output. The logfile of the nxagent contained this:

```
Session: Session resumed at 'Fri Jan 12 08:45:42 2018'.
Info: Screen [0] resized to geometry [3840x1200] fullscreen [1].
X Error: BadRROutput (invalid Output parameter) 162
  Extension:    146 (RANDR)
  Minor opcode: 9 (RRGetOutputInfo)
  Resource id:  0xff
X Error: BadRROutput (invalid Output parameter) 162
  Extension:    146 (RANDR)
  Minor opcode: 9 (RRGetOutputInfo)
  Resource id:  0xff
Info: Screen [0] resized to geometry [3840x1200] fullscreen [1].
```


nxagent was initially started (via xinit) like this:
```
[...]

PR #637, opened 2018-01-16
hw/nxagent/Screen.c: Additionally check for noRRXineramaExtension set…

sunweaver:
… to FALSE before using the nxagentAdjustRandRXinerama() function and providing a Xinerama-like user experience.

 Fixes ArcticaProject/nx-libs#634.

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Screen.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Screen.c
@@ -3797,7 +3797,7 @@ int nxagentChangeScreenConfig(int screen, int width, int height, int mmWidth, in
 
   if (r != 0)
   {
-    if (nxagentOption(Xinerama))
+    if (nxagentOption(Xinerama) && (noRRXineramaExtension == FALSE))
     {
       nxagentAdjustRandRXinerama(pScreen);
     }


uli42 (approved):
yes, looks good.

sunweaver:
```
14:49 <@sunweaver> uli42: we should at least check for both values.
14:49 <@sunweaver> if nxagentOption(Xinerama) == TRUE and noRRXinerama == FALSE
14:49 <@sunweaver> ?
14:50 <@sunweaver> because we don't want the nxagentAdjustRandRXinerama() function to be used when Xinerama is FALSE... RIght?
14:54 <@sunweaver> uli42: https://github.com/ArcticaProject/nx-libs/pull/637
14:55 -ArcticaGIT:#arctica- [nx-libs] sunweaver opened pull request #637: hw/nxagent/Screen.c: Additionally check for noRRXineramaExtension set… 
          (3.6.x...pr/keep-disabled-rrxinerama-disabled-on-reconnects) https://git.io/vNl64
15:02 < uli42> sunweaver: yes, we the check shoudl also look for the Xinerama option.
15:03 < uli42> sunweaver: approved
```

PR #648, opened 2018-01-23
Lift xkb to xorg-xserver-7.1/1.1.0 state

uli42:
In XORG-7_1 xkbconfig.c had been dropped. We must keep this file
because it is used by hw/nxagent/Keyboard.c to load the xkb
configuration file. Maybe we should move the relevant code to there or
drop xkb configuration file support from nxagent.


Ionic (approved):
Okay, + tiny split-up.

PR #650, opened 2018-01-23
Pr/update miext and mi

uli42:


nx-X11/programs/Xserver/mi/miinitext.c
```diff
@@ -228,6 +224,11 @@ extern void SecurityExtensionInit(void);
 extern void XFree86BigfontExtensionInit(void);
 #endif
 #ifdef GLXEXT
+/*
+typedef struct __GLXprovider __GLXprovider;
+extern __GLXprovider __glXMesaProvider;
+extern void GlxPushProvider(__GLXprovider *impl);
```
Ionic, line 230: Hm, the original code has these lines non-commented-out.

They seem to stem from the original AIGLX inclusion... I guess we won't actually need this in nxagent and it might not be actually working with the built-in software renderer, but we should probably still keep this in mind.

nx-X11/programs/Xserver/mi/miinitext.c
```diff
@@ -447,6 +445,9 @@ InitExtensions(argc, argv)
 #endif
 #endif
 #ifdef GLXEXT
+    /*
```
Ionic, line 448: Same here.

Ionic (approved):
Okay, with comments.

Ionic:
Actually, I noticed that I can also backport https://cgit.freedesktop.org/xorg/xserver/commit/?h=xorg-server-1_1_0&id=39ce5f1544029412f4060f3e89ce1d87222ef42b

Will do this in a separate commit after merging this PR.

PR #651, opened 2018-01-23
Pr/update os

uli42:
contains/depends on #642


Ionic (approved):
Okay, with tiny split.

Issue #657, opened 2018-02-09
Make MagicPixel behaviour optional


Hybrid512:
In fullscreen mode, there is a "MagicPixel" behaviour that minimize the session by clicking in the top right corner of the screen.

This behaviour should be made optional since, in TCE mode, it makes the desktop disappear and there is no way to get it back apart by rebooting the client and restore the suspended session.

This is troubling for users and absolutely not wanted.

It would be great to make this behaviour optional so that people wanting it could keep it and others like me can simply disabling it.

PR #658, opened 2018-02-21
nxagent: Add support for optionally enabling/disabling the magic pixel…

sunweaver:
… feature.

Fixes ArcticaProject/nx-libs#657.


Ionic (approved):
Typo in commit message, otherwise looks good.

Issue #653, opened 2018-01-25
Patch to add version info to nxagent -help


realsimix:
Attached small patch adds the version info to the output of nxagent -help:

```
$ nxagent -help 2>&1 | head -n 8
NXAGENT - Version 3.5.99.13

Usage: nxagent [<options>] [:<display>]
-a #                   mouse acceleration (pixels)
-ac                    disable access control restrictions
-audit int             set audit trail level
-auth file             select authorization file
bc                     enable bug compatibility
```
Would be nice if this could be added.
[nx-libs-3.6.x-20180124-nxagent_version.patch.txt](https://github.com/ArcticaProject/nx-libs/files/1663441/nx-libs-3.6.x-20180124-nxagent_version.patch.txt)

uli42:
Thank you! This also fixes #639.

However, why are you adding 
```
+#define NXAGENT_VERSION_STRING NX_VERSION_CURRENT_STRING
```
instead of simply using `NX_VERSION_CURRENT_STRING` directly?

Also, please have a look at `checkX2goAgent` in `hw/nxagent/Init.c`. The help should also consider this check. 

Please enclose your changes in `#ifdef NXAGENT_SERVER`. This simplyfies backports.

Finally while your are working at that it would be great if you could have a look at #606

realsimix:
Thanks for your input!

I tried to improve it as suggested. Can you have a look at it now?
[nx-libs-3.6.x-20180124-nxagent_version.patch.txt](https://github.com/ArcticaProject/nx-libs/files/1664483/nx-libs-3.6.x-20180124-nxagent_version.patch.txt)

realsimix:
While rethinking it, I'll create another version with nxagent -V to print version info.

realsimix:
Attached patch does it:
```
$ nxagent -help
Usage: nxagent [<options>] [:<display>]

-a #                   mouse acceleration (pixels)
-ac                    disable access control restrictions
-audit int             set audit trail level
-auth file             select authorization file
...
-S                     enable shadow mode
-B                     enable proxy binding mode
-V                     print version information
-options file          file containing nx/nx options
```
```
$ nxagent -V
NXAGENT - Version 3.5.99.13
```
Tries to be consistent with nxproxy which uses `nxproxy -v`, manpage is also updated.
[...]

PR #660, opened 2018-02-21
Add -version cmdline option

sunweaver:
This PR formalizes Simon's patch provided via #653.

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -336,6 +336,12 @@ int ddxProcessArgument(int argc, char *argv[], int i)
     return 0;
   }
 
+  if (!strcmp(argv[i], "-version"))
+  {
+    nxagentShowVersionInfo();
+    exit(0);
+  }
+
   /*
    * This had to be '-options' since the beginning
    * but was '-option' by mistake. Now we have to
@@ -2112,6 +2118,7 @@ void ddxUseMsg()
   ErrorF("-R                     enable rootless mode\n");
   ErrorF("-S                     enable shadow mode\n");
   ErrorF("-B                     enable proxy binding mode\n");
+  ErrorF("-version               show version information and exit\n");
 }
 
 static int nxagentGetDialogName()
@@ -2609,3 +2616,7 @@ void nxagentSetCoalescence()
   nxagentChangeOption(DisplayCoalescence, timeout);
 }
 
+void nxagentShowVersionInfo(void)
+{
+    ErrorF("NXAGENT - Version " NX_VERSION_CURRENT_STRING "\n");
+}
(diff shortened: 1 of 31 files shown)


nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -2591,3 +2598,7 @@ void nxagentSetCoalescence()
   nxagentChangeOption(DisplayCoalescence, timeout);
 }
 
+void nxagentShowVersionInfo()
```
Ionic: `nxagentShowVersionInfo(void)` please.

nx-X11/programs/Xserver/os/utils.c
```diff
@@ -556,7 +556,12 @@ AdjustWaitForDelay (void * waitTime, unsigned long newdelay)
 
 void UseMsg(void)
 {
+#ifdef NXAGENT_SERVER
+    extern const char *__progname;
+    ErrorF("Usage: %s [<options>] [:<display>]\n\n", __progname);
```
Ionic: We should avoid `__progname`, since it's not portable. Can you instead pass `argv[0]` around?

nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -2078,6 +2078,7 @@ void ddxUseMsg()
   ErrorF("-install               install colormaps directly\n");
 
   ErrorF("The NX system adds the following arguments:\n");
+  ErrorF("-options file          file containing nx/nx options\n");
```
Ionic: Can be either a file name or a plain string containing the options.

Ionic (changes requested):

nx-X11/programs/Xserver/hw/nxagent/Args.c
sunweaver: > nxagentShowVersionInfo(void) please.

Done. Also added another commit that fixes this issues for all functions in hw/nxagent/.

nx-X11/programs/Xserver/os/utils.c
sunweaver: > We should avoid __progname, since it's not portable. Can you instead pass argv[0] around?

I left the original commit as is and added another commit that drops __progname usage from the entire nxagent (we already used it elsewhere). The commit is also in this PR.

nx-X11/programs/Xserver/hw/nxagent/Args.c
sunweaver: > Can be either a file name or a plain string containing the options.

Right. Amended.

sunweaver:
@Ionic: next round for #660. Thanks.

Ionic (approved):

Issue #661, opened 2018-02-21
FTBFS on non-ELF platforms


Ionic:
Forcefully enabling new ELF dtags for RUNPATH works fine... on ELF-based systems.

It doesn't work so well on non-ELF-based systems, like Windows or OS X.

PR #663, opened 2018-02-25
Fix FTBFS on non-ELF platforms.

Ionic:
Check if target system is ELF-based and only add flag that enables new ELF dtags on ELF-based target systems.

Fixes: ArcticaProject/nx-libs#661


sunweaver (approved):
Nice! Thanks for working on this.

PR #672, opened 2018-03-05
Pr/saver unvalidated lengths proc screen saver unset attributes

sunweaver:
Two patches from X.org that are missing in Xserver/Xext/saver.c. One of them addresses a CVE.

### DIFF
--- a/nx-X11/programs/Xserver/Xext/saver.c
+++ b/nx-X11/programs/Xserver/Xext/saver.c
@@ -1342,6 +1342,8 @@ ProcScreenSaverUnsetAttributes (ClientPtr client)
        PanoramiXRes *draw;
        int i;
 
+       REQUEST_SIZE_MATCH(xScreenSaverUnsetAttributesReq);
+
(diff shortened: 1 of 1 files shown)


uli42:
looks good to me

nx-X11/programs/Xserver/Xext/saver.c
```diff
@@ -1515,7 +1517,6 @@ SProcScreenSaverSuspend (ClientPtr client)
 
     swaps(&stuff->length);
     REQUEST_SIZE_MATCH(xScreenSaverSuspendReq);
-    swapl(&stuff->suspend);
```
Ionic: NAK.

X.Org removed this, but as far as I can tell this is actually an oversight/a bug. I'll poke X.Org to find out why they thought that might be a good idea.

Ionic (changes requested):
Rejecting while investigation pending.

sunweaver:
@Ionic: I have moved the dodgy commit, now only the CVE fix is left. Please merge.

Ionic (approved):

PR #675, opened 2018-03-08
Pr/fix debug/test warnings

uli42:


sunweaver (approved):
lgtm

nx-X11/programs/Xserver/hw/nxagent/Events.c
```diff
@@ -835,7 +835,7 @@ void nxagentDispatchEvents(PredicateFuncPtr predicate)
 
   #ifdef TEST
   fprintf(stderr, "nxagentDispatchEvents: Going to handle new events with "
-              "predicate [%p].\n", predicate);
+              "predicate [%p].\n", *(void **)&predicate);
```
Ionic, line 824: I think we should change that to
```C
   #ifdef TEST
   {
       unsigned char *cp = (unsigned char*)&predicate;

       fprintf(stderr, "nxagentDispatchEvents: Going to handle new events with "
                   "predicate [0x");
       for
       #if (X_BYTE_ORDER == X_BIG_ENDIAN)
           (size_t i = 0; i < sizeof(predicate); ++i)
       #else
           (size_t i = sizeof(predicate); i > 0; --i)
       #endif
       {
           fprintf(stderr, "%0.2x", cp[i]);
       }
       fprintf(stderr, "].\n");
    }
    #endif
```
as per [this advice](https://stackoverflow.com/a/34251235).

Admittedly longer and uglier, but not relying on UB...

nx-X11/programs/Xserver/hw/nxagent/Events.c
```diff
@@ -2502,7 +2502,7 @@ FIXME: This can be maybe optimized by consuming the
                        nxagentExposeQueue.exposures[index].remoteRegion, &sum);
 
       #ifdef TEST
-      fprintf(stderr, "nxagentHandleExposeEvent: Added region for window [%ld] to position [%d].\n",
```
Ionic, line 2503: Please drop this whole commit.

I know we've been "fixing" these issues before, but that probably was a very bad idea to begin with. We're just shadowing *real* problems with that.

`nxagentWindow()` is a macro for getting a `Window` object, which is an `XID`, which can either be `CARD32` in X server or `unsigned long` (which does have a variable, machine-dependent size) in Xlib context. Provided that we pass that to an Xlib function later on, the code is wrong (or rather, the original code uses the "correct" size specifier in the `fprintf()` call, but falls short of handling everything else correctly...)

Let's not mask this.

Ionic (changes requested):

nx-X11/programs/Xserver/hw/nxagent/Events.c
uli42: over tiem I have seen more places like this. So maybe we should create a macro for that. It is way too ugly to have it in place..

nx-X11/programs/Xserver/hw/nxagent/Events.c
uli42: Agreed. So what's your suggestion to get rid of the compilation errors in that case?

nx-X11/programs/Xserver/hw/nxagent/Events.c
Ionic: Macros would make sense, yeah, if we have to do that in different places.

Probably something like
```C
#if (X_BYTE_ORDER == X_BIG_ENDIAN)
    #define PRINT_FPTR(fptr) \
        do {\
            unsigned char *cp = (unsigned char*)&fptr;\
            fprintf(stderr, "[0x");\
            for (size_t i = 0; i < sizeof(fptr); ++i)\
            {\
                fprintf(stderr, "%0.2x", cp[i]);\
            }\
            fprintf(stderr, "]");\
        while (0)
#else
    #define PRINT_FPTR(fptr) \
        do {\
            unsigned char *cp = (unsigned char*)&fptr;\
            fprintf(stderr, "[0x");\
            for (size_t i = sizeof(fptr); i > 0; --i)\
            {\
                fprintf(stderr, "%0.2x", cp[i]);\
            }\
            fprintf(stderr, "]");\
        while (0)
#endif
```

In that case we'd be able to replace the above snippet with just
```C
   #ifdef TEST
   {
       unsigned char *cp = (unsigned char*)&predicate;

       fprintf(stderr, "nxagentDispatchEvents: Going to handle new events with "
                   "predicate ");
       PRINT_FPTR(predicate);
       fprintf(stderr, ".\n");
    }
    #endif
```

nx-X11/programs/Xserver/hw/nxagent/Events.c
Ionic: Hmm, let me think about that and look through more code (including pre-processed one) and **do not** drop this commit.

It might be fine after all.

More comments later, I guess.

Ionic: Your modified hunk can stay.

`nxagentWindow(pWin)` is not used in other places than this message.

nx-X11/programs/Xserver/hw/nxagent/Events.c
```diff
@@ -2522,7 +2522,7 @@ FIXME: This can be maybe optimized by consuming the
               nxagentSplashCount == 1 && X -> xexpose.count == 0)
       {
         #ifdef DEBUG
-        fprintf(stderr, "nxagentHandleExposeEvent: Clearing root tile window id [%ld].\n",
```
Ionic, line 2523: This hunk can also stay.

`nxagentWindowPriv(nxagentRootTileWindow)->window` is passed to XClearWindow, which takes a `Window64` parameter (i.e., `unsigned long`) and implicit conversion rules make sure that the smaller value passed here is promoted to `unsigned long`, so no problem here.

nx-X11/programs/Xserver/hw/nxagent/Events.c
```diff
@@ -2761,7 +2761,7 @@ int nxagentHandleClientMessageEvent(XEvent *X, enum HandleEventResult *result)
       else
       {
         #ifdef TEST
-        fprintf(stderr, "Events: WM_DELETE_WINDOW arrived Atom = %ld.\n", wmAtom);
```
Ionic, line 2762: Seems to be okay as well.

Events implementation is scary and complicated - all Events are 32 bytes wide, but how the data is interpreted depends on the events type. `Xlib` and the server handle the heavy lifting of swapping values if necessary and laying data out correctly when transferring over the wire.

The crucial part is this one:
```C
 wmAtom = (Atom) X -> xclient.data.l[0];
```

Since `X->xclient.data.l[0]` is typed `long`, it will be accessed as such and we don't need to care about it. `Atom` will be a `CARD32` object, which means that a conversion needs to take place, but since `XID` in X server contexts are always `CARD32` objects and clients are confined to using a range predefined by the server only, the conversion should work fine.

`wmAtom` isn't actually used later on, which is also good.

nx-X11/programs/Xserver/hw/nxagent/Events.c
```diff
@@ -4107,7 +4107,7 @@ void nxagentSynchronizeExpose(void)
       {
         #ifdef TEST
         fprintf(stderr, "nxagentSynchronizeExpose: Going to call miWindowExposures"
-                    " for window [%ld] - rects [%ld].\n", nxagentWindow(pWin),
```
Ionic, line 4130: Also okay, only used in this message.

I guess we can keep this commit as-is. Retracting my previous note to delete it.

sunweaver:
@uli42: ping, @Ionic: ping. I see various message from @Ionic, ACK'ing the original changes. What's the status of this?

uli42:
Well, Ionic was not sure if we should include that or not. The current state is "yes".

sunweaver:
> Well, Ionic was not sure if we should include that or not. The current state is "yes".

@Ionic: can you then approve this PR and merge?

sunweaver:
@Ionic: ping.

PR #676, opened 2018-03-09
Use `libtirpc` on newer OpenSuSE and Fedora versions.

Ionic:
This PR adds code to utilize `libtirpc` on newer Fedora and OpenSuSE versions for the core X server.

Note that `libX11` could also use secure RPC - if that was turned on in the first place. We don't do this and do not plan on doing that in the future, since `libNX_X11` is only used for internal communication. It's successor `libxcb` doesn't seem to include `<rpc/rpc.h>` anyway (and so do recent `libX11` versions that are based on `libxcb`). Thus, our `libX11` copy was purposely not modified.


sunweaver:
Travis fails due main Makefile syntax flaw... Please fix.

Ionic:
Found out what caused the issue and added it on top of the queue as 7986f001c2d1ef821fefbb67bd0b180b95f7cfed.

While it was only affected older GNU Make versions, the behavior has been incorrect in any case.

nx-X11/programs/Xserver/Imakefile
```diff
@@ -296,11 +296,19 @@ ZLIB      =  $$(pkg-config --exists 'zlib' >/dev/null 2>&1 && pkg-config --libs
...
+TIRPC_LIB =  $$(pkg-config --libs 'libtirpc')
+#endif
+#endif
+
 NXAGENT = hw/nxagent/LibraryTargetName(nxagent)
 NXCOMPEXT =  hw/nxagent/compext/LibraryTargetName(compext)
 NXAGENTLIBS = PreFbLibs $(NXAGENT) FbPostFbLibs $(NXCOMPEXT) $(NXAGENT) $(MI)
 NXAGENTSYSLIBS = $(LDPRELIBS) $(XLIB) $(SYSLIBS) $(XPMLIB) $(XMLLIB)
-NXCOMPEXTSYSLIBS = $(PIXMANLIB) $(ZLIB) $(PNGLIB) $(JPEGLIB)
+NXCOMPEXTSYSLIBS = $(PIXMANLIB) $(ZLIB) $(PNGLIB) $(JPEGLIB) $(TIRPC_LIB)
```
sunweaver: The TIRPC_LIB is not required for nxcompext build. Thus, it should be moved up and added to NXAGENTSYSLIBS. IMHO.

nx-X11/programs/Xserver/os/Imakefile
```diff
@@ -75,10 +75,17 @@ XDMAUTHOBJS =
 XDMAUTHSCRS =
 #endif
 
+TIRPC_INCLUDES =
```
sunweaver, line 78: Move further down into a 
```
#if UseTIRPC
... 
#else
<here>
#endif
```

nx-libs.spec
```diff
@@ -410,7 +410,14 @@ chmod a+x my_configure;
...
 export LOCAL_LDFLAGS="%{__global_ldflags}"
 export CDEBUGFLAGS="%{?__global_cppflags} %{?__global_cflags} %{?optflags}"
-make %{?_smp_mflags} CONFIGURE="$PWD/my_configure" PREFIX=%{_prefix} LIBDIR=%{_libdir} CDEBUGFLAGS="${CDEBUGFLAGS}" LOCAL_LDFLAGS="${LOCAL_LDFLAGS}" SHLIBGLOBALSFLAGS="${SHLIBGLOBALSFLAGS}"
+IMAKE_DEFINES=''
+FORCE_TIRPC='NO'
+%if 0%{?fedora} > 27 || 0%{?suse_version} > 1500
+FORCE_TIRPC='YES'
+%endif
+IMAKE_DEFINES="-DUseTIRPC=${FORCE_TIRPC}"
+export IMAKE_DEFINES
```
sunweaver: Why export IMAKE_DEFINES???

Either export _or_ handing it over to the main Makefile as also done in the make call below. IMHO, I'd prefer the main Makefile to be aware of IMAKE_DEFINES (similar to the FONT_DEFINES variable).

-> So handing it over to main Makefile is the preference. An export should not be needed then.

sunweaver (changes requested):
@Ionic: thanks for working on this PR. I request some changes, see my comments.

nx-X11/programs/Xserver/Imakefile
Ionic: Neither is `pixman` as far as I can tell (though definitely a dependency of `nxagent`.) I'll try moving both and see what happens.

nx-X11/programs/Xserver/os/Imakefile
Ionic: The intent was to clear the variable in all cases, also when `HasSecureRPC` is false.

If moving that into these conditionals, I'd need to do it two times when an added `#else` branch since the rest of the code also sets or clears the vars based on conditionals. IMHO the current solution provides a smaller and more elegant diff.

nx-libs.spec
Ionic: I did so at first, but didn't like that too well since it added `IMAKE_DEFINES="$(IMAKE_DEFINES)"` all over the place in the top-level `Makefile`.

It may make sense to push `IMAKE_DEFINES` down, since with that a lot of options can be changed in the `nxagent` build process without using even more pass-down variables. I could potentially even rework the `XFONT` and `XEXT` stuff and drop that, instead relying on passing `imake` defines and having the `nxagent` imake system dealing with the details.

Does that make sense?

sunweaver:
On  Di 13 Mär 2018 10:15:51 CET, Mihai Moldovan wrote:

> Ionic commented on this pull request.
>
>
>
>> @@ -75,10 +75,17 @@ XDMAUTHOBJS =
>  XDMAUTHSCRS =
>  #endif
>
> +TIRPC_INCLUDES =
>
> The intent was to clear the variable in all cases, also when  
> `HasSecureRPC` is false.
>
> If moving that into these conditionals, I'd need to do it two times  
> when an added `#else` branch since the rest of the code also sets or  
> clears the vars based on conditionals. IMHO the current solution  
> provides a smaller and more elegant diff.

Ack. Leave it as is.
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

sunweaver:
On  Di 13 Mär 2018 11:38:13 CET, Mihai Moldovan wrote:

> Ionic commented on this pull request.
>> @@ -410,7 +410,14 @@ chmod a+x my_configure;
>  export SHLIBGLOBALSFLAGS="%{__global_ldflags}"
>  export LOCAL_LDFLAGS="%{__global_ldflags}"
>  export CDEBUGFLAGS="%{?__global_cppflags} %{?__global_cflags} %{?optflags}"
> -make %{?_smp_mflags} CONFIGURE="$PWD/my_configure"  
> PREFIX=%{_prefix} LIBDIR=%{_libdir} CDEBUGFLAGS="${CDEBUGFLAGS}"  
> LOCAL_LDFLAGS="${LOCAL_LDFLAGS}"  
> SHLIBGLOBALSFLAGS="${SHLIBGLOBALSFLAGS}"
> +IMAKE_DEFINES=''
> +FORCE_TIRPC='NO'
> +%if 0%{?fedora} > 27 || 0%{?suse_version} > 1500
> +FORCE_TIRPC='YES'
> +%endif
> +IMAKE_DEFINES="-DUseTIRPC=${FORCE_TIRPC}"
> +export IMAKE_DEFINES
>
> I did so at first, but didn't like that too well since it added  
> `IMAKE_DEFINES="$(IMAKE_DEFINES)"` all over the place in the  
> top-level `Makefile`.

Ok...

> It may make sense to push `IMAKE_DEFINES` down, since with that a  
> lot of options can be changed in the `nxagent` build process without  
> using even more pass-down variables.

While this is a good idea and a good point...

> I could potentially even rework the `XFONT` and `XEXT` stuff and  
> drop that, instead relying on passing `imake` defines and having the  
> `nxagent` imake system dealing with the details.

... we should really invest this energy into imake->autotools transition.

> Does that make sense?

Yes and no.

So, I'd say let's introduce IMAKE_DEFINES everywhere but not touch  
FONT_DEFINES and XEXT stuff for now.

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

Ionic:
Implemented requested changes and tested building on Fedora Rawhide and OpenSuSE Tumbleweed. Looks good, Travis seems to be happy, too.

sunweaver (approved):
Ok now. Thanks for working on this!

Issue #694, opened 2018-05-17
running nxagent without window manager breaks xinerama


uli42:
If you run nxagent agains an X server without window manager it ignores the xinerama config.

(this is currently being discussed on the x2go-dev mailing list)

```
Am 17.05.2018 um 14:33 schrieb Ulrich Sibiller:
> On Thu, May 17, 2018 at 2:13 PM, Oleksandr Shneyder
> <o.shneyder@phoca-gmbh.de> wrote:
>> Nope, it looks for me, that nxproxy window get information about screens
>> from WM. If WM is not started, nxproxy get just the window geometry.
>> With started WM we have correct display information on server side.
>> Xinerama starts to work with WM even if X2Go Session already running. I
>> made some experiments like:
>>
>> 1. Starting tce without WM.
>> 2. Connect to TC with ssh.
>> 3. Open X2Go Session. At this moment Xinerama in X2Go session not
>> working. I have one big screen over two monitors.
>
> What exactly do you mean by "open x2go session"?

I mean start X2Go session from TC

>
>> 4. Starting WM on the TC. At this moment Xinerama starts to work in the
>> running X2Go session immediately.
>>
>> It worked in TCE for me in 100% of cases. Independent when I'm starting
>> WM - before X2Go Client, after X2Go Client or even when X2Go Session
>> already running. I'll implement this solution in classic TCE and will
>> let my customers test it.



> I'd prefer understanding what's going on first. nxagent is running

You can easily reproduce this issue by running x2goclient without WM.
It's not really important for me now, why it's not working without WM. I
think NX using some WM-Based function to determine screen configuration
on the client side. I don't have resources to investigate this issue
right now and it's not critical for me.


>  if (nxagentGrabServerInfo.grabstate == SERVER_GRABBED &&
> nxagentGrabServerInfo.client != NULL)
....................
> And XineramaQueryScreens ins finally called in
> nxagentAdjustRandRXinerama. BUT: If nxagentAdjustRandRXinerama FAILS
> to get that data it will NOT fall back to the version you see but to
> reporting ONE screen (called NX1, IIRC). So all this fails somewhere
> else.

No, if I'm starting the session from TC without WM, I don't have screen
with name NX1. xrandr output looks in this case like this:

Screen 0: minimum 320 x 240, current 2560 x 1024, maximum 2560 x 1200
default connected 2560x1024+0+0 0mm x 0mm
   1360x768      60.00
   320x240       60.00
   1024x600      60.00
[...]

PR #695, opened 2018-05-17
Update Xinerama on XMapEvent

uli42:
Fixes ArcticaProject/nx-libs#694

Not sure about possible side effects...


sunweaver (approved):
I forced-push to this PR's update branch and removed mmwidth, mmheight from the nxagentChangeScreenConfig() calls to reflect recent API changes in Screen.c

But otherwise, ok. Works like a charme with Arctica Greeter's remote logon feature.

PR #705, opened 2018-06-09
fix window title encoding (use utf8) for nxagent

vpashka:


sunweaver:
vpashka: thanks for this PR. Can you please explain more about the issue you resolve with this? Is there a test case where we can see what your commit fixes?

sunweaver:
@vpashka: follow-up msg, as I did the highlighting wrong...

@vpashka: thanks for this PR. Can you please explain more about the issue you resolve with this? Is there a test case where we can see what your commit fixes?

vpashka:
This PR solves the problem of incorrect display of the window name in the national coding (for example for me it is ru_RU.UTF-8).

As a minimal example of showing problems, try such:

```
#include <X11/Xlib.h>
#include <X11/Xutil.h>
#include <stdio.h>
#include <stdlib.h> // prevents error for exit on line 18 when compiling with gcc
#include <X11/Xlocale.h>

// #define FIX

int main( int argc, char* argv[] ) 
{
    Display *d;
    int s;
    Window w;
    XEvent e;
    XTextProperty  title;  
    char* localized_ru_string = "Rus: русский текст";
    int ret = 0;

    /* open connection with the server */
    d=XOpenDisplay(NULL);
    if(d==NULL) {
        printf("Cannot open display\n");
        exit(1);
    }
    s=DefaultScreen(d);

    /* create window */
    w=XCreateSimpleWindow(d, RootWindow(d, s), 10, 10, 300, 100, 1,
                          BlackPixel(d, s), WhitePixel(d, s));

    // Process Window Close Event through event handler so XNextEvent does Not fail
    Atom delWindow = XInternAtom( d, "WM_DELETE_WINDOW", 0 );
    XSetWMProtocols(d , w, &delWindow, 1);

    /* select kind of events we are interested in */
    XSelectInput(d, w, ExposureMask | KeyPressMask);

    /* map (show) the window */
    XMapWindow(d, w);
    
#ifdef FIX    
    ret = Xutf8TextListToTextProperty(d, &localized_ru_string, 1, XCompoundTextStyle, &title);  
    if( ret == 0 )
      XSetWMName(d, w, &title); 
#else
    XStoreName(d, w, localized_ru_string);
#endif
    
    /* event loop */
    while(1) {
        
        XNextEvent(d, &e);
        
	/* draw or redraw the window */
        if(e.type==Expose) {
            XFillRectangle(d, w, DefaultGC(d, s), 20, 20, 10, 10);
        }

        if(e.type==KeyPress)
            break;

        if(e.type==ClientMessage)
            break;
    }

    XDestroyWindow(d, w);
    XCloseDisplay(d);

    return 0;
}
```

P.S. I apologize for my English, I'm not very good at it.

sunweaver:
Thanks for the test case code. We will check that in the next couple of days and report back to you.

And: don't worry about your English, it is fully fine. Plus: you C coding skills are even better!!! Welcome to the nx-libs hacking world.

vpashka:
Thank you so much. I will wait..

sunweaver:
On  Mo 11 Jun 2018 09:54:02 CEST, Pavel Vainerman wrote:

> Thank you so much. I will wait..

Please join us on IRC, if you like: #arctica on Freenode.

Thanks+Greets,
Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

sunweaver:
@vpashka: @Ionic: @uli42: I can reproduce the issue. @Ionic: did you find time for reviewing the patch?

nx-X11/programs/Xserver/hw/nxagent/Screen.c
```diff
@@ -1910,6 +1913,9 @@ N/A
                            nxagentIconPixmap,
                            argv, argc, sizeHints);
```
sunweaver: Above the below code block, I read:

```
    /* FIXME: deprecated, replaced by XSetWmProperties() */
    XSetStandardProperties(nxagentDisplay,
                           nxagentDefaultWindows[pScreen->myNum],
                           nxagentWindowName,
                           nxagentWindowName,
                           nxagentIconPixmap,
                           argv, argc, sizeHints);
```

Have you tried replacing XSetStandardProperties() call by X(utf8)SetWmProperties()? Mayb the below code calling XSetWMName() would be obsolete then.

Can you check that?

sunweaver (changes requested):
I wonder, if the call to XSetWMName() using xtpWmName could be avoided by replacing XSetStandardProperties() by X(utf8)SetWMProperties().

Furthermore, I am a bit confused about your code example. Obviously, you can fix the display from withing the X client test code.

If I get this right, you set a localized window title in your test code and expect nxagent to convert that localized string to UTF and display it correctly.

I am unsure, how this is generally handled in nxagent. If we consider a situation, where some localized session gets started in nxagent and this session is resumed from a system with a different locale. Then I'd expect nxagent to display things correctly on the different locale X2Go / NX client where it gets resumed.

So, in a generic world, the Xserver part in nxagent would be locale gnostic of the session running inside nxagent.

The bridge between Xserver part of nxagent and Xclient part of nxagent should then do some UTF-8 conversion, so whatever the locale of the session inside is and whatever the locale of the NX client is, the displayed fonts end up correctly (because they are UTF-8 converted internally in nxagent). This hints that your code is the right way to go.

And... are there more spots where locale / encoding conversion would be necessary?

@vphaska: Please give some feedback and share your view on this. Thanks!

sunweaver:
@vpashka : ^^^ (see above)

nx-X11/programs/Xserver/hw/nxagent/Screen.c
vpashka: I apologize for not having responded.
Now I'm a little busy. As soon as I'm free,
I will try to test this function.

P.S. It seems I tried to use this function,
but it did not work. Unfortunately, I can not remember.

vpashka: I checked the "Xutf8SetStandardProperties" function, it really works.
But with 'XSetStandardProperties ()' nxagent is terminated.

I updated the commit.

sunweaver (approved):

PR #334, opened 2017-02-16
Xserver/xkb: cleanup

sunweaver:
This PR requires PR #319 to be merged first.

This PR also is still WIP. I also need to attribute the original commit authors on this PR.

PR #706, opened 2018-06-18
Gh sunweaver/pr/xserver xkb cleanup

uli42:
I have rebased Sunweaver's branch from #334. Unfortunately now the problem with non-working clipboard in nested rootless nxagents still happens. Even worse: the patch that previously worked is not helping here! I guess that problem lies somewhere deeper.


Makefile
```diff
@@ -227,6 +227,8 @@ install-full:
 	$(INSTALL_FILE) etc/nxagent.keyboard $(DESTDIR)$(ETCDIR_NX)/
 
 	$(INSTALL_DIR) $(DESTDIR)$(PREFIX)/share/nx
+	$(INSTALL_FILE) nx-X11/lib/X11/XErrorDB $(DESTDIR)$(PREFIX)/share/nx/
+	$(INSTALL_FILE) nx-X11/lib/X11/Xcms.txt $(DESTDIR)$(PREFIX)/share/nx/
```
sunweaver: This seems to be a rebasing flaw. The rgb.txt file stuff has been dropped months if not years ago.

Drop this commit.

debian/control
```diff
@@ -39,6 +39,8 @@ Vcs-Browser: https://github.com/ArcticaProject/nx-libs/
 Package: nx-x11-common
 Architecture: all
 Multi-Arch: foreign
+Pre-Depends:
+ dpkg (>= 1.15.7.2),
```
sunweaver: Also unrelated change in debian/control. Drop this commit. Pre-Depends is not necessary here.

sunweaver (changes requested):
Two little todos. Please fix and rebase against latest 3.6.x branch.

sunweaver:
@uli42: I have rebased this branch and dropped the two bogus commits.

sunweaver (approved):

PR #709, opened 2018-06-19
Makefile: pass down Xfont2 define for Imake

### LINKED ISSUE #708: Makefile broken regarding Xfont2
running `make` in the main directory on a system that has libXfont2 works fine for the first run. But it creates several `Makefile`s that still refer to libXfont1:
```
FONT_DEFINES = 
    XFONTLIB = -lXfont
```

With those files in place subsequent builds in the same tree (without cleaning up everything - my usual workflow when debugging) fail with xfont related problems as some files are compiled with/for xfont2 and the changed ones are getting compiled with/for xfont1. This is very annoying and reduces my productivity a lot...

uli42:
Fixes #708


sunweaver (approved):
I guess, we could remove FONT_DEFINES with this PR from the main Makefile.

But as this whole imake stuff will go away $soonish, I'll simply merge this PR to ease the developers' days.

uli42:
Well, I have noticed that running make Makefile in a nx-X11 subdir breaks it again. We should really ensure that either 
a) the variable get's passed down by the Makefiles, too
or
b) a config file in the config dir get's adapted to  include HasXfont2

PR #710, opened 2018-06-19
Pr/fix mask

uli42:


uli42:
Looking at the code behind XkbGetKeyboard reveals the XkbGetKeyboard() patch is wrong (and the orginal code is correct).  So please ignore that one. The other patch is ok.

sunweaver:
> Looking at the code behind XkbGetKeyboard reveals the XkbGetKeyboard() patch is wrong (and the orginal code is correct). So please ignore that one. The other patch is ok.

@uli42: please fix this branch then (in case there are commits that we should ignore)

sunweaver:
Obsolete by 7bb238065d70f35413e7a32f242db7aa7fab9308.

Thus, closing.

PR #712, opened 2018-06-21
Pr/misc

uli42:
some small fixes/improvements,.


sunweaver (approved):
Ok, but NXdixfonts.c goes out of sync with dixfonts.c. To avoid that, I will commit the appropriate changes to dixfonts.c after merging.

PR #717, opened 2018-07-25
Pr/xkb 1.3.0.0

uli42:
this lifts xkb to a newer level. It also drops support for our own NX_XKB_(ALTERNATE_)BASE_DIRECTORY stuff and uses the xorg/xkb code that is doing the same in principle. We lose the ability to check multiple own xkb directories at runtime but I doubt we ever need this as it is really a compile time thing. Consequently the directory to use and the default keymapping are now coming from the makefile. <s>So we need a way to get the proper xkb default path to the makefile which I would like to leave to someone else who is more familiar with all the makefile magic in the tree...</s> Makefile magic is included.

This should be included before my (old) PR #492 that adds keyboard=clone as this new code base makes it much easier to implement that feature. So once this is merged I'll review that PR once again.


uli42:
@sunweaver You wanted to check this at your free Saturday evening. Any progress?

nx-X11/programs/Xserver/xkb/Imakefile
```diff
@@ -40,7 +40,7 @@ XKB_DDXDEFS = XkbServerDefines
    LINTLIBS = ../dix/llib-ldix.ln ../os/llib-los.ln
 
 DEFINES = $(XKB_DDXDEFS) $(NX_DEFINES)
-XKB_DEFINES = -DXKB_BASE_DIRECTORY=\"$(LIBDIR)/xkb\" $(XKB_DISABLE)
+XKB_DEFINES = -DXKB_BASE_DIRECTORY=\"/usr/share/X11/xkb\" -DXKB_BIN_DIRECTORY=\"/usr/bin\" -DXKB_DFLT_RULES_FILE=\"base\" -DXKB_DFLT_KB_LAYOUT=\"us\" -DXKB_DFLT_KB_MODEL=\"pc102\" $(XKB_DISABLE)
```
sunweaver: urgh... this looks really hard-coded and hidden... Why not set all those in the main Makefile (I saw some changes like that on your big-1.4.2 branch, so maybe this got changed again. If so, please drop this PR / rebase it with the follow up.

nx-X11/programs/Xserver/xkb/Imakefile
```diff
@@ -40,7 +40,7 @@ XKB_DDXDEFS = XkbServerDefines
    LINTLIBS = ../dix/llib-ldix.ln ../os/llib-los.ln
 
 DEFINES = $(XKB_DDXDEFS) $(NX_DEFINES)
-XKB_DEFINES = -DXKB_BASE_DIRECTORY=\"/usr/share/X11/xkb\" -DXKB_BIN_DIRECTORY=\"/usr/bin\" -DXKB_DFLT_RULES_FILE=\"base\" -DXKB_DFLT_KB_LAYOUT=\"us\" -DXKB_DFLT_KB_MODEL=\"pc102\" $(XKB_DISABLE)
+XKB_DEFINES = -DXKB_BASE_DIRECTORY=\"$(SYSTEMXKBCONFDIR)\" -DXKB_BIN_DIRECTORY=\"$(SYSTEMXKBBINDIR)\" -DXKB_DFLT_RULES_FILE=\"base\" -DXKB_DFLT_KB_LAYOUT=\"us\" -DXKB_DFLT_KB_MODEL=\"pc102\" $(XKB_DISABLE)
```
sunweaver, line 43: ah, here we go... are the XKB RLMVO defaults sooo default that we should hard code them here? (Repeating myself...)

sunweaver (changes requested):
Hi,

I asked a question about hard-code RLVM in some imake config file.

Further more, commit msg of 25ca9b0653c4c3d30c6c29e2ab447d4a6003d26f contains some typos: "... where no longer ..."-> "... were no longer...", "adaptiuon" -> "adaptation"

Except from that, the PR looks good. Please fix the issues / comment on my remarks and then update this PR branch.

nx-X11/programs/Xserver/xkb/Imakefile
uli42: well, we need some default, us and pc102  are the values nx has always been using as defaults, base is the best choice nowadays (and is identical to xfree86 that nx was shipped with). We can of course change them to something else (to what?), but I don't see much benefit, since x2go (and with bad luck also the desktop) will set it anyway. IMHO there cannot be an automatism at the compile time, this is client side runtime decision. As I stated in the pr message I want to rebase my long-time pending keyboard overhaul once this pr is merged. Then I can easily add the runtime RMLVO auto-detection.

sunweaver: ok, then please fix commit msg of 25ca9b0 and we are good to go.

uli42: done

sunweaver (approved):
LGTM now.

PR #718, opened 2018-07-25
add list with versions of the xserver modules

uli42:
must be updated whenever a modules is lifted to a new upstream code base.


sunweaver (approved):
Nice!

PR #725, opened 2018-08-23
Pr/decode options

uli42:
decode URL encoded characters in options string. This is a much lighter (and cleaner) solution for the problem that #551 tries to solve.

(I did not succeed in linking the URLdecode stuff from nxcomp so I cloned it. If I get that right nxagent is not using anything from nxcomp directly currently.)

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -1037,14 +1037,14 @@ int ddxProcessArgument(int argc, char *argv[], int i)
   return 0;
 }
 
-static void nxagentParseOptions(char *name, char *value)
+static void nxagentParseSingleOption(char *name, char *value)
 {
   int size, argc;
 
   char *argv[2] = { NULL, NULL };
 
   #ifdef TEST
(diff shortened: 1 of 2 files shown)


nx-X11/programs/Xserver/hw/nxagent/Args.c
```diff
@@ -1037,6 +1037,39 @@ int ddxProcessArgument(int argc, char *argv[], int i)
...
+        *(to++) = hexval(str[1]) * 16 + hexval(str[2]);
+        str += 3;
+      }
+      else
+        *(to++) = *(str++);
+    }
+    *to = '\0';
+  }
+}
+
```
sunweaver, line 1072: Why copy of code here? We should rather _use_ the version found in nxcomp, shouldn't we?

sunweaver (commented):
This is more a comment / question than a request for changes: Why don't you simply use the coder version of URLDecodeInPlace from nxcomp straight away?

We need to make URLDecodeInPlace a public symbol and then simply include NX.h in Args.c.

uli42:
I wrote that in the PR comment: I could not manage to make it work.
Besides: the nxagent does not seem to be using any nxcomp function.

sunweaver (approved):
Ok, then...

Issue #301, opened 2016-12-13
atom.io editor fails to start possibly due to issue with BIG-REQUESTS extension


vatral:
When starting it under NX, this happens:

```
$ atom -f
[4380:1213/145919:ERROR:browser_main_loop.cc(231)] Running without the SUID sandbox! See https://chromium.googlesource.com/chromium/src/+/master/docs/linux_suid_sandbox_development.md for more information on developing with the sandbox on.
[4380:1213/145921:ERROR:browser_main_parts.cc(126)] X IO error received (X server probably went away)

```

The problem is described and discussed here: https://github.com/atom/atom/issues/4360 and they suggest it's related to the BIG-REQUESTS extension. I have tried the `x11trace -e` workaround and it fixed it.

I have reproduced it on the 3.6.x branch.

Issue #631, opened 2018-01-09
Visual Studio Code not working


uli42:
I was just trying to run the current version of Visual Studio Code (https://code.visualstudio.com/; I chose the 64bit tarball) on Centos7 with today's HEAD.

It does not open up but hangs the whole session instead. 

The last lines from xtrace:
```
...
001:<:009d:  4: Request(43): GetInputFocus 
001:>:009d:32: Reply to GetInputFocus: revert-to=PointerRoot(0x01) focus=0x03400026
001:<: Warning: buffer filled!
001:<:009e:8388652: Request(18): ChangeProperty mode=Replace(0x00) window=0x04a00001 property=0x12a("_NET_WM_ICON") type=0x6("CARDINAL") data=0x00000400,0x00000400,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000
000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x
00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x000000
...
0000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000;
[...]

PR #729, opened 2018-10-06
nxagent: Fix an excessive request size limitation that broke big-requ…

uli42:
…ests.

Fixes ArcticaProject/nx-libs#301, #631

Backport from xorg-xserver:

  commit ca82d4bddf235c9b68d51d68636bab40eafb9889
  Author: Eric Anholt <eric@anholt.net>
  Date:   Fri Aug 31 13:00:23 2007 -0700

    Bug #7186: Fix an excessive request size limitation that broke big-requests.

    MAXBUFSIZE appears to be a leftover of some previous time.  Instead, just
    use maxBigRequestSize when bigreqs are available (limiting buffers to ~16MB).
    When bigreqs are not available, needed won't be larger than the maximum
    size of a non-bigreqs request (256kB).

### DIFF
--- a/nx-X11/programs/Xserver/os/io.c
+++ b/nx-X11/programs/Xserver/os/io.c
@@ -301,7 +301,8 @@ ReadRequestFromClient(ClientPtr client)
 	 */
 
 	oci->lenLastReq = 0;
-	if (needed > MAXBUFSIZE)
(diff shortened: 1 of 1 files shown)


sunweaver (approved):
Nice!

PR #731, opened 2018-10-21
Pr/clipboard debug

uli42:
lots of improvements to clipboard, mostly debugging.

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
@@ -1092,8 +1092,7 @@ void nxagentNotifySelection(XEvent *X)
         fprintf(stderr, "%s: GetWindowProperty() returned [%s]\n", __func__, GetXErrorString(result));
         #endif
         if (result == BadAlloc || result == BadAtom ||
-                result == BadMatch || result == BadValue ||
-                    result == BadWindow)
+                result == BadWindow || result == BadValue)
         {
           fprintf (stderr, "Client GetProperty failed. Error = %s", GetXErrorString(result));
           lastServerProperty = None;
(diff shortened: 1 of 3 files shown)


nx-X11/programs/Xserver/hw/nxagent/NXproperty.c
```diff
@@ -93,9 +93,9 @@ nxagentWMStateRec;
 #undef  TEST
 #undef  DEBUG
 
-int 
+int
 ProcChangeProperty(ClientPtr client)
-{	      
+{
```
sunweaver: Why would we want to diverge hw/nxagent/NXproperty.c and dix/proporty.c white-space wise. Is this a good idea?

nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
```diff
@@ -163,6 +163,8 @@ const char * GetClientSelectionStageString(int stage)
 #define PrintClientSelectionStage()
 #endif
 
+#define SAFE_XFree(what) do {if (what) {XFree(what); what = NULL;}} while (0)
+
```
sunweaver: Shouldn't we define this much more globally, so that it can be reused at various other places?

sunweaver: @Ionic uses this macro in his Xinerama tiling branch, too.

sunweaver (changes requested):
Please check my comments. Currently, Github has out-of-sync webmirrors, so commenting on code ttw is really difficult and tricky.

About the FIXME regarding the reset of "i = 0" in one of the commits (Github would not let me comment inline): Not necessarily do we need to reset the "i" counter, currently it seems that the counter "i" is used to iterate over all selections, first up to nxagentCurrentSelection, then up to nxagentMaxSelections. This could be wanted. However, I am unsure and would have to investigate the clipboard code in more depths.

uli42:
I have moved the macro to Utils.h, but please note that it is  a
different one from Ionic's (which calls free()  while mine calls
XFree())

You are also right about the whitespace stuff, I did not have in mind
that NXproperty.c is derived from Xorg code in dix. So I reverted that
one.

PR #743, opened 2018-12-04
Pr/improve keyboard2

uli42:
merge after #742 

This adds:
- keyboard=clone
- keyboard=rmlvo/...
- global variables for remote keyboard

PR #744, opened 2018-12-07
Pr/x2go bypass

uli42:
depends on #743 

X2go keyboard bypass

If keyboard=auto in x2go, x2go expects <s>nx</s>x2goagent to create a keyboard file. x2go will later read that and pass it to setxkbmap. With #743 and this patch nxagent will do all this itself and instruct x2go to skip the keyboard configuration


sunweaver (approved):
:+1:

Issue #449, opened 2017-04-27
Fullscreen on startup buggy


uli42:
Activating fullscreen on startup can be done in two ways:

- `fullscreen=1` in options string
- `-geometry fullscreen`

On startup nxagent first opens the (unused) nxagentConfineWindow (in Display.c) at pos 0,0 with size 1x1. It then determines the size of the session window:
- if `-geometry` is provided its values are taken
- if `-geometry fullscreen` is provided, the full size if the display is taken
- if neither is provided 3/4 of the display's size are used

If `-geometry` is not provided at all but the options string contains `fullscreen=1` the fullscreen mode is also activated but the the values for size and pos of nxagent's window are not set properly because the options strings seems to be checked later on.

It turns out that this is a problem for x2goclient because it gets the wrong size of the nxagent window (called proxy window there) in thinclient environment. Current x2goserver (4.1.0.0) is not providing `-geometry fullscreen`.

We should check if we can make the code aware of fullscreen mode if the options strings contains it so that both ways lead to the same result on startup already

PR #761, opened 2019-01-13
Pr/window fixes

### LINKED ISSUE #757: Tooltips, pull-down menus, and context menus too far away (X2Go #1359)
https://bugs.x2go.org/cgi-bin/bugreport.cgi?bug=1359

"I've recently started experiencing a problem when running single
applications over X2Go.  Quite often small windows such as tooltips,
pull-down menus, and context menus get positioned far, far away from
where they are supposed to be.  For example, I'm currently writing this
message in Claws Mail; when I activate the "Edit" pull-down menu, the
menu doesn't appear under the menu bar but rather way over on my second
screen.  (I use a dual-monitor setup.)

The problem happens with a lot of different applications, but not
always consistently.  For instance, sometimes I might hover my pointer
over a widget that causes a tooltip to display halfway across the
screen, and a minute or two later I'll do the same thing but this time
the tooltip will display right over the widget like it's supposed to.

I'm not sure if this is a bug in X2Go or in my window manager, and if
it is in X2Go, whether its the client or server that's at fault.  I'm
running the X2Go Server 4.1.0.4 on openSUSE Tumbleweed, and connecting
from X2Go Client 4.1.1.1 on openSUSE Leap 15.0, using KDE Plasma
5.12.6.  (I can't upgrade the client due to Bug #1358.)"

uli42:
some fixes regarding window size calculation. Fixes #757 nad #449

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Display.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Display.c
@@ -1356,7 +1356,7 @@ FIXME: Use of nxagentParentWindow is strongly deprecated.
   {
     if (!(nxagentUserGeometry.flag & WidthValue))
     {
-      if (nxagentOption(Fullscreen))
+      if (nxagentOption(Fullscreen) || nxagentOption(Rootless))
       {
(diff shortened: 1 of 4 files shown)


sunweaver (approved):
NICE! Merging...

Issue #737, opened 2018-11-26
CUPS is broken on OSX starting from El Capitan


vatral:
Starting from El Capitan, the CUPS service is not running by default on OSX, and is started as needed. Also, it automatically shuts down after being unused for some time.

As a result, forwarding the print port fails to work correctly.

We're currently looking into this. I will have to look more deeply into how is an application supposed to do this properly, but my current idea is to make NX start cupsd by invoking cupsctl before connecting to the port.

I'm opening the bug in case you're aware of the issue, or have an idea of how it should be fixed.

uli42:
This is not an nx-libs bug. I'd rather not integrate cups start into nx-libs. This is clearly the task of the session management software (x2go, qvd etc.).

vatral:
Okay, so here's what I gathered so far:

CUPS is not running by default anymore on OSX, so it's not listening to the port at all (and there's no socket listener that starts it, it seems. At least on TCP). It also shuts down automatically after a while.

So here is the rough idea I had in mind: I want to add an option to NX to add a configurable command that would be executed on an incoming connection. This then would be provided by QVD. The command in the case of the print port would do whatever is needed to start CUPS.

How does that sound to you? I'd like to coordinate our efforts here, and that sounds like potentially useful functionality.

uli42:
On Mon, Dec 17, 2018 at 11:43 PM Vadim Troshchinskiy
<notifications@github.com> wrote:
>
> Okay, so here's what I gathered so far:
>
> CUPS is not running by default anymore on OSX, so it's not listening to the port at all (and there's no socket listener that starts it, it seems. At least on TCP). It also shuts down automatically after a while.
>
> So here is the rough idea I had in mind: I want to add an option to NX to add a configurable command that would be executed on an incoming connection. This then would be provided by QVD. The command in the case of the print port would do whatever is needed to start CUPS.
>
> How does that sound to you? I'd like to coordinate our efforts here, and that sounds like potentially useful functionality.

I am totally fine with that. Maybe we should also add the opposite,
too: a configurable command that is executed on disconnect.

But I am wondering (I don't know OSX): How is CUPS used on that system
at all? Who is starting the service when a user wants to print?

Uli

Ionic:
[...]

PR #764, opened 2019-01-22
Make it possible to forward client ports to unix sockets

vatral:
This is to address bug #737 

Modern versions of OSX don't have CUPS running all the time, and start it when /private/var/run/cupsd is opened.

The code already supports working with unix sockets, but specifying "cups=unix:/private/var/run/cupsd" doesn't work due to issues in the current commandline parser.

This should fix that. Please check the changes to ParseEnvironmentOptions, as I'm not 100% sure if there might be some edge case I'm not seeing.

### DIFF
--- a/nxcomp/src/ChannelEndPoint.cpp
+++ b/nxcomp/src/ChannelEndPoint.cpp
@@ -209,8 +209,6 @@ ChannelEndPoint::getUnixPath(char **unixPath) const {
 
   if (unixPath)
     *unixPath = NULL;
-  else
-    return false;
 
   long p;
   char *path = NULL;
@@ -230,7 +228,9 @@ ChannelEndPoint::getUnixPath(char **unixPath) const {
       return false;
   }
 
-  *unixPath = strdup(path);
+  // Only return value wanted
+  if ( unixPath )
+    *unixPath = strdup(path);
 
   return true;
 }
--- a/nxcomp/src/Loop.cpp
+++ b/nxcomp/src/Loop.cpp
@@ -7803,6 +7803,7 @@ int ParseEnvironmentOptions(const char *env, int force)
   strcpy(opts, env);
 
   char *nextOpts = opts;
+  bool nxdisplay_found = false;
 
   //
   // Ensure that DISPLAY environment variable
@@ -7824,14 +7825,17 @@ int ParseEnvironmentOptions(const char *env, int force)
   else if (strncasecmp(opts, "nx/nx,", 6) == 0)
   {
     nextOpts += 6;
+    nxdisplay_found = true;
   }
   else if (strncasecmp(opts, "nx,", 3) == 0)
   {
     nextOpts += 3;
+    nxdisplay_found = true;
   }
   else if (strncasecmp(opts, "nx:", 3) == 0)
   {
     nextOpts += 3;
+    nxdisplay_found = true;
   }
   else if (force == 0)
   {
@@ -7860,7 +7864,7 @@ int ParseEnvironmentOptions(const char *env, int force)
 
   value = strrchr(nextOpts, ':');
 
-  if (value != NULL)
+  if (value != NULL && nxdisplay_found )
   {
     char *check = value + 1;
 


sunweaver:
@vtq: thanks for this PR. I'll take a look.

sunweaver:
@vtq: Can you provide me with a simple nxagent / nxproxy startup command line pattern that gives me a test scenario for this?

vatral:
@sunweaver 

OSX:

    nxproxy -S cups=unix:/private/var/run/cupsd localhost:40

Linux:

    nxproxy -S cups=unix:/var/run/cups/cups.sock localhost:40

nxcomp/src/ChannelEndPoint.cpp
```diff
@@ -209,8 +209,6 @@ ChannelEndPoint::getUnixPath(char **unixPath) const {
 
   if (unixPath)
     *unixPath = NULL;
-  else
-    return false;
```
sunweaver, line 213: Bail out, if unixPath is unset. Ok.

nxcomp/src/ChannelEndPoint.cpp
```diff
@@ -230,7 +228,9 @@ ChannelEndPoint::getUnixPath(char **unixPath) const {
       return false;
   }
 
-  *unixPath = strdup(path);
+  // Only return value wanted
+  if ( unixPath )
+    *unixPath = strdup(path);
```
sunweaver, line 233: unixPath will never be NULL, so this if-clause is not required. (Remember, you added a return false further above).

nxcomp/src/Loop.cpp
```diff
@@ -7860,7 +7864,7 @@ int ParseEnvironmentOptions(const char *env, int force)
 
   value = strrchr(nextOpts, ':');
 
-  if (value != NULL)
+  if (value != NULL && nxdisplay_found )
```
sunweaver, line 7867: IMHO, nxdisplay_found the way you introduced it will always be TRUE down here. I think this whole commit is superfluous.

sunweaver (changes requested):
Please go through my inline comments. At the end of the day, the only change you need is the ``return false`` at the top of getUnixPath(). Please re-visit and give feedback.

nxcomp/src/ChannelEndPoint.cpp
vatral: I didn't add a "return false", I removed it. The fix is that getUnixPath has to return the right boolean even when passed a NULL.

The call tree starts here:

    ChannelEndPoint::validateSpec() {
       isTCP_ = getTCPHostAndPort();
       isUnix_ = getUnixPath();


getUnixPath is declared:   

    bool getUnixPath(char **path = NULL) const;

So, unixPath is a pointer to a pointer, which will be NULL when called from validateSpec(). validateSpec() doesn't actually want the path, it only wants the true/false return value from the function.

What was broken is that the function always returned false when not given a buffer, even if it actually was an unix path.

nxcomp/src/Loop.cpp
vatral: Not always, no. Look at the branch with the 'force' argument.

Here's a demonstration:
```

 nxproxy-gdb  -S cups=unix://private/vda localhost:40
GNU gdb (GDB) Fedora 8.1.1-3.fc28
...
Reading symbols from /home/vadim/git/nx-libs/nxproxy/src/nxproxy...done.
(gdb) break Loop.cpp:7867
No source file named Loop.cpp.
Make breakpoint pending on future shared library load? (y or [n]) y
Breakpoint 1 (Loop.cpp:7867) pending.
(gdb) run
Starting program: /home/vadim/git/nx-libs/nxproxy/src/nxproxy -S cups=unix://private/vda localhost:40

[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".

Breakpoint 1, ParseEnvironmentOptions (env=0x7fffffffe4bb "cups=unix://private/vda", force=1) at Loop.cpp:7867

7867      if (value != NULL && nxdisplay_found )
(gdb) print nxdisplay_found
$1 = false
(gdb) print value
$2 = 0x7fffffffc8d9 "://private/vda"
(gdb) bt
#0  ParseEnvironmentOptions (env=0x7fffffffe4bb "cups=unix://private/vda", force=1) at Loop.cpp:7867
#1  0x00007ffff7b0fceb in ParseCommandLineOptions (argc=4, argv=0x7fffffffe068) at Loop.cpp:8842
#2  0x00007ffff7af21cd in NXTransParseCommandLine (argc=4, argv=0x7fffffffe068) at Loop.cpp:1287
#3  0x0000000000400a06 in main (argc=4, argv=0x7fffffffe068) at Main.c:93
```

vatral: The point of that change is that "cups=unix://private/vda" contains a ":" in it, and that this causes execution to hit https://github.com/ArcticaProject/nx-libs/blob/889de46f759913742478d2cfe20f1f4a6c2effdc/nxcomp/src/Loop.cpp#L7873

If that branch is avoided, then option parsing works just fine.

That said, this is the bit I'm not quite sure about. The code here seems to do two different things with the same bit of code, and allowing a : in the value of "cups" breaks stuff.

This should probably be rewritten though, as this means there are two valid options at this point:

    nx/nx,foo=1,bar=2,etc=3:100

And:

    cups=unix://private/blah.


The problem is that the first option only works with a single ":". Do this and things break:

    nx/nx,foo=1,bar=2,cups=unix:/foo/bar,etc=3:100

nxcomp/src/ChannelEndPoint.cpp
sunweaver: > I didn't add a "return false", I removed it. The fix is that getUnixPath has to return the right boolean even when passed a NULL.

Oh well, do I need to think about red/green visual issues?

> The call tree starts here:
> 
> ```
> ChannelEndPoint::validateSpec() {
>    isTCP_ = getTCPHostAndPort();
>    isUnix_ = getUnixPath();
> ```
> 
> getUnixPath is declared:
> 
> ```
> bool getUnixPath(char **path = NULL) const;
> ```
> 
> So, unixPath is a pointer to a pointer, which will be NULL when called from validateSpec(). validateSpec() doesn't actually want the path, it only wants the true/false return value from the function.
> 
> What was broken is that the function always returned false when not given a buffer, even if it actually was an unix path.

Ok, thanks for this explanation. Sounds valid.

sunweaver (approved):
Ok, then, let's merge.

uli42:
see #775 for a problem that was introduced by this commit. We should re-open this issue (I can't seem to find how this can be done)

sunweaver:
@uli42: PRs can't be reopened. No need to reopen, your bug report is sufficient (#775).

PR #767, opened 2019-02-09
add nxdialog.py for displaying dialogs

uli42:
By adding NX_CLIENT=<path to nxdialog.py> to the envrionment
certain actions in nx will open dialogs. If the options string contains
"menu=1" rootless windows will show a popup menu when the mouse hits
the center of the title bar offering session disconnect/terminatation.

@sunweaver: can you please take care of the following:
- check if licensing is ok. Do we need to add our license header?
- check integration in to debian packaging stuff including dependencies (compilation's not required)

@Ionic: can you please check the dependencies/packaging for rpm based distros

### DIFF
--- a/nxdialog/nxdialog
+++ b/nxdialog/nxdialog
@@ -44,6 +44,9 @@
 
 # If an "NX_CLIENT" environment variable is not provided to nxagent
 # nxcomp library assumes this script is located in /usr/NX/bin/nxclient
+#
+# Example:
+# nxdialog --dialog yesno --message "message text" --caption "message title" --parent 0
 
 from __future__ import print_function
 
--- a/nxdialog/nxdialog
+++ b/nxdialog/nxdialog
@@ -30,6 +30,7 @@
 # - added missing docstrings
 # - pylint improvements
 # - removed neatx entry from the pulldoww menu
+# - use PyGObject instead of PyGtk and thus Gtk3
 
 """nxdialog program for handling dialog display."""
 
@@ -41,10 +42,10 @@ import os
 import signal
 import sys
 
-import gtk
-
-import pygtk
-pygtk.require("2.0")
+import gi
+gi.require_version('Gtk', '3.0')
+# pylint: disable=wrong-import-position
+from gi.repository import Gtk, Gdk, GdkX11
 
 PROGRAM = "nxdialog"
 
@@ -95,58 +96,62 @@ class PullDownMenu(object):
     def show(self):
         """ Shows popup and returns result. """
 
-        win = gtk.gdk.window_foreign_new(self.window_id)
+        display = Gdk.Display.get_default()
+        win = GdkX11.X11Window.foreign_new_for_display(display, self.window_id)
 
-        menu = gtk.Menu()
+        menu = Gtk.Menu()
         menu.connect("deactivate", self.menu_deactivate)
 
         # TODO: Show title item in bold font
-        title = gtk.MenuItem(label="Session control")
+        title = Gtk.MenuItem(label="Session control")
         title.set_sensitive(False)
         menu.append(title)
 
-        disconnect = gtk.MenuItem(label=DISCONNECT_TEXT)
+        disconnect = Gtk.MenuItem(label=DISCONNECT_TEXT)
         disconnect.connect("activate", self.item_activate, DISCONNECT)
         menu.append(disconnect)
 
-        terminate = gtk.MenuItem(label=TERMINATE_TEXT)
+        terminate = Gtk.MenuItem(label=TERMINATE_TEXT)
         terminate.connect("activate", self.item_activate, TERMINATE)
         menu.append(terminate)
 
-        menu.append(gtk.SeparatorMenuItem())
+        menu.append(Gtk.SeparatorMenuItem())
 
-        cancel = gtk.MenuItem(label=CANCEL_TEXT)
+        cancel = Gtk.MenuItem(label=CANCEL_TEXT)
         menu.append(cancel)
 
         menu.show_all()
 
         menu.popup(parent_menu_shell=None, parent_menu_item=None,
                    func=self.pos_menu, data=win,
-                   button=0, activate_time=gtk.get_current_event_time())
+                   button=0, activate_time=Gtk.get_current_event_time())
 
-        gtk.main()
+        Gtk.main()
 
         return self.result
 
     def item_activate(self, _, result):
         """ called when a menu item is selected """
         self.result = result
-        gtk.main_quit()
+        Gtk.main_quit()
 
     @staticmethod
     def menu_deactivate(_):
         """ called when menu is deactivated """
-        gtk.main_quit()
+        Gtk.main_quit()
 
     @staticmethod
-    def pos_menu(menu, parent):
+    def pos_menu(menu, _xpos, _ypos, *data):
         """ Positions menu at the top center of the parent window. """
+        parent = data[0]
+
         # Get parent geometry and origin
-        (_, _, win_width, _, _) = parent.get_geometry()
-        (win_x, win_y) = parent.get_origin()
+        _, _, win_width, _ = parent.get_geometry()
+        _, win_x, win_y = parent.get_origin()
 
         # Calculate width of menu
-        (menu_width, _) = menu.size_request()
+        #menu_width = menu.get_preferred_width().natural_width
+        menu_width = menu.get_allocated_width()
 
         # Calculate center
         center_x = win_x + ((win_width - menu_width) / 2)
@@ -164,12 +169,13 @@ def show_yes_no_suspend_box(title, text):
     @return: Choosen action
 
     """
-    dlg = gtk.MessageDialog(type=gtk.MESSAGE_QUESTION, flags=gtk.DIALOG_MODAL)
+    dlg = Gtk.MessageDialog(type=Gtk.MessageType.QUESTION,
+                            flags=Gtk.DialogFlags.MODAL)
     dlg.set_title(title)
     dlg.set_markup(text)
     dlg.add_button(DISCONNECT_TEXT, DISCONNECT)
     dlg.add_button(TERMINATE_TEXT, TERMINATE)
-    dlg.add_button(CANCEL_TEXT, gtk.RESPONSE_CANCEL)
+    dlg.add_button(CANCEL_TEXT, Gtk.ResponseType.CANCEL)
 
     res = dlg.run()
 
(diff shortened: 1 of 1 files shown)


sunweaver:
Hi,

On  Sa 09 Feb 2019 12:55:16 CET, Ulrich Sibiller wrote:

> By adding NX_CLIENT=<path to nxdialog.py> to the envrionment
> certain actions in nx will open dialogs. If the options string contains
> "menu=1" rootless windows will show a popup menu when the mouse hits
> the center of the title bar offering session disconnect/terminatation.

Nice! This deprecates the local branch that I have for nxdialog stuff.  
Thanks for working on this.

I remember that I had worked on the evokation of nxdialog in my local  
branch. Need to revisit and check what I had there. I think it was  
something about the default error messages and how nxagent behaves if  
the NXCLIENT variable is unset / the default executable cannot be found.

> @sunweaver: can you please take care of the following:
> - check if licensing is ok. Do we need to add our license header?

License wise, this is ok.

> - check integration in to debian packaging stuff including  
> dependencies (compilation's not required)

Will do later. Not having RPM nor DEB packaging is not a blocker for  
merging this PR. The packaging stuff can follow.

> @Ionic: can you please check the dependencies/packaging for rpm based distros
> You can view, comment on, or merge this pull request online at:

If Ionic is unavailable, I can work on this, too. Just let me know.

I will go over the Python code and test the script later, then send  
out my PR approval.

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

nxdialog/nxdialog.py
```diff
@@ -0,0 +1,368 @@
+#!/usr/bin/env python
+#
+
+# Copyright (C) 2008 Google Inc.
```
sunweaver: @uli42: please add at least yourself as copyright holder here. As you have reworked the code.

nxdialog/nxdialog.py
```diff
@@ -0,0 +1,368 @@
...
+# - pulled in the few parts of the neatx python modules that are actually
+#   to make it a standlone script
+# - added usage output
+# - dropped logging code, print errors to stderr
+# - can handle the "yesno" dialog type
+# - added missing docstrings
+# - pylint improvements
+# - removed neatx entry from the pulldoww menu
+
+"""nxdialog program for handling dialog display."""
```
sunweaver: Commenting this here, but it is about the filename used: please rename nxdialog.py to nxdialog and maybe even move it into ``<base>/nxdialog/bin/``. We at least will have to add a man page to this script, so there will be a folder ``<base>/nxdialog/data/`` (or ``man/``) coming soon.

nxdialog/nxdialog.py
```diff
@@ -0,0 +1,368 @@
...
+import optparse
+import os
+import signal
+import sys
+
+import gtk
+
+import pygtk
+pygtk.require("2.0")
+
```
sunweaver: Bad idea, using PyGTK. It is unmaintained upstream and will be removed from distros soon (Debian removal planned for Debian 11). So, we need to port this to Gtk3.

nxdialog/nxdialog.py
```diff
@@ -0,0 +1,368 @@
...
+  def Show(self):
+    """ Shows popup and returns result. """
+
+    win = gtk.gdk.window_foreign_new(self._window_id)
+
+    menu = gtk.Menu()
+    menu.connect("deactivate", self._MenuDeactivate)
+
+    # TODO: Show title item in bold font
+    title = gtk.MenuItem(label="Session control")
```
sunweaver: Shouldn't we make this whole nxdialog i18n aware? This can be a second PR, but let's bear this in mind.

nxdialog/nxdialog.py
```diff
@@ -0,0 +1,368 @@
...
+
+def ShowYesNoSuspendBox(title, text):
+  """ Shows a message box to disconnect/terminate session.
+
+  @type title: str
+  @param title: Message box title
+  @type text: str
+  @param text: Message box text
+  @return: Choosen action
+
```
sunweaver: These __doc__ strings seem to be epydoc Syntax. Epydoc never got ported to Python3, so I recommend using a Sphinx compatible __doc__ string syntax.

nxdialog/nxdialog.py
```diff
@@ -0,0 +1,368 @@
+#!/usr/bin/env python
+#
```
sunweaver: Cough cough... Python2 is end-of-life in 2020.

nxdialog/nxdialog.py
```diff
@@ -0,0 +1,368 @@
...
+                             help="class of the message (info, warning, error) \
+                             default: info) [currently unimplemented]"),
+        optparse.make_option("--local", action="store_true", dest="local",
+                             help="specify that proxy mode is used \
+                             [currently unimplemented]"),
+        optparse.make_option("--allowmultiple", action="store_true",
+                             dest="allowmultiple",
+                             help="allow launching more than one dialog with \
+                             the same message [currently unimplemented]"),
+    ]
```
sunweaver: optparse module is deprecated since Python 2.7. Switch to argparser module required.

See: https://stackoverflow.com/questions/3217673/why-use-argparse-rather-than-optparse#3217687

sunweaver (changes requested):
I left some comments, the most grave ones, I guess. Without testing the nxdialog code at runtime.

@uli42: as my guess is (and I may be mistaken) that I am more at home in the Python realm than you: shall I pick this up and work myself through my own comments? I can't promise a quick work-through, but I am interested in doing it. What do you think?

nxdialog/nxdialog.py
uli42: I guess for now it is better than nothing. Besides, I have just adapted the code...

nxdialog/nxdialog.py
uli42: yes, but remember the strings actually come from nxagent, not from nxdialog itself.

nxdialog/nxdialog.py
uli42: sigh...

nxdialog/nxdialog.py
uli42: i have renamed it but as it is only one file there's no sense in having a bin directorxy. There's also no bin directory for nxagent or nxproxy

nxdialog/nxdialog.py
uli42: done

nxdialog/nxdialog
```diff
@@ -0,0 +1,369 @@
+#!/usr/bin/env python
```
mjtrangoni, line 1: @uli42 See [here](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8-beta/html-single/8.0_beta_release_notes/index#web_servers_databases_dynamic_languages_2).

```
In RHEL 8, executable Python scripts are expected to use hashbangs (shebangs) specifying explicitly at least the major Python version.
```

This should be **python2** or even better **python3**

nxdialog/nxdialog
```diff
@@ -0,0 +1,369 @@
...
+    parser = optparse.OptionParser(option_list=self.BuildOptions(),
+                                   formatter=optparse.TitledHelpFormatter())
+    return parser.parse_args()
+
+  def BuildOptions(self):
+    """ build options for the parser """
+    return [
+        # nxagent 3.5.99.18 only used yesno, ok, pulldown and yesnosuspend
+        # yesno dialogs will always kill the session id "yes" is selected
+        optparse.make_option("--dialog", type="string", dest="dialog_type",
```
mjtrangoni: @uli42 why not [argparse](https://docs.python.org/dev/library/argparse.html?highlight=s#upgrading-optparse-code)?

mjtrangoni (commented):
I think this should be tested with python3, because python2 is EOL soon [see](https://pythonclock.org/).

nxdialog/nxdialog.py
mjtrangoni: +1

nxdialog/nxdialog
uli42: i have changed it to python2. I am wondering why I haven't done that right at the beginning.

uli42: python3 conversion is a bit of a problem since we also want to use it on RHEL6 and RHEL7 where python3 is not available by default.

nxdialog/nxdialog
uli42: because I have just adapted the original code from neatx which used optparse.

nxdialog/nxdialog.py
uli42: I have switched from PyGtk to PyGObject and Gtk3. Please check!

mjtrangoni:
@uli42 what’s about gtk3 and RHEL6?

sunweaver:
Hi

On  Do 28 Feb 2019 07:53:30 CET, Mario Trangoni wrote:

> @uli42 what’s about gtk3 and RHEL6?

GTK2 and Python will be end-of-life in 2020, correct?

I'd suggest not packaging nxdialog on distros that don't support  
Python3/GTK3. Period.

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

mjtrangoni:
@sunweaver I agree with you. This would be Fedora and firstly RHEL 8.

uli42:
I don't see a point in forcing python3 if we lose lots of distros by this.
In an enterprise environment you cannot simply switch to the latest and
greatest distro for various reasons (budget, hardware support, number of
machines, application support,...). Maybe we can write the code feasible
for python2 _and_ 3

nxdialog/nxdialog.py
uli42: converted to argparse

nxdialog/nxdialog
uli42: converted to argparse

nxdialog/nxdialog.py
uli42: This is a little standalone tool. No need for an api documentation. I'd prefer to leave the Epydoc parts just because they are easily understandable. No one will ever generate an api document from that.

nxdialog/nxdialog.py
uli42: RHEL7 will be around much longer than November 30, 2020 (End of Full Support, see https://access.redhat.com/support/policy/updates/errata) and still has no official python3 support. As this is an Enterprise Distribution there's no way in dropping python2 support as long as is is around.

mjtrangoni: @uli42 I meant this [here](
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8-beta/html-single/configuring_basic_system_settings/#python-shebang-mangling).

You could handle RHEL7 as an exception.

mjtrangoni: @uli42 Debian/Ubuntu seems to go the same way. See [here](
https://wiki.ubuntu.com/Python).

uli42:
Turned out very few changes were necessary for python3 compatibility. <s>When packaging for distros without python3  the shebang must be replaced by "python2"</s>

sunweaver (approved):
Nice! Looks good now.

PR #773, opened 2019-02-15
Revert "nxagent: rework Bool handling"

uli42:
This reverts commit 16cd2bbe1c4425e3fa557f9ca0723aa94a50b071.

It turned out that I had missed some of the Booleans being actually
tristate variables. I think I can fix this (they do not need to be
tristate) but I revert this for now to get back to a working state
(there are reports about non-working fullscreen mode and reconnect
problems).


sunweaver (approved):
ACK. Thanks for the regression fix.

Issue #770, opened 2019-02-14
keystrokes slightly broken


uli42:
The default keystroke file is missing "Shift=1" for viewport_move_up/right/down:
```
<!DOCTYPE NXKeystroke>
<keystrokes>
<keystroke action="close_session" Control="1" AltMeta="1" key="t" />
<keystroke action="switch_all_screens" Control="1"  AltMeta="1" key="f" />
<keystroke action="fullscreen" Control="1" Shift="1" AltMeta="1" key="f" />
<keystroke action="minimize" Control="1" AltMeta="1" key="m" />
<keystroke action="defer" Control="1" AltMeta="1" key="e" />
<keystroke action="ignore" Control="1" AltMeta="1" key="BackSpace" />
<keystroke action="force_synchronization" Control="1" AltMeta="1" key="j" />
<keystroke action="resize" Control="1" AltMeta="1" key="u" />
<keystroke action="viewport_move_left" Control="1" Shift="1" AltMeta="1" key="Left" />
<keystroke action="viewport_move_left" Control="1" Shift="1" AltMeta="1" key="KP_Left" />
<keystroke action="viewport_move_up" Control="1" AltMeta="1" key="Up" />
<keystroke action="viewport_move_up" Control="1" AltMeta="1" key="KP_Up" />
<keystroke action="viewport_move_right" Control="1" AltMeta="1" key="Right" />
<keystroke action="viewport_move_right" Control="1" AltMeta="1" key="KP_Right" />
<keystroke action="viewport_move_down" Control="1" AltMeta="1" key="Down" />
<keystroke action="viewport_move_down" Control="1" AltMeta="1" key="KP_Down" />
<keystroke action="viewport_scroll_left" Control="1" AltMeta="1" key="Left" />
<keystroke action="viewport_scroll_left" Control="1" AltMeta="1" key="KP_Left" />
<keystroke action="viewport_scroll_up" Control="1" AltMeta="1" key="Up" />
<keystroke action="viewport_scroll_up" Control="1" AltMeta="1" key="KP_Up" />
<keystroke action="viewport_scroll_right" Control="1" AltMeta="1" key="Right" />
<keystroke action="viewport_scroll_right" Control="1" AltMeta="1" key="KP_Right" />
<keystroke action="viewport_scroll_down" Control="1" AltMeta="1" key="Down" />
<keystroke action="viewport_scroll_down" Control="1" AltMeta="1" key="KP_Down" />
[...]

PR #774, opened 2019-02-15
keystrokes.cfg: Add missing Shift=1 for viewport_move_up/down/right

uli42:
These keystrokes where the same as their viewport_scroll variants and made
the latter being disabled.

Fixes ArcticaProject/nx-libs#770


sunweaver (approved):
Nice catch!

PR #776, opened 2019-02-17
Pr/session types

uli42:


sunweaver (approved):
ACK. I cleaned up the commits a bit. Both commits on this PR contained changes to nxcomp/src/Misc.cpp and the session type names in the help output of nxproxy. I flattened those two changes. The Keyboard.c fix stays as a separate commit.

uli42:
Thx. That happens when you are trying to do several things in parallel...
;-)

Issue #756, opened 2019-01-04
libXcomp.so.3: undefined symbol: pthread_key_create


sunweaver:
14:53 < oe1skw> hello, can anyone help me with an x2go problem on ubuntu : /usr/bin/nxproxy: symbol lookup error: /usr/lib/x86_64-linux-gnu/libXcomp.so.3: undefined symbol: pthread_key_create

sunweaver:
A workaround for the DEB builds has been applied to the 3.5.99.19 release: https://github.com/ArcticaProject/nx-libs/commit/da9a1bc0b9a42a9f586d5ac0d0392a89f4fe821b

This needs more love and work and better ideas.

@Ionic: Do you think you could take a look at this???

sunweaver:
Another finding is that m4/ax_pthread.m4 tests pthread availability and flag usage using gcc, but nxcomp is built using g++.

might it be required to apply a patch like this to ax_pthread.m4?

```
diff --git a/m4/ax_pthread.m4 b/m4/ax_pthread.m4
index 5fbf9fe0d..4d7eddc0c 100644
--- a/m4/ax_pthread.m4
+++ b/m4/ax_pthread.m4
@@ -89,7 +89,6 @@ AC_DEFUN([AX_PTHREAD], [
 AC_REQUIRE([AC_CANONICAL_HOST])
 AC_REQUIRE([AC_PROG_CC])
 AC_REQUIRE([AC_PROG_SED])
-AC_LANG_PUSH([C])
 ax_pthread_ok=no
 
 # We used to check for pthread.h first, but this fails if pthread.h
@@ -481,5 +480,4 @@ else
         ax_pthread_ok=no
         $2
 fi
-AC_LANG_POP
 ])dnl AX_PTHREAD

```

Ionic:
This workaround shouldn't be necessary in the first place.

It's essentially what I discovered myself and tried to workaround using `m4/ax_pthread.m4` and explicitly mentioning `$(PTHREAD_LIBS)` or the like in `Makefile.am`.

It only checking the C compiler shouldn't be a problem, though, we're interested in the flags needed to pull-in pthread support.

Something must have changed in between me applying that workaround and the new breakage here?

Ionic:
Maybe the thing that changed is actually GCC, i.e., maybe it doesn't need `-lpthread` any longer. I'd have to look up what the variables are set to in a build.

What Debian platforms are affected by this problem exactly? Buster? Sid?

[...]

PR #790, opened 2019-03-01
nxcomp/src/Makefile.am: Correct usage of PTHREAD_CFLAGS and PTHREAD_L…

sunweaver:
…IBS. PTHREAD_CFLAGS is also supposed to be used when linking. See comments in m4/ax_pthread.m4.

Attributes #756.

### DIFF
--- a/nxcomp/src/Makefile.am
+++ b/nxcomp/src/Makefile.am
@@ -121,7 +121,6 @@ libXcomp_la_LIBADD =					\
     @JPEG_LIBS@						\
     @PNG_LIBS@						\
     @Z_LIBS@						\
-    @PTHREAD_LIBS@					\
     $(NULL)
 
 AM_CXXFLAGS = 						\
@@ -129,17 +128,18 @@ AM_CXXFLAGS = 						\
     $(JPEG_CFLAGS)					\
     $(PNG_CFLAGS)					\
     $(Z_CFLAGS)						\
-    $(PTHREAD_CFLAGS)					\
     $(NULL)
 
 AM_CPPFLAGS =						\
     -I$(top_srcdir)/include				\
+    $(PTHREAD_CFLAGS)					\
     $(NULL)
 
 libXcomp_la_LDFLAGS =					\
     -version-number @LT_COMP_VERSION@			\
     -no-undefined					\
     @PTHREAD_LIBS@					\
+    $(PTHREAD_CFLAGS)					\
     $(NULL)
 
 libXcompincludedir = $(includedir)/nx


sunweaver:
@Ionic: can you take a look at this one? ax_pthread.m4 says in its comment part that the PTHREAD_CFLAGS shall be added to the linker options. Should we do this?

Ionic:
Yeah, I've been looking into this as well and you're generally right.

Technically, though, we didn't need it, since `libtool` already adds `C(XX)FLAGS` to the `C(XX)LINK` commands. Adding the CFLAGS to `AM_CPPFLAGS` is something I definitely missed, but likewise didn't show up because the preprocessor has never been used standalone.

nxcomp/src/Makefile.am
```diff
@@ -121,25 +121,25 @@ libXcomp_la_LIBADD =					\
     @JPEG_LIBS@						\
     @PNG_LIBS@						\
     @Z_LIBS@						\
-    @PTHREAD_LIBS@					\
```
Ionic, line 124: Did you remove that because we already have `@PTHREAD_LIBS@` in `..._LDFLAGS`?

I tend do agree with that, because `LIBADD` is only meant to be used with flags like `-L` and `-l`, while `LDFLAGS` can contain any linker flag. We don't really know what `AX_PTHREAD` will spit out as `PTHREAD_LIBS`, so it's probably safer to use `LDFLAGS` unless we'd knew that it will only ever contain `LIBADD`-supported linker flags (even though that's currently the case, as said, but it's not guaranteed that it won't change later on).

sunweaver:
On  Mo 08 Apr 2019 00:53:24 CEST, Mihai Moldovan wrote:

> Yeah, I've been looking into this as well and you're generally right.
>
> Technically, though, we didn't need it, since `libtool` already adds  
> `C(XX)FLAGS` to the `C(XX)LINK` commands. Adding the CFLAGS to  
> `AM_CPPFLAGS` is something I definitely missed, but likewise didn't  
> show up because the preprocessor has never been used standalone.

Yeah, it is more a formalistic changeset than a technically required change.

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

sunweaver:
On  Mo 08 Apr 2019 01:10:07 CEST, Mihai Moldovan wrote:

> Ionic commented on this pull request.
>
>
>
>> @@ -121,25 +121,25 @@ libXcomp_la_LIBADD =					\
>      @JPEG_LIBS@						\
>      @PNG_LIBS@						\
>      @Z_LIBS@						\
> -    @PTHREAD_LIBS@					\
>
> Did you remove that because we already have `@PTHREAD_LIBS` in `..._LDFLAGS`?
>
> I tend do agree with that, because `LIBADD` is only meant to be used  
> with flags like `-L` and `-l`, while `LDFLAGS` can contain any  
> linker flag. We don't really know what `AX_PTHREAD` will spit out as  
> `PTHREAD_LIBS`, so it's probably safer to use `LDFLAGS` unless we'd  
> knew that it will only ever contain `LIBADD`-supported linker flags  
> (even though that's currently the case, as said, but it's not  
> guaranteed that it won't change later on).

Exactly. That was my reasoning here, too.

Mike
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4354) 8390 139

Ionic (approved):

PR #791, opened 2019-04-04
Fix libXcomp pthread underlinking.

Ionic:
As per discussion in ArcticaProject/nx-libs#756.

Do not merge yet, I'll have to do a few test builds to see if that works correctly.


Ionic:
Okay, looks good so far.

Underlinking issue is gone in `libXcomp`, but the overlinking issue in `nxproxy` still there. There's nothing I could really do about that, though, since `libtool` seems to be adding `-pthread` by itself without us doing any of this stuff... and it's not a real problem, so we can probably ignore that.

sunweaver (approved):
Thanks! LGTM.

PR #810, opened 2019-05-22
Fix NXevents.c: Reactivate errornously commented function

uli42:
CheckMotion() had been commented in
add881931f2e702fb1952f4e1baba04b3dc536ee as it looked identical to the
version from dix/events.c except for some commented code. But this
based (probably) on a thinko - code that had been disabled by NX
became active again this way. Fix this by removing the comments and
by adding #ifdef/else to emphasize the difference.

Some other small fixes are also included


sunweaver (approved):
LGTM. Thanks for spotting the CheckMotion issue.

PR #814, opened 2019-06-10
Pr/use selection callback

uli42:
drop own copy of ProcSetSelectionOwner by providing a callback function. Unfortunately we need to introduce another trap to avoid endless loops.


sunweaver (approved):
Nice!

PR #818, opened 2019-06-22
rework xkb device private handling

uli42:
We can only free the xkbDevicePrivate because we do not know the
details if any other (possible) extension. So let's limit to that one
private for now and call the new xkbFreePrivates from dix (where such
a function is completely missing).


sunweaver (approved):
LGTM

PR #820, opened 2019-06-22
TravisCI: Update GCC from 8.x to 9.x

mjtrangoni:
Hi @sunweaver,

This update the latest GCC used from 8.x to 9.x. See this [RHEL article](https://developers.redhat.com/blog/2019/03/08/usability-improvements-in-gcc-9/). 
I also added a `${CC} --version` command, so we can see exactly which compiler version was used, as this flag works with both GCC and CLANG.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -117,5 +117,7 @@ before_install:
 script:
    # run static analysis tools
    - ./run-static-analysis.sh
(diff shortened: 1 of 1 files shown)


sunweaver (approved):
Nice! Thanks!

PR #828, opened 2019-07-23
nxagent: simplify nxagentWaitEvents()

uli42:
no more need to pass down a struct, we now only pass the milliseconds
and let the function do the rest.


sunweaver (approved):
LGTM, I'll merge later this afternoon.

PR #831, opened 2019-08-10
Pr/fix background

uli42:
fix background being garbled. Also added -wr switch to have a black background. After that it became obvious we show an invisible splash window in nxagent mode. Fixed that, too.


uli42:
WARNING: DO NOT MERGE!

This crashes in nxagentClearSplash, must recheck.

PR #842, opened 2019-09-01
nxagent: drop nxagentRenderTrap

uli42:
This flag is set when internal pixmap creation returns a pixmap with
depth 0. Following that the RENDER extension is effectively
disabled. However: there's no code that could create a pixmap with
depth 0, so let's simplify our code by removing nxagentRenderTrap
altogether.


sunweaver (approved):
Ok.

PR #844, opened 2019-09-17
Keystroke.c: Grammar fix in stdout message (Current known -> Currentl…

sunweaver:
…y known).


uli42 (approved):
This is ok for me, but please note that the old "current" was also correct, IMHO, meaning the "current keystrokes" nxagent knows about. This can change with the reload of the keystrokes config.

PR #845, opened 2019-09-28
nxagent/Imakefile: remove ../../exports/lib

uli42:
nxagent links system libraries that link against libX11. Unfortunately,
nxagent (and libXcompshad) require a modified libX11 version,
nicknamed libNX_X11, for proper functioning. Fortunately, this one can
act as a drop-in replacement for the system libX11. So we’ll hack our
way out: add a DT_NEEDED entry for libX11 by linking against the
system library when building nxagent and link the other system
libraries later and set DT_RUNPATH to a special directory containing
symlinks from libNX_X11 to libX11.

This tricks the loader into pulling in the "fake" libX11 version
without checking its SONAME and thus satisfying the DT_NEEDED entry
early on - specifically also for the system libraries.

Ex.:
readelf -a /usr/bin/nxagent | grep NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libX11.so.6]
...
 0x0000000000000001 (NEEDED)             Shared library: [libNX_X11.so.6]
...

If, however, you run the build a second time, it will find the newly
created libX11 links in ../../exports/lib (used as an additional
library search path) instead of the system libX11. The results in a
binary looking like this:

 0x0000000000000001 (NEEDED)             Shared library: [libNX_X11.so.6]
...

With such a setup, the whole magic falls apart and system libraries
will suddenly pull in the actual system libX11 file.

By removing LDPRELIBS from NXAGENTSYSLIBS, ../../exports/lib is no
longer used during building/linking and the binaries are built
reproducibly..


sunweaver (approved):
LGTM

sunweaver:
@Ionic: you take a second look and merge?

sunweaver (approved):
Approving again. Clicked some recycle symbol in Github UI and my approval got removed. Re-adding it.

Ionic (changes requested):
Note quite happy with the commit message yet, I'll rework it and force-push to this branch.

uli42:
As this commit will kindof vanish once we integrate the automake
branch I'd say leave it as it is and add a better commit message when
integrating that linking stuff into the automake branch
(https://github.com/uli42/nx-libs/tree/mjtrangoni/xserver-autotools).
Which, BTW, is something I hope you can integrate, too, since I was
failing so far...

Ionic:
I already did so now and had to rebase the branch against current 3.6.x anyway.

This issue took us hours to understand, so it should get properly documented. The commit itself also won't vanish but always be visible in the history. :)

Please check it.

Ionic (approved):

Ionic:
(Answering here, because Github won't show comments on outdated commits.)

> Well, one more thing: Are you really sure nxcompshad requires _our_ libX11?

I'm... not. We're using the same mechanism in `libXcompshad`, though, because it was historically part of the `nxagent` build tree and thus always linked against our "fake" version. I've never tried using it with the system libX11... and doing so is difficult anyway, because the only consumer currently is `nxagent` which will effectively hide the system libX11.

PR #848, opened 2019-10-03
Fix segfault in argument parsing

### LINKED ISSUE #847: Segfault when missing value
run `nxagent -ac :55 -options nx/nx,fullscreen` will produce a segfault:

```
arning: Option file doesn't contain a port specification.

Program received signal SIGSEGV, Segmentation fault.
__strcmp_avx2 () at ../sysdeps/x86_64/multiarch/strcmp-avx2.S:101
101     ../sysdeps/x86_64/multiarch/strcmp-avx2.S: Datei oder Verzeichnis nicht gefunden.
(gdb) bt
#0  __strcmp_avx2 () at ../sysdeps/x86_64/multiarch/strcmp-avx2.S:101
#1  0x0000555555614933 in nxagentParseSingleOption (name=0x555555afa4f6 "fullscreen", value=0x0) at Args.c:1174
#2  0x0000555555615749 in nxagentParseOptionString (string=0x555555afa4f6 "fullscreen") at Args.c:1634
#3  0x00005555556157d7 in nxagentProcessOptions (string=0x555555afa4f0 "nx/nx,fullscreen") at Args.c:1653
#4  0x00005555556128eb in ddxProcessArgument (argc=4, argv=0x7fffffffdd98, i=1) at Args.c:304
#5  0x00005555556039bb in ProcessCommandLine (argc=4, argv=0x7fffffffdd98) at utils.c:718
#6  0x0000555555599975 in main (argc=4, argv=0x7fffffffdd98, envp=0x7fffffffddc0) at main.c:211
```

The problem here is that `value` is NULL, which is something strcmp cannot handle.

uli42:
Using arguments that need a parameter without providing one would lead
a segfault due to calling strcmp() will NULL. Triggered by running
"nxagent -ac :1 -options nx/nx,fullscreen" or similar,

Fixes ArcticaProject/nx-libs#847

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -1103,6 +1103,9 @@ static void nxagentParseSingleOption(char *name, char *value)
 
   URLDecodeInPlace(value);
 
+  if (!value)
+    value = "";
+
   if (!strcmp(name, "kbtype") ||
           !strcmp(name, "keyboard") ||
               !strcmp(name, "id") ||


Ionic (approved):
Looks okay... the argument parsing code generally is a bit unsafe when it comes to `name`/`value`/`argv`, but I haven't found any location that would try to `free()` or modify this data.

PR #849, opened 2019-10-03
Screen.c: fix a memleak

uli42:
Remove some code that is no longer needed because the code requiring
it has been removed some time ago (commit
643e13bf3de6704f634d60342b738e0002f057b9). Fixes a small memleak that
turned up after switching to fullscreen once.


Ionic (approved):
LGTM.

PR #852, opened 2019-10-07
NXextension.c: Drop redundant code

uli42:
used to be different from upstream because of special NXRenderTrap
handling but as we dropped that in
5aee32e9d8db69fa0ce372e7b2b08696224e48ec we can now remove this code,
too.


uli42:
superseeded by #861

PR #864, opened 2019-10-28
Simplify (NX)xvdisp.c

uli42:
NXxvdisp.c only exists to set/unset nxagentXvTrap before/after
dispatch. There's no need to duplicate the original code. We now
rename the original dispatch functions and call them in our dispatch
code.

Also drop check for sun and cygwin, as they never appeared in xorg
upstream code.


sunweaver (approved):
LGTM

PR #868, opened 2019-10-29
NXglxext.c: use upstream versions of Dispatch functions

uli42:
instead of a full copy. We still have our own function because we
need to handle the nxagentGlxTrap. This trap is now set before the
start of the dispatcher while it has been set only directly before
calling the dispatched function.

Saves ~50 duplicated lines.


sunweaver (approved):
LGTM

PR #870, opened 2019-10-30
Pr/fix broken damage

uli42:
Fix an previous commit that broke NXdamage.c, then backport an Xorg upstream fix and finally drop NXdamage.c


sunweaver (approved):
LGTM

PR #871, opened 2019-10-30
nxagent: make second build work

uli42:
nxagent links system libraries that link against libX11. Unfortunately,
nxagent (and libXcompshad) require a modified libX11 version,
nicknamed libNX_X11, for proper functioning. Fortunately, this one can
act as a drop-in replacement for the system libX11. So we’ll hack our
way out: add a DT_NEEDED entry for libX11 by linking against the
system library when building nxagent and link the other system
libraries later and set DT_RUNPATH to a special directory containing
symlinks from libNX_X11 to libX11.

This tricks the loader into pulling in the "fake" libX11 version
without checking its SONAME and thus satisfying the DT_NEEDED entry
early on - specifically also for the system libraries.

Ex.:
readelf -a /usr/bin/nxagent | grep NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libX11.so.6]
...
 0x0000000000000001 (NEEDED)             Shared library: [libNX_X11.so.6]
...

If, however, you run the build a second time, it will find the newly
created libX11 links in ../../exports/lib (used as an additional
library search path) instead of the system libX11. The results in a
binary looking like this:

 0x0000000000000001 (NEEDED)             Shared library: [libNX_X11.so.6]
...

With such a setup, the whole magic falls apart and system libraries
will suddenly pull in the actual system libX11 file.

By removing LDPRELIBS from NXAGENTSYSLIBS, ../../exports/lib is no
longer used during building/linking and the binaries are built
reproducibly..

Note: while -L parameters in the linker call are checked in order they
still apply to ALL -l parameters, regardless odf their position! So we
had to ensure ../../exports/lib is not contained at all.

Thanks to Mihai for figuring this out.


Ionic:
But we already have that as b55c66e8ab3ed004043c54101aae8935ca7471eb?!

uli42:
is already merged (#845), for unknown reasons I have missed that.

uli42:
Interesting, I had checked in my local tree and it was not there. I have
pulled again and now it is there.

Issue #874, opened 2019-11-15
Check etersoft commits


uli42:
https://gitlab.eterfund.ru/dimbor/nx-libs/commits/dev2

There's a clipboard fix, some motif stuff (?), some locale stuff and channel stuff (separate issue for that: #873)

uli42:
@dimbor_ru: can you tell us more about those motif changes?

dimbor-ru:
About motif&locale. This old patch is not worthy of your attention at least for two reasons:
1) The first versions of wine (on server side) and nxwin (on client) could not boast of strict adherence to ICCCM. These facts gave me a reason to bully in relation to them "for a specific task".
2) This hack is also because it was time-consuming and too lazy to organize the conversion of nxwin window titles under cygwin. ;)
The patch is present in my repo only for backward compatibility purposes. And its impact on other applications is unpredictable.

uli42:
Thanks for the clarification. So it boils down to the clipboard patch, which essentially matches one of the FIXMEs I have introduced. So I will extend the existing code rather than using your patch (which drops all the comments).

dimbor-ru:
Ok. Sorry for being careless with your code. I had a hole and I had to shut it up quickly. Naturally, this is a draft. I offered only #873, the rest you yourself have found from my repo ;)

uli42:
I have now written a patch based on your clipboard changes and added some more targets, see #875

PR #875, opened 2019-11-15
Pr/utf8 clipboard

uli42:
refers to #874


sunweaver (approved):
LGTM, seems I have seen the UTF-8 fix earlier, not sure why we don't already have it.

uli42:
The problem is that this is not the correct solution to the underlying
problem. More some kind of workaround. The right fix is asking the owning
client what targets it supports and passing that information to the server
side X clients. This involves adding another state machine. Maybe we could
add code that requests the targets right after the clipboard ownership is
claimed by a client.

Mike Gabriel <notifications@github.com> schrieb am Fr., 15. Nov. 2019,
22:58:

dimbor-ru:
I confirm its working now similar to the original patch #874  (regarding the localization of the Linux client clipboard)

PR #881, opened 2019-12-17
Pr/various4

uli42:


uli42:
will move these two into various3 (which I had thought to have been pulled already)

PR #883, opened 2019-12-29
nxagent: Use XlibAtom data type where appropriate

uli42:
XlibAtom should be used wherever remote atoms are handled. This is
important as Xlib data types and server data types of the same name do
not always have identical sizes.

See also https://lists.freedesktop.org/archives/xorg-devel/2015-August/047245.html

I wanted to create a 32bit version of this for testing but I did not manage to instruct imake to do so...

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Atoms.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Atoms.c
@@ -325,7 +325,7 @@ int nxagentQueryAtoms(ScreenPtr pScreen)
 
 typedef struct {
     Atom local;
-    Atom remote;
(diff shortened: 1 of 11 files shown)


sunweaver:
Unfortunately, this needs to be rebased against recent 3.6.x (I merged newest PRs first, sorry!).

uli42:
rebase complete

sunweaver (approved):
This looks very good. Hopefully, you spotted all places. What about the other Xlib* types? Are they used wrongly, too, maybe?

uli42:
Yes, could be there are places for the other Xlib types as well. I
have so far only checked the Atom stuff.

PR #888, opened 2020-01-03
Display.c: drop icon file support

uli42:
It would read nothing because the default path is non-existing on most
systems. So the builtin icons are used anyway. Also remove the xpm
files from the distribution.

### DIFF
--- a/debian/nxagent.install
+++ b/debian/nxagent.install
@@ -3,6 +3,5 @@ usr/bin/nxagent
 usr/lib/*/nx/X11/
 usr/share/man/man1/nxagent.1*
 usr/share/nx/VERSION.nxagent
-usr/share/pixmaps/nxagent.xpm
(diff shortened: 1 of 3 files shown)


sunweaver (approved):
I guess, removing icon side-loading is sane and will not cause us trouble regarding backwards compatibility.

uli42:
aaaargh, I have just pushed a corrected commit text...

sunweaver:
On  So 05 Jan 2020 21:45:18 CET, Ulrich Sibiller wrote:

> aaaargh, I have just pushed a corrected commit text...

urgh, ok. So, that got lost then?
-- 

DAS-NETZWERKTEAM
mobile: +49 (1520) 1976 148
landline: +49 (4351) 850 8940

uli42:
Just checked, no, the right one is in 3.6.x.

PR #889, opened 2020-01-04
Pr/fix startup slowness

uli42:
This will fix a several second delay on startup. Turns out our xpm file are suboptimal for the nx transport.


sunweaver (commented):
LGTM

PR #896, opened 2020-01-09
Pr/various4

uli42:


sunweaver:
@uli42: I will review this branch between 2pm and 3pm today... So, no more changes, please.

nx-X11/programs/Xserver/hw/nxagent/Client.h
```diff
@@ -59,11 +59,15 @@ extern int nxagentClientPrivateIndex;
...
 
+#undef COUNT_CLIENT_BYTES
+
+#ifdef COUNT_CLIENT_BYTES
 #define nxagentClientAddBytes(pClient, size)	\
   (nxagentClientPriv(pClient) -> clientBytes += (size))
 
 #define nxagentClientBytes(pClient)	\
     (nxagentClientPriv(pClient) -> clientBytes)
+#endif
```
sunweaver: COUNT_CLIENT_BYTES will never be defined here? What's the purpose of this change?

sunweaver (approved):
I spotted several typos in commit msgs which I will fix + rebase now.

I left one comment, but it is not grave an issue enough to delay a merge of this massive PR any further. Please revisit it anyway (about some undef THIS ifdef THIS following directly on one another).

uli42:
That code is not used anywhere so I masked it using an ifdef. See commit
message:
    Client.c: mask unused clientBytes stuff

    The only place where this was referenced outside Client.c was a
    commented section in Handlers.c. For now lets mask all that by an
    (unset) define.

We can leave out the undef if you want...

PR #899, opened 2020-01-16
Pr/refactor timeout

uli42:
-timeout (mis)used the screensaver code which required to have own copies of some dix procs. By using a separate timer for -timeout we can now use the dix scrennsaver stuff without modification.


sunweaver (approved):
LGTM

PR #902, opened 2020-01-18
Pr/big 1.4.2 upgrade v3

uli42:
This is not a real PR but merely for keeping track of my xorg 1.4.2 upgrade efforts. I think this is almost finished, YMMV.


uli42:
Replaced by https://github.com/uli42/nx-libs/tree/pr/big_1.4.2_upgrade_v4

PR #918, opened 2020-04-07
Pr/clipboard dump

uli42:


sunweaver:
@uli42: Please rebase this PR (#918). It has conflicts since the merge of #896.

uli42:
done

sunweaver (changes requested):
Is this keystroke functional for all sessions? Or only for sessions running inside nxagent that had DEBUG flags enabled and build time? Do I get it right from this PR that also the selection content gets printed to the log if dump_clipboard is enabled?

I feel awkward with this, if the above is the case. I'd rather suggest having this feature when build with DEBUG enabled or adding an extra cmdline switch that enables the dump_clipboard keystroke. I think that clipboard leakage is one of the major design problems in Xnest/nxagent (because xclient and nested xserver share all clipboard data, nothing is sandboxed). I don't feel that we should give critics of this more food.

uli42:
On Fri, May 8, 2020 at 6:54 AM Mike Gabriel <notifications@github.com> wrote:
>
> @sunweaver requested changes on this pull request.
>
> Is this keystroke functional for all sessions? Or only for sessions running inside nxagent that had DEBUG flags enabled and build time? Do I get it right from this PR that also the selection content gets printed to the log if dump_clipboard is enabled?

It is functional in normal builds. This is totally independent from DEBUG mode.

> I feel awkward with this, if the above is the case. I'd rather suggest having this feature when build with DEBUG enabled or adding an extra cmdline switch that enables the dump_clipboard keystroke. I think that clipboard leakage is one of the major design problems in Xnest/nxagent (because xclient and nested xserver share all clipboard data, nothing is sandboxed). I don't feel that we should give critics of this more food.

No, the dump contains all the internal variables and Atoms. The
intention is to help people that report clipboard problems. We can
then ask them to dump the clipboard state and post the result.

BTW: printing the clipboard "contents" it no possible without doing
some communication with other clients. This will again change the
state so it would be counter productive.

In DEBUG mode there some other locations in the code that include the
first 20 or so bytes of the selections. But this is nothing that's
included in a regular build.

I suggest you compile it and test it.

I will double check later to ensure no clipboard leakage.

I have another branch that also includes a reset_clipboard keystroke
ro reinialize in case of problems but that's not ready yet.

Uli

sunweaver (approved):
ACK.

PR #922, opened 2020-05-05
Use nxdialog as nxclient replacement

uli42:
if neither NX_CLIENT ist provided nor /usr/NX/bin/nxclient is existing
try /usr/bin/nxdialog.

Currently the path is hardcoded.


sunweaver (approved):
Yep, this should work while not breaking backwards compatibility.

PR #934, opened 2020-06-18
README.md: Add a "Binary Builds" section

uli42:
People are asking where to find current builds so add a section describing that.


sunweaver (approved):
Ok.

PR #938, opened 2020-07-07
travisCI: Fix PVS job

mjtrangoni:
@sunweaver @uli42,

As requested, this should fix the case a Job is started and no environment variable is set.

### DIFF
--- a/.pvs-studio.sh
+++ b/.pvs-studio.sh
@@ -13,11 +13,19 @@ before_install() {
 
 build_script() {
   if [ "$PVS_ANALYZE" = "yes" ]; then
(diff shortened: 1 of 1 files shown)


uli42:
I think without the variables the script should exit with 0 and possibly
print some information. Otherwise the build will still fail without
license. Or am I missing something?

Uli

Mario Trangoni <notifications@github.com> schrieb am Di., 7. Juli 2020,
19:44:

mjtrangoni:
Hi @uli42,

As *PVS_USERNAME* and *PVS_KEY* are hidden, they do not apply to other branches other than the default one (3.6.x). It should be only working only on this branch. Because of that, and as you proposed, this is exiting with 0 if those variables are missing.

PR #944, opened 2020-10-16
randr: fix broken macros / memory corruption

### LINKED ISSUE #943: Crash of 32bit nxagent / memory corruption
nxagent crashes on shutdown with segmentation fault on 32 bit. The reason is that the CloseScreen cascade (any code can add a new CloseScreen function to a screen, that is called when the screen is to be closed and must store and call the previous function. This cascade is broken and eventually in RandR code tries to call a function at 0x0 (null pointer). 

Handling the new and the previous function pointer is done using two macros:
```
#define wrap(priv,real,mem,func) {\
    ((ScreenPtr)priv)->mem = ((ScreenPtr)real)->mem; \
    ((ScreenPtr)real)->mem = func; \
}

#define unwrap(priv,real,mem) {\
    ((ScreenPtr)real)->mem = ((ScreenPtr)priv)->mem; \
}
```

The same is present on 64bit but does not lead to a crash here. I have examined the 64bit scenario in gdb:

`pScrPriv->CloseScreen` is the place where the previous function pointer is stored via the `wrap` macro within randr.c:
```
(gdb) print pScrPriv->CloseScreen 
$1 = (RRCloseScreenProcPtr) 0x0
```
But it is zero, which is unexpected (and not checked in the code - hence the crash! The `wrap` macro uses some typecast. If we use that for the lookup, too, we see:
```
(gdb) print ((ScreenPtr)pScrPriv)->CloseScreen 
$2 = (CloseScreenProcPtr) 0x8357a0 <miSpriteCloseScreen>
```

Ah, so `miSpriteCloseScreen` is the previous CloseScreen function. The cast has changed the effective address for storing that pointer:
```
(gdb) print pScrPriv
$3 = (rrScrPrivPtr) 0xbe1e70
(gdb) print &((pScrPriv)->CloseScreen)
$4 = (RRCloseScreenProcPtr *) 0xbe1f08   <-- correct location
(gdb) print &(((ScreenPtr)pScrPriv)->CloseScreen)
$5 = (CloseScreenProcPtr *) 0xbe1f60  <--- used location
```
-> the previous function should have been stored in 0xbe1f08 but was written to 0xbe1f60. So the cast in the macro is wrong!

Turns out that this cast is something we introduced during backport of RandR 1.5, xorg server upstream does not have it. After reverting to the original macro definition it works as desired:
```
#define wrap(priv,real,mem,func) {\
    priv->mem = real->mem; \
    real->mem = func; \
}

#define unwrap(priv,real,mem) {\
    real->mem = priv->mem; \
}
```

Plus: I wanted to know what data gets corrupted by this, so what is at 0xbe1f60? After some trial and error (is there a better way?) I found it:
```
(gdb) print &((pScrPriv)->rotations)
$9 = (Rotation *) 0xbe1f60
```
So on 64bit the rotations of the screen get corrupted. We are not supporting rotations right now so this is not a real problem. But it is a bug anyway and must be fixed.

Update:
for the 32bit it is not the (mainly) unused `rotations` but the heavily used `numCrtcs`.  This is overwritten by RandR and therefore kills the the stored previous function.

By fixing the macros the old value is stored at the correct position and will not be overwritten by unrelated code.

PR to come soon.

uli42:
By casting the rrScrPriv to ScreenPtr we are using the offsets from
the Screen structure for referencing data in the rrScrPriv structure
causing data corruption. As both macros use the same casting this
still works until the location where the data resides is overwritten. For
64bit the location was the rotations member in rrScrPriv which was not
problematic because it is barely used. But on 32bit it was numCrtcs
which is heavily used. On unwrap this was 0 and lead to a segfault.

Fixes ArcticaProject/nx-libs#943

### DIFF
--- a/nx-X11/programs/Xserver/randr/randr.c
+++ b/nx-X11/programs/Xserver/randr/randr.c
@@ -69,12 +69,12 @@
 static int RRNScreens;
 
 #define wrap(priv,real,mem,func) {\
-    ((ScreenPtr)priv)->mem = ((ScreenPtr)real)->mem; \
-    ((ScreenPtr)real)->mem = func; \
+    priv->mem = real->mem; \
+    real->mem = func; \
 }
 
 #define unwrap(priv,real,mem) {\
-    ((ScreenPtr)real)->mem = ((ScreenPtr)priv)->mem; \
+    real->mem = priv->mem; \
 }
 
 static int ProcRRDispatch(ClientPtr pClient);


sunweaver (approved):
This changes the code to what I see in xserver from X.org. Well done! Good catch!

PR #946, opened 2020-10-17
Xserver autotools

uli42:
This is a "bookmark" PR to not lose track of this work. It needs to be rebased to current version. And it needs a review of course. see also #947


eoli3n:
It would be really helpful to drop imake for this. Do you plan to finish it ?

uli42:
Of course. But as always there are many time constraints...

Jonathan Kirszling ***@***.***> schrieb am Di., 11. Mai 2021,
18:07:

PR #950, opened 2020-10-18
Pr/clipboard gathering2020 plus target forwarding

uli42:
"bookmark" PR to keep track of this work. The code need polishing-

With this branch clipboard gets extended to allow copy/paste of all data types the owning application offers. E.g. you can copy an image in gimp locally and paste that to your LibreOffice document in the nx session.

Previously NX could only handle a few text based formats.


uli42:
superceded by #1018

PR #976, opened 2020-12-29
Musl build

eoli3n:
Fix a part of https://github.com/ArcticaProject/nx-libs/issues/975#issuecomment-752145408


uli42:
Note that we are using quilt for patching the mesa subtree, so the mesa patch cannot be included unmodified

eoli3n:
> Note that we are using quilt for patching the mesa subtree, so the mesa patch cannot be included unmodified

I'm not the author of the patch, I don't even know what it does, I just tried it to build for void-musl from:
https://git.alpinelinux.org/aports/tree/community/nx-libs?h=master

If it needs some modifications, I'll pass to someone who knows what he does.

uli42:
Made a new patch with quilt integration, see #982

PR #978, opened 2020-12-31
nxagent: make full screen mode selectable at startup

### LINKED ISSUE #923: fullscreen on single screen not configurable
nxagent knows `-geometry fullscreen` (cmdline) and `fullscreen=1` (options file). Both work but will open a fullscreen session covering all screens. There's no option to restrict that to a single screen like with the relevant keystrokes. 

So let's support `fullscreen=2` meaning "only span one screen"

For the command line options we could use `-geometry singlescreen`.

Comments?

uli42:
fullscreen= now accepts the new value "2" for a fullscreen covering
only one screen. -geometry and geometry= now know additional keywords
"allscreens" and "onescreen".

Fixes ArcticaProject/nx-libs#923

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Args.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Args.c
@@ -571,11 +571,15 @@ int ddxProcessArgument(int argc, char *argv[], int i)
   {
     if (++i < argc)
     {
-      if (!strcmp(argv[i],"fullscreen"))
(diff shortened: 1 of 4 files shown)


sunweaver:
@uli42: I merged #980 prior to this one (because this one was unclean and contained some of the Bool-PR changes). I now rebased this one. Please revisit and review, if the changes are ok as they are now. Thanks.

uli42:
rebased my branch

PR #980, opened 2020-12-31
Pr/rework booleans v2

uli42:
This is touching almost all files making more use of the Bool (pseudo) type. Leading to leaner code, IMHO.


sunweaver (approved):
LGTM

PR #984, opened 2021-01-11
Revert "Screen.c: fix resizing bug with mutter window manager"

### LINKED ISSUE #983: Mouse unusable with nxagent (~3.5.99.24 and later)
My test setup was broken because of a dead motherboard. After quite some time I was now able to fix it and did tests with recent versions of nx-libs.

I've discovered a strange issue where the mouse is unusable. Mouse movement works but it seems like mouse buttons are mixed up somehow and mouse clicks don't work.

I've tested with some snapshots I had on disk:
nxagent 20200630 OK
nxagent 20201022 BROKEN
nxagent 20201105 BROKEN
nxagent 20210109 BROKEN

Does anybody remember changes which could result in such issues?

Thanks,
Simon

uli42:
Fixes ArcticaProject/nx-libs#983

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Init.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Init.c
@@ -610,7 +610,14 @@ void OsVendorEndRedirectErrorFFunction(void)
   nxagentEndRedirectToClientsLog();
 }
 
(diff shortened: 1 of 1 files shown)


sunweaver (approved):
Nice findings!

PR #990, opened 2021-01-18
Pr/fix fullscreen

uli42:
This fixed the crash and extends the usage output, but it does not fix `-geometry singlescreen` not working. I would still merge it since fixes the crash. I will try to find the real bug soon...


sunweaver (approved):
ACK.

PR #1008, opened 2021-03-09
Support Cygwin64

vatral:
Hello! I've been getting things up to date here, and thought I'd do a 64 bit build of recent code.

It turns out that 64 bit Cygwin defines `__CYGWIN__` instead of `__CYGWIN32__`, but other than that things seem to be working okay in my testing so far.

### DIFF
--- a/nxcomp/src/Auth.cpp
+++ b/nxcomp/src/Auth.cpp
@@ -238,7 +238,7 @@ int Auth::getCookie()
 
   char command[DEFAULT_STRING_LIMIT];
 
-  #if defined(__CYGWIN32__)
+  #if defined(__CYGWIN__) || defined(__CYGWIN32__)
 
   snprintf(command, DEFAULT_STRING_LIMIT,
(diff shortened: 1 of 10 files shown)


sunweaver (approved):
@vatral: Thanks for this. LGTM. Hope you are doing well!

sunweaver:
Waiting for the new Github Actions CI. If CI succeeds, I will merge.

uli42:
I am wondering if checking for __CYGWIN alone would be sufficient, without
distinguishing betweeen 32 and 64 bit.

vatral:
@sunweaver Thank you!

@uli42 Possibly. It's been a while since then, but I recall having some reason for doing it that way, though I don't remember what. Either it didn't work, or I didn't know when exactly `__CYGWIN__` was introduced and decided to be careful about it. Googling suggests `__CYGWIN__` should work and is an old addition: http://www.cygwin.com/faq/faq.html#faq.programming.64bitporting-cygwin64

So hmm... it would make sense to test just to make sure.

PR #1012, opened 2021-04-05
fix building with binutils >= 2.36.

Ponce:
The l option of ar in the newer binutils versions switched
from being unused to being used to specify dependencies
so here should be safely removed


sunweaver (approved):
LGTM. Thanks.

PR #1016, opened 2021-05-16
Forward ClientMessages to nxproxy side

### LINKED ISSUE #1015: Tricky minimize/maximize/restore things in rootless
Some programs are now ashamed to use wm and its buttons to perform these operations. They at least continue to use the ICCCM and Extended WM hints specifications. Namely WM_CHANGE_STATE and _NET_WM_STATE client messages sent to root window. Attached are [two green buttons](https://github.com/ArcticaProject/nx-libs/files/6402491/xminmax.c.gz) to illustrate this "outrageous" behavior. It not worked in rootless on linux nx-client (naturally on nxwin too).

I started exploring ways to solve the problem. Can be forcibly replaced root window in these client messages and deliver them to real wm. The only thing that stops me is that a similar question has already been discussed here: #830
I don't want to break something globally with my future patches and I want to hear your opinions.

uli42:
This should help with clients requesting window manager actions like
maximizing or minimizing. This is a first version as it only handles
messages of type WM_STATE_CHANGE and _NET_WM_STATE. But ICCCM and EWMH
know some more.

The other direction, setting of properties by the WM, is already
implemented in Rootless.c.

Fixes ArcticaProject/nx-libs#1015

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Events.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Events.c
@@ -4505,6 +4505,70 @@ int nxagentWaitEvents(Display *dpy, useconds_t msec)
   return 1;
 }
 
+void ForwardClientMessage(ClientPtr client, xSendEventReq *stuff)
+{
+    Atom netwmstate = MakeAtom("_NET_WM_STATE", strlen("_NET_WM_STATE"), False);
+    Atom wmchangestate = MakeAtom("WM_CHANGE_STATE", strlen("WM_CHANGE_STATE"), False);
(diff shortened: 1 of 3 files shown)


sunweaver:
@uli42: is this PR (#1016) ready to be merged? You didn't put any reviewer. Please add me as reviewer, if this one is good to go.

uli42:
This is not perfect as there might be more cases to handle. But it was/is sufficient to handle @dimbor's case. Should be revisited one day, but the current state should be merged.

sunweaver (approved):
LGTM

PR #1017, opened 2021-06-07
Fix builds against Fedora 34 and beyond.

sunweaver:


uli42 (approved):
Looks good to me. I do not have a Fedora environment to test, though. Needs to be adapted in x2goserver, too, I think.

mjtrangoni:
Nice! The GH Actions has already paid off!

PR #1024, opened 2021-10-18
nxagent: extend x2go check

uli42:
do not only depend on the program name but also check the DISPLAY
variable and the command line parameters. This mainly helps for testing
some special constructs when you want to use an alternative nxagent
for x2go connections.


sunweaver (approved):
LGTM. Two tests against Debian sid failed, but I guess that was related to missing -DUseTIRPC in debian/rules.

I have rebased this PR update branch now and wait for CI builds to succeed. Then I will merge.

PR #1026, opened 2021-10-30
GH actions: Always build with -DUseTIRPC=YES (also on Debian and Ubuntu, all versions).

sunweaver:


mjtrangoni:
@sunweaver what's about doing `-DUseTIRPC=YES` default enabled?

PR #1028, opened 2021-11-15
Pr/fix clipboard callbacks

### LINKED ISSUE #1027: xfixes forwarding not working
In current HEAD (b7034724abb859db36a3fa4312e4827141114f45) there's a bug regarding xfixes forwarding for selection changes. I think I know how to fix it but I am lacking time.

The problem is as follows:
- run a desktop nx session ("outer nxagent")
- from that session run an rootless nx session with konsole ("inner nxagent")
- make a selection in the konsole inside the rootless session -> clipboard owner is konsole
- make a selection on the real X server -> the outer nxagent will get a Xfixes SelectionSetOwner event from the real X server but will not trigger that event for the inner nxagent
- if the previous clipboard owner (konsole) relies on getting Xfixes events on every clipboard change and does not try to use XGetSelectionOwner() it will assume it is still the owner. The effect is that pastes in konsole will refer to itself and thus paste the previous clipboard content. This problem can e.g. be seen in konsole from RHEL 7.9 but not in the one from debian 10.

The source of all this is 7aa969cd4ee5fe6ecf74f82442e4a0a7491191c1 where I was driven by wanting to get rid of the trap and totally missed what the callback is required for.

I will re-implement the old behaviour soon (a simple revert is not working due to other changes), maybe I can drop the trap by other means.

uli42:
This fixes #1027 thus removing a release blocker

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
@@ -377,6 +377,11 @@ static void printSelectionStat(int index)
 
   fprintf(stderr, "  CurrentSelections[].client             %s\n", nxagentClientInfoString(curSel.client));
   fprintf(stderr, "  CurrentSelections[].window             [0x%x]\n", curSel.window);
+  if (curSel.pWin)
+    fprintf(stderr, "  CurrentSelections[].pWin               [%p] (-> [0x%x])\n", (void *)curSel.pWin, WINDOWID(curSel.pWin));
+  else
+    fprintf(stderr, "  CurrentSelections[].pWin               -\n");
+  fprintf(stderr, "  CurrentSelections[].lastTimeChanged    [%u]\n", curSel.lastTimeChanged.milliseconds);
   return;
 }
 
@@ -906,12 +911,15 @@ void nxagentHandleSelectionClearFromXServerByIndex(int index)
     return;
   }
 
+  UpdateCurrentTime();
+  TimeStamp time = ClientTimeToServerTime(CurrentTime);
+
   if (IS_LOCAL_OWNER(index))
   {
     /* Send a SelectionClear event to (our) previous owner. */
     xEvent x = {0};
     x.u.u.type = SelectionClear;
-    x.u.selectionClear.time = GetTimeInMillis();
+    x.u.selectionClear.time = time.milliseconds;
     x.u.selectionClear.window = lastSelectionOwner[index].window;
     x.u.selectionClear.atom = CurrentSelections[index].selection;
 
@@ -921,8 +929,11 @@ void nxagentHandleSelectionClearFromXServerByIndex(int index)
      * Set the root window with the NullClient as selection owner. Our
      * clients asking for the owner via XGetSelectionOwner() will get
      * this for an answer.
+     * Set the CurrentSelection data just as ProcSetSelectionOwner does.
      */
+    CurrentSelections[index].lastTimeChanged = time;
     CurrentSelections[index].window = screenInfo.screens[0]->root->drawable.id;
+    CurrentSelections[index].pWin = NULL;
     CurrentSelections[index].client = NullClient;
 
     clearSelectionOwnerData(index);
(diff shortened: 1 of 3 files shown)


sunweaver:
@uli42: Don't forget to add me as reviewer, when ready with this.

uli42:
it _IS_ ready

nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
```diff
@@ -921,15 +929,34 @@ void nxagentHandleSelectionClearFromXServerByIndex(int index)
...
     clearSelectionOwnerData(index);
 
     setClientSelectionStage(index, SelectionStageNone);
 
     invalidateTargetCache(index);
+
+    /*
+     * now call the callbacks. This is important, especially for
+     * xfixes events to work properly. Keep in mind that this will
+     * also call out own callback so we must be prepared there to not
```
sunweaver: s/own callback/own callbacks/ ?

sunweaver: or s/out/our/ ?

nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
```diff
@@ -95,7 +95,7 @@ const int nxagentMaxSelections = 2;
 
 /* store the remote atom for all selections */
 static XlibAtom *remoteSelectionAtoms = NULL;
-static Atom *localSelelectionAtoms = NULL;
+static Atom *localSelectionAtoms = NULL;
```
sunweaver, line 98: Did I miss that? I obviously did. Tsts...

nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
```diff
@@ -2394,16 +2394,7 @@ static void setSelectionOwnerOnXServer(Selection *pSelection)
...
-              nxagentClientInfoString(lastSelectionOwner[index].client),
-                  nxagentClientInfoString(pSelection->client));
-  fprintf(stderr, "%s: lastSelectionOwner.window [0x%x] -> [0x%x]\n", __func__,
-              lastSelectionOwner[index].window, pSelection->window);
-  fprintf(stderr, "%s: lastSelectionOwner.windowPtr [%p] -> [%p] [0x%x] (serverWindow: [0x%lx])\n", __func__,
-              (void *)lastSelectionOwner[index].windowPtr, (void *)pSelection->pWin,
-                  pSelection->pWin ? nxagentWindow(pSelection->pWin) : 0, serverWindow);
-  fprintf(stderr, "%s: lastSelectionOwner.lastTimeChanged [%u]\n", __func__,
-              lastSelectionOwner[index].lastTimeChanged);
+  printSelectionStat(index);
```
sunweaver, line 2398: Shouldn't the part above be shipped in a separate commit?

sunweaver: This doesn't feel related to the TimeStamp data type change...

nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
```diff
@@ -877,8 +877,9 @@ void invalidateTargetCaches(void)
 
 /*
  * This is called from Events.c dispatch loop on reception of a
- * SelectionClear event. We receive this event if someone on the real
- * X server claims the selection ownership we have/had.
+ * SelectionClear or a XFixes selection event from the real X
```
sunweaver: s/a Xfixes/an Xfixes/

sunweaver (changes requested):
Some minor changes for the comments. Please amend and then let's merge.

nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
uli42: no, we only have one callback. Of course it must read "our"...

nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
uli42: yes, it does relate to the TimeStamp data type change. I would have needed to change all those printfs where the lastTimeChanged was referred to. The easier way was to replace those printfs by a call to central subroutine that already printed everthing and was already uptodate regarding the TimeStamp change.

nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
uli42: fixed it otherwise...

PR #1047, opened 2022-05-21
nxagent: add keystrokes to switch clipboard mode on the fly

xavierog:
This pull request adds nxagent keystrokes to switch clipboard mode on the fly.
A typical use case for this feature is to prevent access to the real X server's clipboard most of the time, except for a couple seconds when this actually proves necessary. Otherly put, it simply makes nxagent's"clipboard mode" feature more flexible. Should it prove unnecessary or harmful to some users, the whole thing can be disabled by reconfiguring keystrokes.

### DIFF
--- a/nx-X11/programs/Xserver/hw/nxagent/Events.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Events.c
@@ -1071,6 +1071,14 @@ void nxagentDispatchEvents(PredicateFuncPtr predicate)
             nxagentDumpClipboardStat();
             break;
           }
+          case doClipboardBoth:
+          case doClipboardClient:
+          case doClipboardServer:
+          case doClipboardNone:
+          {
+            nxagentSwitchClipboardMode(ClipboardBoth + result - doClipboardBoth);
+            break;
+          }
           default:
           {
             FatalError("nxagentDispatchEvent: handleKeyPress returned unknown value\n");
--- a/nx-X11/programs/Xserver/hw/nxagent/Keystroke.c
+++ b/nx-X11/programs/Xserver/hw/nxagent/Keystroke.c
@@ -103,6 +103,10 @@ char * nxagentSpecialKeystrokeNames[] = {
        "autograb",
 
        "dump_clipboard",
+       "clipboard_both",
+       "clipboard_client",
+       "clipboard_server",
+       "clipboard_none",
 
        NULL,
 };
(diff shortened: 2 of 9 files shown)


uli42:
I am sorry that you did not get a reaction to this so far - I just detected it today and will have a look at it

nx-X11/programs/Xserver/hw/nxagent/Events.c
```diff
@@ -1071,6 +1071,14 @@ void nxagentDispatchEvents(PredicateFuncPtr predicate)
             nxagentDumpClipboardStat();
             break;
           }
+          case doClipboardBoth:
+          case doClipboardClient:
+          case doClipboardServer:
+          case doClipboardNone:
+          {
+            nxagentSwitchClipboardMode(ClipboardBoth + result - doClipboardBoth);
```
uli42, line 1079: I don't really like that calculation since it is not self-explanatory. Please add at least a comment the thought behind this calculation. There's another location where I saw that same calculation, comment there too, please

nx-X11/programs/Xserver/hw/nxagent/Keystroke.c
```diff
@@ -725,6 +733,12 @@ Bool nxagentCheckSpecialKeystroke(XKeyEvent *X, enum HandleEventResult *result)
     case KEYSTROKE_DUMP_CLIPBOARD:
       *result = doDumpClipboard;
       break;
+    case KEYSTROKE_SET_CLIPBOARD_BOTH:
+    case KEYSTROKE_SET_CLIPBOARD_CLIENT:
+    case KEYSTROKE_SET_CLIPBOARD_SERVER:
+    case KEYSTROKE_SET_CLIPBOARD_NONE:
+      *result = doClipboardBoth + stroke - KEYSTROKE_SET_CLIPBOARD_BOTH;
```
uli42, line 740: explain calculation (as mentioned above)

uli42 (changes requested):
I don't see code that (re-)initializes the clipboard when switching between modes. This might especially become important when nxagent was started with clipboard=none initially.

Also we need a way to disable the keystrokes since the feature might compromise security (if the admin sets the clipboard to some mode in the global config it should not be possible for the user to enable it on the fly.

sunweaver:
@xavierog Can you please give feedback on the questions asks in the code review of this PR? Thanks.

@uli42 I @xavierog will not respond, I think the idea of being able to switch between clipboard modes is basically great, but we should add yet another option that enables/disables this feature server-side/client-side. Disabled clipboard (by admin / by client user) is a feature and we would undermine it here. What do you think?

sunweaver:
@uli42 I forgot to finish my sentence above. What you be willing to finalize this PR if there is not feedback from @xavierog?

xavierog:
Hi!
Indeed, this PR is getting close to being one year old.
In practice, I am not dead, just super busy. Good news, it should get better very soon, and I should be able to dive again in this patch, answer the questions and apply the suggestions above.

sunweaver:
@xavierog This is awesome to hear!!!

uli42:
I agree that it is a nice feature. We could even extend it to a mode where only one clipboard action is allowed (pasting once) and afterwards it is disabled again. This way you get a bit more security.

PR #1061, opened 2023-06-27
nx-libs.spec: fix hashbang when manually replacing instead of using pathfix.py.

Ionic:
When I reworked the hashbang replacement line a while ago to fully replace the first line instead of just search-replacing a python binary, I forgot to add the actual hashbang character, which broke things.


sunweaver (approved):
LGTM.

PR #1062, opened 2023-08-25
Add support for new arch Loongarch

fangyaling:


sunweaver:
@fangyaling patch looks good to me. Will merge after CI has run and completed.

PR #1063, opened 2023-08-27
nx-X11/config/cf: Add loongarch64 support to nx-libs.

sunweaver:


sunweaver:
Already submitted as #1062. Thus, closing.

PR #1068, opened 2023-09-25
Fix dlopen() privilege escalation.

### LINKED ISSUE #1067: Vulnerability Report: Path traversal and Code Execution in dlopen via environment variable
### Vulnerability Report: Path traversal and Code Execution in dlopen via environment variable

Affected Project & Line:
https://github.com/ArcticaProject/nx-libs/blob/3ef7845746c4ec1ac75825ccebc17168e0400cfa/nx-X11/extras/Mesa_6.4.2/src/glx/x11/dri_glx.c#L199

#### Summary
Applications using nx-libs may be vulnerable to attacks where a local attacker could execute arbitrary code through a maliciously crafted library, loaded via the dlopen() function. This could potentially lead to privilege escalation.

#### Details
The application reads unsanitized data from the environment variable. This tainted path is subsequently used directly by dlopen() without sufficient validation, allowing directory traversal and possibly loading external malicious libraries.

The security check currently implemented using [specific security check, e.g., geteuid() == getuid()] does not adequately protect against this vulnerability.

#### Reproduction Steps
	1. Set the affected environment variable to a path containing a maliciously crafted library.
	2. Run the application or initiate the specific function that calls dlopen().
	3. Observe that the malicious code within the library gets executed.

#### Impact
Attackers with local access can load and execute arbitrary code in systems using the affected application. This can lead to data corruption, data theft, and potentially complete system compromise depending on the application's privileges.

#### Recommendation
	• Implement thorough input validation for the paths loaded via the environment variables.
	• Use a whitelist of allowed paths or directory names to mitigate the risk of arbitrary directory traversal.
	• Drop elevated privileges immediately after they are no longer required.
	• Regularly audit and review the code to ensure that all paths from which libraries or other external resources are loaded are properly validated.
	• Check and compare the real group ID and the effective group ID with getgid() and getegid()

#### CVSS Score
High 7.8 [CVSS:3.1/AV:L/AC:L/PR:L/UI:N/S:U/C:H/I:H/A:H]

#### Root Cause Analysis
The root cause of this vulnerability stems from the lack of input validation when reading paths from environment variables and the subsequent insecure use of such paths with the dlopen() function.

### Additional References/More information
https://docs.google.com/document/d/1lRE2lc00WAYa-427crBFO1yBzU7fSUmQIanh9W8Rglo/edit?usp=sharing

vatral:
Fixed by implementing the recommended GID check.

Closes #1067

### DIFF
--- a/nx-X11/extras/Mesa.patches_6.4.2/5003_dlopen-escalation.patch
+++ b/nx-X11/extras/Mesa.patches_6.4.2/5003_dlopen-escalation.patch
@@ -1,0 +1,13 @@
+Index: Mesa_6.4.2/src/glx/x11/dri_glx.c
+===================================================================
+--- Mesa_6.4.2.orig/src/glx/x11/dri_glx.c
++++ Mesa_6.4.2/src/glx/x11/dri_glx.c
+@@ -196,7 +196,7 @@ static __DRIdriver *OpenDriver(const cha
+       }
+    }
+ 
+-   if (geteuid() == getuid()) {
++   if (geteuid() == getuid() && getgid() == getegid()) {
+       /* don't allow setuid apps to use LIBGL_DRIVERS_PATH */
+       libPaths = getenv("LIBGL_DRIVERS_PATH");
+       if (!libPaths)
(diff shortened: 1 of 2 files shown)


uli42 (changes requested):
Hello Vadim, thanks for the fix! Please note: We normally do not change the mesa source code directly but add some patches using quilt. This way the mesa code stays unchanged, making updates (hopefully) simpler.

uli42 (approved):
LGTM (I cannot test currently due to broken equipment)

PR #1079, opened 2024-06-29
nxcomp/src/Loop.cpp: Support NXTransDialog caption / title being pref…

sunweaver:
…ixed with a custom prefix string.


nx-X11/programs/Xserver/hw/nxagent/Clipboard.c
```diff
@@ -1969,14 +1973,14 @@ void nxagentHandleSelectionNotifyFromXServer(XEvent *X)
...
-         * termined amount of data in a single GetProperty and by discar-
-         * ding the remaining part. This is not the optimal solution (we
-         * could get the remaining part if it doesn't fit in a single
-         * reply) but, at least with text, it should work in most situa-
-         * tions.
+         * The state machine is able to work in two phases. In the
+         * first phase we get the size of property data, in the second
+         * we get the actual data. We save a round-trip by requesting
+         * a predetermined amount of data in a single GetProperty and
+         * by discar- ding the remaining part. This is not the optimal
```
uli42: "discar- ding" looks ugly and doe snot make sende in the middle of the paragraph.

nxcomp/src/Loop.cpp
```diff
@@ -14731,7 +14731,19 @@ static void handleAlertInLoop()
...
+
+      if (!caption_prefix) {
+          // Use the hard-coded 'NX -' as default.
+          strncpy(caption, ALERT_CAPTION_PREFIX, DEFAULT_STRING_LENGTH);
+          caption[DEFAULT_STRING_LENGTH-1] = '\0';
+      }
+      else
+      {
+          // Use what gets provided via NX_DIALOG_CAPTIONPREFIX (used in X2Go)
+          strncpy(caption, caption_prefix, DEFAULT_STRING_LENGTH-1);
```
uli42, line 14744: this strncpy is different from the one above regarding the number of characters to be copied. While both are technically correct they should at least be unified.

sunweaver: Can you suggest a clean solution? (You are the C dev... ;-) ).

uli42: str(n)cpy is always tricky. I would do the getenv only once. Then strdup either the result or the default. And free the dupped string at shutdown.

sunweaver: @uli42 Can you help with this MR and fine-polish it? Your C is so much better than mine.

uli42:
I am currently limited hw-wise. But I think this looks good except that you
use strncpy in two different ways. You should make them identical regarding
the length argument.

I suggest you test and merge it. You can leave this ticket open and I will
revisit eventually.

PR #1081, opened 2024-08-02
Update GitHub Workflows

mjtrangoni:
Hi,

This PR intend to get the CI Workflows to a current state-of-the-art status.

* Update the base runner images from Ubuntu 20.04 to 22.04.
* Replace EOL centos:7 with rockylinux:8
* Remove EOL ubuntu:16.04, adds ubuntu:24.04
* Update checkout workflow
* Configure dependabot


mjtrangoni:
@uli42 PTAL. We should fix/adapt some of the linters results like python3 issues (Shall we remove python2 support?).
Also `cppcheck` and GCC are updated as well, and may report some newer issues/fixes.

PVS licenses should be ok again the next time we merge something.

